PR #164, opened 2019-08-15
Website fixing banner and header jump on load

### LINKED ISSUE #157: misc website fixes and tweaks
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Just an ongoing list of small issues that anyone can pick up and help out with.

## Details
1. [x] On the home page, the logos at the bottom aren't in evenly spaced columns
1. [x] The images like at the bottom of the Getting Started -> Build and Deploy section forces a horizontal scroll bar
1. [x] Remove browser prefixes from CSS files (also delete any unused / unneeded CSS)
1. [x] Upon first load, the logo in the banner "jumps out", looks like a [FOUC](https://en.wikipedia.org/wiki/Flash_of_unstyled_content) maybe?

hutchgrant:
## Related Issue
Resolves #157 final item

## Summary of Changes

for the banner:

* set a responsive width and height on the banner img

for the header:

* change from the h4 to a custom class, the global style's h4 specificity is higher because it's set to important which is messing with the style on load.
* override global list item margin style

### DIFF
--- a/www/components/banner/banner.css
+++ b/www/components/banner/banner.css
@@ -47,6 +47,11 @@
       & h3 {
         text-shadow: 1px 1px rgba(0, 0, 0, 0.6);
       }
+
+      & img {
+        max-width: 300px;
+        max-height: 300px;
+      }
     }
 
     & span.off {
--- a/www/components/banner/banner.css
+++ b/www/components/banner/banner.css
@@ -44,12 +44,13 @@
         text-shadow: 1px 1px rgba(0, 0, 0, 0.6);
       }
       & h3 {
+        padding-top: 10px;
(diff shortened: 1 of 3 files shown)


hutchgrant:
Though the header jump is now less obvious, we still have an issue where font size is temporarily larger in the header navigation items.

hutchgrant:
Bundling the image had no resulting visual affect vs linking to the asset locally, as that's all it's really doing anyway.

thescientist13:
> as that's all it's really doing anyway.

I don't think so.  With `import` the image is being processed and loaded via JavaScript, as opposed to just raw HTML.  The latter should always be faster, even if imperceptible to the eye 👁

PR #174, opened 2019-08-19
improve loading performance / minification for JavaScript polyfills

### LINKED ISSUE #169: loading performance impacted due to lack of minification
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Short of completing [differential loading](https://github.com/ProjectEvergreen/greenwood/issues/9) there are probably still some improvements we could make to how our polyfills are being loaded, as per [Lighthouse](https://github.com/ProjectEvergreen/greenwood/issues/31) by minifying files.
Screen Shot 2019-08-17 at 2 08 50 PM

## Details
Currently these two polyfills account for 80KB, which is a lot. 
Screen Shot 2019-08-17 at 2 04 46 PM

For example, the [minified version of polyfills.js](https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.0.0/polyfill.min.js) is `213KB`!!! (vs [`50.2KB`](https://cdnjs.cloudflare.com/ajax/libs/babel-polyfill/7.0.0/polyfill.js))

Should see the same for _web-components-bundle.js_.

Also, we should probably add some unit tests for this in our smoke test.

thescientist13:
## Related Issue
resolves #169 

## Summary of Changes
1. Use minified version of babel polyfill (further optimizations can happen as part of #9 ) (saves about 160KB)
1. Added unit tests

## TODO 
1. [ ] Find minified version of webcomponentjs - would be nice to self bundle somehow instead.  Maybe copy assets webpack plugin?
1. [x] ~~Evaluate webpack minification options~~ - tried terser, didn't have much of an impact


thescientist13:
Using minified version of Babel polyfills saved about 160KB and moving `<script>` tags to the end increase Lifghthouse score ➕ 2
Screen Shot 2019-08-19 at 4 59 26 PM

thescientist13:
Seem to confirm that the [404 page doesn't serialize as expected](#176 ) as adding the same polyfill and bundling script tag checks don't assert the same in the 404 suite in _smoke-test.js_.  In other words, these fail
```javascript
it('should have two polyfill <script> tags in the <body>', function() {
  const scriptTags = dom.window.document.querySelectorAll('body script');
  const polyfillScriptTags = Array.prototype.slice.call(scriptTags).filter(script => {
    return script.src.indexOf('//cdnjs.cloudflare.com') >= 0;
  });

  expect(polyfillScriptTags.length).to.be.equal(2);
});
```

PR #188, opened 2019-08-26
release/0.4.0

thescientist13:
## Related Issue
The minimum needed atomic changeset to deliver / release the initial 'Plugins / Packages work.
- #24
- #182 
- #179 
- #183 
- #187 (+ #178 , #140)
- #138 
- #193 
- #207 
- #213 (+ #16 ) 
- #221 
- #9 

As this is a release branch, the following branch maintenance should be followed
- PRs _into_ this branch should be squashed merged
- Rebase this branch when all work is done
- Merge (_NO_ squash) into master when ready

## Summary of Changes
1. [RFC] Integrated Lerna as a monorepo
1. [RFC] Index Hooks Plugins API
1. Google Analytics Plugin + introduce Composite Plugins concept
1. [RFC] Webpack Build Plugins
1. Polyfills Plugin:  #187 (+ pull deps from npm #178) 
1. Fixing what appears to have been some long standing race condition issues in our build, and consistencies partial exacerbated by the introduction of #187 .  Also absorbed #16 as part of the enhancement.
1. [RFC] Web Components loader

So there's more JS now because of Google Analytics, but even supporting IE11 the main bundle size went down about 30KB.  Once we can [decouple the polyfills](https://github.com/ProjectEvergreen/greenwood/issues/193), evergreen only projects will have a super small initial payload size!  I'm expecting to see some pretty good results though from Alpha Testing the Getting Started repo.

Lighthouse score is about the same, but not bad for all around for full browser support down including IE11.  🎉 

#### BEFORE
before-network
before-lighthouse

#### AFTER
after-network
after-lighthouse

## TODO
1. ~~Evaluate final plugins API (see #179) and document~~ made [an issue](https://github.com/ProjectEvergreen/greenwood/issues/209) for tracking 
1. ~~Performance Testing~~
1. ~~Alpha Testing `alpha.0`~~ 
    - https://github.com/thescientist13/greenwood-getting-started/pull/7
    - https://github.com/thegreenhouseio/www.thegreenhouse.io/pull/128
    - #208 ?
1. ~~`alpha.1`~~
1. ~~_[RFC] Polyfill feature detection: #9 * (could easily be done as a post release PR)_~~
1. - `alpha.2`

## Questions / Thoughts
1. ~~I don't think browserslist is implemented.  Should be done as part of #9? 😬~~ - DONE! 
1. ~~Might be good to create a ticket for PWA / Manifest plugin~~ Done #194 
[...]

PR #189, opened 2019-09-01
Rfc/issue 182 template hooks

### LINKED ISSUE #182: [RFC] Plugins: Index Hooks
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Taken from a comment in https://github.com/ProjectEvergreen/greenwood/issues/17#issuecomment-523706157 in regards to a general discussion around the API and capturing the specific action item re: the `hooks` plugin type here.

## Details
In particular _index.html_ provides the most opportunity right now as things like polyfills, analytics, and PWA features like _manifest.json_ and ServiceWorkers need to be installed in a global scope and /or unbundled.  These would essentially be pre-defined find / replace hooks in _index.html_.

### Usage
A great example use case is injecting Google Analytics into _index.html_. 

In Greenwood's _index.html_ we would add this marker
```html
  ...

  <%= htmlWebpackPlugin.options.hookGreenwoodAnalytics %>
```

Then in Greenwood config, you tap into it like this:
```javascript
module.exports = {
  title: 'My App'

  ...

  plugins: [{
    type: 'hook',
    provider: (compilation) => {
      // you can access things like config, context if you need from compilation
      return {
        hookGreenwoodAnalytics: `
          <!-- Google Analytics -->
          <script>
            window.ga=window.ga||function(){(ga.q=ga.q||[]).push(arguments)};ga.l=+new Date;
            ga('create', 'UA-XXXXXXX', 'auto');
            ga('send', 'pageview');
          </script>
          <script async src='https://www.google-analytics.com/analytics.js'></script>
          <!-- End Google Analytics -->
        `
      ]
    }
  }]
}
```

> The Greenwood plugin equivalent would essentially just wrap all of this for a user and just ask for `analyticsId`.

It should be noted that if these specific hook types are too limiting (we should support a basic set of common hooks), users can just make their own _index.html_ and configure all of these Greenwood hooks themselves (if they wanted) or just create their own.

thescientist13:
## Related Issue
resolves #182 

## Summary of Changes
Updated TODOs of #188 (please review first)
1. Created an API for injecting custom HTML into _index.html_ for:
    - polyfills
    - analytics
1. Added unit tests
1. Added documentation (created Plugins nav item in header)

Screen Shot 2019-08-29 at 9 04 31 PM

## Misc Thoughts / Questions / Observations
1. I wonder if we should get rid of _404.html_?  Maybe it could be it's own (composite) plugin?
1. Could introduce new hooks for _manifest.json_ and ServiceWorkers as part of a PWA plugin (should make a ticket)

### DIFF
--- a/packages/cli/src/config/webpack.config.common.js
+++ b/packages/cli/src/config/webpack.config.common.js
@@ -51,6 +51,24 @@ module.exports = ({ config, context }) => {
     }
   ];
 
+  const customHooks = config.plugins
+    .filter((plugin) => plugin.type === 'hook')
+    .map((plugin) => {
+      const supportedHooks = ['hookAnalytics', 'hookPolyfills'];
+      const providerResult = plugin.provider();
+
+      return Object.keys(providerResult).map((key) => {
+        if (key !== 'type' && supportedHooks.indexOf(key) >= 0 && providerResult[key]) {
+          return new HtmlWebpackPlugin({
+            // filename: context.indexPageTemplate,
+            template: path.join(context.scratchDir, context.indexPageTemplate),
+            [key]: providerResult[key]
+          });
+        }
+      })[0];
+
+    });
+
   return {
     entry: {
       index: path.join(context.scratchDir, 'app', 'app.js')
@@ -119,7 +137,9 @@ module.exports = ({ config, context }) => {
         filename: path.join(context.publicDir, context.indexPageTemplate),
         template: path.join(context.scratchDir, context.indexPageTemplate),
         chunksSortMode: 'dependency'
-      })
+      }),
+
+      ...customHooks
     ]
   };
 };
--- a/packages/cli/src/config/webpack.config.common.js
+++ b/packages/cli/src/config/webpack.config.common.js
@@ -51,23 +51,16 @@ module.exports = ({ config, context }) => {
     }
   ];
 
-  const customHooks = config.plugins
+  const customHooks = Object.assign({}, ...config.plugins
     .filter((plugin) => plugin.type === 'hook')
-    .map((plugin) => {
-      const supportedHooks = ['hookAnalytics', 'hookPolyfills'];
-      const providerResult = plugin.provider();
-
+    .map((plugin) => plugin.provider())
+    .filter((providerResult) => {
       return Object.keys(providerResult).map((key) => {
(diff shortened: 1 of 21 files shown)


www/pages/plugins/hooks.md
```diff
@@ -0,0 +1,66 @@
...
+<html lang="en" prefix="og:http://ogp.me/ns#">
+  <head>
+
+    <meta charset='utf-8'>
+    <meta name='viewport' content='width=device-width, initial-scale=1'/>
+    <meta name='mobile-web-app-capable' content='yes'/>
+    <meta name='apple-mobile-web-app-capable' content='yes'/>
+    <meta name='apple-mobile-web-app-status-bar-style' content='black'/>
+
+    <%= htmlWebpackPlugin.options.hookSpaIndexFallback %>
```
hutchgrant: Shouldn't we provide an example of usage of a hook besides the spafallback script.

e.g. If we give example for analytics, shouldn't we also give example of the hook usage.

packages/cli/src/config/webpack.config.common.js
```diff
@@ -51,6 +51,17 @@ module.exports = ({ config, context }) => {
     }
   ];
 
+  const customHooks = Object.assign({}, ...config.plugins
+    .filter((plugin) => plugin.type === 'hook')
```
hutchgrant: Can we call these webpackhooks? hooks in general we may have more of typess of them I'm not sure, maybe we can change that later.  when I saw template hooks, I thought it would have to do with page-templates not index.

www/pages/plugins/hooks.md
thescientist13: I see what you mean.  Will add.

packages/cli/src/config/webpack.config.common.js
thescientist13: So maybe there are two things here
1. `hooks` are the API, so other plugin types would be handled somewhere else because they would be doing a different thing specific to their own API.
1. The name `hooks` / Template Hooks may not be the most direct / semantic name, so let think of something else we could try and call it instead.

thescientist13: What about calling the plugin type _index_ instead?

Issue #134, opened 2019-07-31
[RFC] Global CSS


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of #91 and #110 the current implementation of `@import`ing a font is not working (never saw a network request for any CSS). It did seem to work when [hardcoded right into _index.html_](https://github.com/ProjectEvergreen/greenwood/pull/110#issuecomment-516647503) though.  (Might require a PostCSS plugin for some of it?)  

In addition, when [removing hardcoded CSS](#135 ) it looked global styles weren't really global.

# Details
Per @hutchgrant , the primary issue is most likely related to [ShadowDOM](https://github.com/ProjectEvergreen/greenwood/pull/135#discussion_r309019052)
>

So we should probably do another injection step as part of our build step
1. If a styles/theme.css file is found in the user workspace, inject that directly into the `<head>` of _index.html_ in packages/cli
1. Make that file path configurable (relative to workspace, so just_styles/main.css_ would work)
1. Add unit tests (as part of smoke testing ideally) and config option for `themePath`
1. Add [API Docs](#20 ) for Styles and document this feature (and in Configuration too)

This is essentially a more robust version / continuation of #7

PR #192, opened 2019-09-10
remove jsdom for prod pipeline

### LINKED ISSUE #138: deprecate usage of JSDOM
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
In regards to a comment [here](#137) just questioning usage of `JSDOM`

## Details
Will make an issue for it specifically, but have been wondering if there's a way we can remove our dependency on JSDOM? In particular in _browser.js_ it seems clunky to get the content from puppeteer (a real browser) but then just pass it back into a "fake" browser. 
```javascript
const browser = await puppeteer.launch({
  headless: true,
  args: ['--no-sandbox']
});
const renderer = new Renderer(browser);
const result = await renderer.serialize(url);

// need url: https://github.com/jsdom/jsdom/issues/2005
const dom = new JSDOM(result.content, { url });
const html = dom.serialize();
const target = path.join(outputDirectory, route);
```

I could see how JSDOM could be helpful for unit testing, so maybe we just need to make it a devDependency or just get some sort of simple DOM parsing library? Just seems like a lot to have both Puppeteer and JSDOM.

thescientist13:
## Related Issue
resolves #138 

## Summary of Changes
1. Removed "duplicate" serialization by removing JSDOM from the build pipeline.

Takes a nice chunk out of our `dependency` footprint
```
$ yarn why jsdom
yarn why v1.12.3
[1/4] 🤔  Why do we have the module "jsdom"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "jsdom@14.0.0"
info Has been hoisted to "jsdom"
info This module exists because it's specified in "devDependencies".
info Disk size without dependencies: "3.12MB"
info Disk size with unique dependencies: "7.31MB"
info Disk size with transitive dependencies: "19.57MB"
info Number of shared dependencies: 51
✨  Done in 0.91s.
```


thescientist13:
As far as I can tell, the output is the same.

hutchgrant:
There was a specific reason for using jsdom, something did not serialize properly.  Extra caution with this.

thescientist13:
> There was a specific reason for using jsdom, something did not serialize properly.

What was it?

packages/cli/lib/browser.js
```diff
@@ -12,12 +11,10 @@ module.exports = async (url, label, route, outputDirectory) => {
 
   const renderer = new Renderer(browser);
   const result = await renderer.serialize(url);
-
-  // need url: https://github.com/jsdom/jsdom/issues/2005
-  const dom = new JSDOM(result.content, { url });
```
hutchgrant:
```
const dom = new JSDOM(``, {
  url: "https://example.org/",
  referrer: "https://example.com/",
  contentType: "text/html",
  includeNodeLocations: true,
  storageQuota: 10000000
});
```

> url sets the value returned by window.location, document.URL, and document.documentURI, and affects things like resolution of relative URLs within the document and the same-origin restrictions and referrer used while fetching subresources. It defaults to "about:blank".

Removing this line could potential cause issues related to path/url of  resources. I do not know where specifically this was an issue previously.

hutchgrant: Although from my cursory tests I haven't seen any immediate issues.

hutchgrant: As per the issue cited in the comment: https://github.com/jsdom/jsdom/issues/2005
```
<style type="text/css">
@import url("/css/blok_add.css");
</style>
```
Cannot be parsed without the correct url. Although I do not see that error serializing in puppeteer, maybe because we don't have any examples of relative css imports.

With jsdom:
> The problem can be solved by specifying the parameter url.
> 
> new JSDOM(html, {url: 'http://example.com'})
>

hutchgrant: https://github.com/hutchgrant/ssr-evergreen-demo/blob/b18f31038dd24fb1e003ffdb00a18a20429e1d6c/renderer.js#L31

From my previous SSR-demo which formed the basis of this code base, I previously has puppeteer serialize, then went used JSDOM to manipulate dom, then serialized again.

Perhaps thats why we kept it? In case we needed to change dom post-serialization via puppeteer?

thescientist13: Gotta love unit tests :)

As part of the [PR that added that comment re: JSDOM](https://github.com/ProjectEvergreen/greenwood/pull/137/), we also added unit tests specifically for `@import`.  This was for issue #134 

All unit tests are passing and the website preview is working so I think all the right checks and balances are in place, IMO.

thescientist13: In general, removing an extra translation layer seems like the right move architecturally too.  If we're already pre-rendering with a "real" browser, serializing it through fake one seems unnecessary.

thescientist13: That said, if there are specific test cases we need to validate, please suggest them and I can look to see if they are being covered.

thescientist13:
going to repoint this to release branch to avoid issues with packaging changes due to monorepo

Issue #213, opened 2019-09-27
same unit test failing consistently, inconsistently on CircleCI


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
The tests have started failing consistently inconsistently recently (at least within the release/0.4.0 branch).

First it was this one
https://github.com/ProjectEvergreen/greenwood/pull/195#pullrequestreview-289849840
https://circleci.com/gh/ProjectEvergreen/greenwood/644
```
 5 failing

  1) Build Greenwood With: 
       Custom Webpack Plugin and Default Workspace
         Running Smoke Tests: Custom Webpack Plugin and Default Workspace
           Index (Home) page
             should have a <title> tag in the <head>:

      AssertionError: expected 'My App' to equal 'Greenwood App'
      + expected - actual

      -My App
      +Greenwood App
      
      at Context.<anonymous> (test/smoke-test.js:86:29)

  2) Build Greenwood With: 
       Custom Webpack Plugin and Default Workspace
         Running Smoke Tests: Custom Webpack Plugin and Default Workspace
           Index (Home) page
             should have the correct route tags in the <body>:

      AssertionError: expected 0 to equal 3
      + expected - actual

      -0
      +3
      
      at Context.<anonymous> (test/smoke-test.js:109:37)

  3) Build Greenwood With: 
       Custom Webpack Plugin and Default Workspace
         Running Smoke Tests: Custom Webpack Plugin and Default Workspace
           Index (Home) page
             should have the expected heading text within the index page in the public directory:
     TypeError: Cannot read property 'textContent' of null
      at Context.<anonymous> (test/smoke-test.js:113:64)

  4) Build Greenwood With: 
       Custom Webpack Plugin and Default Workspace
         Running Smoke Tests: Custom Webpack Plugin and Default Workspace
           Index (Home) page
             should have the expected paragraph text within the index page in the public directory:
     TypeError: Cannot read property 'textContent' of null
      at Context.<anonymous> (test/smoke-test.js:119:65)

  5) Build Greenwood With: 
       Custom Webpack Plugin and Default Workspace
         Running Smoke Tests: Custom Webpack Plugin and Default Workspace
           Hello World (dummy) page
             "before each" hook for "should output a hello page directory":
[...]

PR #215, opened 2019-09-30
remove resolve inside recursive walk directory call

thescientist13:
## Related Issue
related to #213 

## Summary of Changes
Seeing if this will help at all.  

Anther thought I will try in a separate PR is to audit the code and make sure all usage of `async / await` is consistent as I see us using it even with sync based `fs` tasks.


thescientist13:
not sure if this needed or not in light of the work done as part of #213 so closing for now.

PR #216, opened 2019-09-30
Enhancement/issue 16 parallel serialization

### LINKED ISSUE #16: Parallel Serialization
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Currently the serialization phase runs serially, one page at a time, each page as a `Promise`.
```javascript
const runBrowser = async (compilation) => {

  try {
    return await Promise.all(compilation.graph.map(({ route, label }) => {
      return browserRunner(`http://127.0.0.1:${PORT}${route}`, label, route, compilation.context.publicDir);
    }));
  } catch (err) {
    // eslint-disable-next-line no-console
    console.log(err);
    return false;
  }
};
```

It would be good to see what options there are to support running these in parallel so multiple pages could be serialized at once.

Otherwise this happens locally 😬 
Screen Shot 2019-09-27 at 7 06 50 PM

## Details
Something to consider is how to manage the "firehose" of creating and serializing pages.  It would probably make sense to support a queue so for very large applications, hundreds and thousands of pages won't get spun up at once and potentially stall / crash the build server or local development workstation.

It should probably be a configurable value from _greenwood.config.js_.

thescientist13:
## Related Issue
resolves #16 but seeing if / hoping it might help with #213 ? 🤷‍♂ 

## Summary of Changes
1. Maintain one instance of `puppeteer` while serializing all pages

> Interestingly enough I thought this would be a big performance boost, but running `yarn build` didn't seem to improve all that much but maybe it will be faster in CI?  It will least it help with this issue though
multple-puppeteers


thescientist13:
resolved as part of #217

Issue #15, opened 2019-03-27
As a user / developer I should have verbose (and quiet) output logging


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Logging can be very useful, as long as it can be controlled.  In my mind, there should be two options minimum
- "normal": output basic status / progress logging
- "verbose": outputs more debug level logging, useful for CI and testing environments 

## Details

1. Provide a `--verbose` option that can be used to enable advanced logging capabilities
1. Introduce a logging library to toggle between info / debug vs. warning and error level logging
1. Add relevant logs where applicable, and at the appropriate level
1. Should include proper error logging / handling / `process.exit`
1. Update README with documentation on `--verbose` flag

thescientist13:
I think good default logging should be enough plus can always use `--trace` with NodeJS I think, so probably not a big need or priority anytime soon

PR #217, opened 2019-10-02
Fixing Errors With Consistent Unit Tests

### LINKED ISSUE #213: same unit test failing consistently, inconsistently on CircleCI

hutchgrant:
## Related Issue
Resolves #213 (+ #16 )

## Summary of Changes

1. ~~Disabled webpack plugins. Tests weren't running properly to begin with because the polyfills plugin is breaking serialization~~
2. ~~Hardcoded polyfills to where they were previously in index.html template~~
3. Utilized fs-extra where possible for asynchronous file functions, mainly in test-bed.js and in lifecycles(scaffold.js, serialize.js)
4. Modified `browser.js` to fix puppeteer launch error related to opening too many instances

Ran test locally 5 times, within the same docker container greenwood is tested with on circle ci, after reproducing issue.  

I suggest the issue was async problems with our test bed, within our lifecycles, and further the new plugin system is potentially flawed.  See serialize.js and webpack.config.common.js.


packages/cli/test/cases/build.config.error-workspace-absolute/build.config.error-workspace-absolute.spec.js
```diff
@@ -1,18 +1,18 @@
...
  * User Command
  * greenwood build
- * 
+ *
  * User Config
  * {
  *   workspace: path.join(__dirname, 'noop')
  * }
- * 
+ *
```
thescientist13: are these coming from your IDE by chance?  It might be good to turn that feature off (for this project anyway).  All the formatting is unfortunately adding a lot a noise to this PR.

packages/plugin-google-analytics/test/cases/default/default.spec.js
```diff
@@ -1,116 +1,116 @@
...
+    before(async function() {
       await setup.runGreenwoodCommand('build');
     });
-    
+
     runSmokeTest(['public', 'index', 'not-found', 'hello'], LABEL);
 
-    describe('Initialization script at the end of the <body> tag', function() {
-      let inlineScript;
+    // describe('Initialization script at the end of the <body> tag', function() {
```
thescientist13: instead of commenting out like this, in a test, to help with readability and review from a PR perspective, I would request using `xdescribe` (or `xit`) for disabling test code.  Less line changes / diffs and makes it clearer and cleaner that a test or suite is being ignored.

hutchgrant: Ah that was it, I forgot the syntax at that moment.

packages/cli/test/cases/build.config.error-workspace-absolute/build.config.error-workspace-absolute.spec.js
hutchgrant: Yea, but what's the point of having spaces there anyway?

hutchgrant:
The polyfills plugin was a major issue for stability to begin with. Now I don't think file functions should be synchronous, what if its a network share or something that takes time to access.  Not everyone has the resources for immediate file access at a given moment. While I agree, it could work synchronously, it's _safer_ to do file loop up and read asynchronously. Especially when we do not know the size of some of these files, or number of them that we are compiling.

packages/cli/test/cases/build.config.error-workspace-absolute/build.config.error-workspace-absolute.spec.js
thescientist13: hmm... that wasn't really the concern though.

> All the formatting is unfortunately adding a lot a noise to this PR.

thescientist13:
> The polyfills plugin was a major issue for stability to begin with.

Hmm, ok.  so do we feel that narrows it down?  Is it the only issue or just a contributing factor?  I'm not as sure now after seeing this report though
https://github.com/thescientist13/greenwood-getting-started/issues/8

> Now I don't think file functions should be synchronous, what if its a network share or something that takes time to access ....

I'm not trying to make anyone's life hard to be clear, but from a build perspective, having things more synchronous could add a bit more predictability to the process as a whole, _especially_ as it comes to reading, compiling, and writing files.  

I'm sure not everything has to be synchronous but I think the things we decide to keep / make async have to be 100% guaranteed to contained / idempotent / side effect free.  I don't think doing a full audit of the build pipeline would hurt to see if there's an intelligent way to approach it.

That said, I can see the argument for making everything `async / await` but again, just have to make sure there's no chance of unpredictability.

Reliability and repeatability has to be the ultimate goal here.  Run after run after run the same input should always yield the same output.  That's what i really care about the most with this.

hutchgrant:
We dont have the specs from thescientist13/greenwood-getting-started#8 not enough information to determine if related.

In any event, I have demonstrated its reliable and consistant now as it stands in this PR. Once you add the plugins thats where the errors come in. So if we want to narrow it down from there further, we might find your answer.

packages/cli/test/cases/build.config.error-workspace-absolute/build.config.error-workspace-absolute.spec.js
hutchgrant: two birds with one stone?

thescientist13: Again, that wasn't really my concern.  I will try and explain on the call instead if we have time.

thescientist13:
> not enough information to determine if related.

well, In all fairness, the same kind of error happened in the ground zero PR.
```shell
5) Build Greenwood With: 
       Custom Webpack Plugin and Default Workspace
         Running Smoke Tests: Custom Webpack Plugin and Default Workspace
           Hello World (dummy) page
             "before each" hook for "should output a hello page directory":
     Error: ENOENT: no such file or directory, open '/root/repo/packages/cli/test/cases/build.plugins-webpack/public/hello/index.html'
```

I'm not suggesting that the plugins could have unbalanced things, but still not 💯 sure if the root cause has been there from the beginning.

packages/plugin-google-analytics/test/cases/default/default.spec.js
hutchgrant: Oh right, the problem with this is that it will fail with exit code 1 still with each of the xdescribe tests listed as "pending".

hutchgrant: `describe.skip()` works though

hutchgrant:
I've readded the webpack analytics plugin and I've reverted all assosciated tests.

I noticed one issue that cropped up, that was likely due to launching multiple instances of puppeteer. I've rewritten our browser lib as a class so that we can reuse the same browser instance for every single page.  This should eliminate the puppeteer error.

Last thing to do it readd polyfills plugin to see if error is still appearing or if we've completely solved the problem of consistent unit tests.

thescientist13:
Thanks, will take a look again.  There's a lot going on here, and with all the formatting changes a bit harder to focus on it specifically.

So are there two distinct bugs like we talked about?  A `fs` related bug and a serialization related bug?  Are we able to identify with any specificity the root cause of either on their own by chance?

hutchgrant:
It was a number of `fs` bugs with lifecycles(scaffold and serialize), one in the implementation of the polyfills plugin, and a puppeteer bug that randomly occurs due to launching of too many instances.

package.json
```diff
@@ -26,10 +26,13 @@
...
     "chai": "^4.2.0",
     "eslint": "^5.15.3",
-    "lerna": "^3.16.4",
     "jsdom": "^14.0.0",
+    "lerna": "^3.16.4",
     "mocha": "^6.1.4",
     "nyc": "^14.0.0",
     "rimraf": "^2.6.3"
+  },
+  "dependencies": {
```
thescientist13: I can fix this up, but this will need to go into the CLI's _package.json_.  For Lerna repos, to install dependencies per package, you just go to into the root folder of that package where _package.json_ is and just run the command, e.g
```shell
$ yarn add fs-extra
```

packages/cli/src/lib/browser.js
```diff
@@ -1,12 +1,28 @@
...
-  });
-  const renderer = new Renderer(browser);
-
-  return await renderer.serialize(url);
-};
\ No newline at end of file
+class BrowserRunner {
+
+  constructor() {
+    this.browser = '', this.renderer = '';
```
thescientist13: minor, but let's put each variable declaration on its own line.  Also, I think using a `''` is misleading.  Probably just better left `undefined`  or maybe `{}`

Also, what would your thoughts be on folding renderer right in browser?  Doesn't seem all that useful keeping them apart and then you don't have to pass down instances of `browser`?  Just a thought.

packages/cli/src/lifecycles/scaffold.js
```diff
@@ -23,49 +23,61 @@ const writePageComponentsFromTemplate = async (compilation) => {
...
+  };
+
+  const writeComponentToFile = async (target, filename, result) => {
+    return new Promise(async(resolve, reject) => {
+      try {
+        await fs.ensureDir(target, { recursive: true });
+        await fs.writeFile(path.join(target, `${filename}.js`), result);
+        resolve();
       } catch (err) {
         reject(err);
```
thescientist13: should we add `console.error` before all `reject`?  Seems coming out of this PR, more logging couldn't hurt (for us and users), at least until we are able to implement something like #15

packages/cli/src/lifecycles/scaffold.js
```diff
@@ -23,49 +23,61 @@ const writePageComponentsFromTemplate = async (compilation) => {
...
       }
     });
   };
 
+  const determineTarget = (file, context) => {
```
thescientist13: what is target in this case?  Do we mean output directory / path?  In NodeJS land, target may be more associated with platform, like browser, nodejs, electron, etc.

I think if what we're doing is more filesystem related, we should stick that more familiar taxonomy / language where possible.

packages/cli/test/cases/build.config.error-workspace-absolute/build.config.error-workspace-absolute.spec.js
thescientist13: So we didn't get to it on the call, but just to emphasize the point of focus was on the _diff_ introduced by the formatting changes, regardless of it was an addition / removal.  

In general (and yes, maybe we have been lax on formatting, maybe we should add [prettier](https://prettier.io/) to the project?) but in the face of an important PR like this (bug, large feature or refactoring) formatting changes purely for the sake of formatting changes are discouraged.  It adds "noise" which is that changes presented are not actually changes to have to worry about, yet it still counts as extra files and changes to the diff that get in the way of what is really important and what a reviewer would _want_ to read, is all.  This PR would be at least few files less and definitely many lines less if just the formatting was removed, which I think speaks for itself.

Always up for consistent formatting, and as a standalone PR that is definitely doable.  So just to politely ask please turn off any auto formatting (for this project) in your IDE so these kinds of changes don't have to keep happening.  Thanks!!! ✌️

test/test-bed.js
```diff
@@ -25,44 +24,70 @@ module.exports = class TestBed {
...
+              const targetDir = path.join(cwd, file.dir);
+              const targetPath = path.join(cwd, file.dir, file.name);
+
+              await new Promise(async(resolve, reject) => {
+                try {
+                  await fs.ensureDir(targetDir, { recursive: true });
+                  await fs.copy(targetSrc, targetPath);
+                  resolve();
+                } catch (err) {
+                  reject();
```
thescientist13: we should probably pass the `err` to the `reject()`

packages/cli/src/lifecycles/scaffold.js
hutchgrant: This specific thrown error is printed back in the main [index.js](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/index.js#L79) already.  If you follow all thrown errors they should be printed back at that same location, if not then of course we need the error output, but from what I can tell additional outputs within a catch clause elsewhere would be redundant.

packages/cli/src/lifecycles/scaffold.js
hutchgrant: Fair point, it's vague.

test/test-bed.js
hutchgrant: Yes good _catch_

packages/cli/src/lib/browser.js
hutchgrant: Yea just initializing, good/bad habit.

thescientist13: good habit for sure, I think setting it as an empty string may just be a little misleading is all?  

That said, not sure what closest primitive of an instance of `browser` or `renderer` would be, so 🤷‍♂

thescientist13: maybe `null`?  idk, not a huge deal

test/test-bed.js
thescientist13: > good  _catch_

this guy.... 🤣

packages/cli/src/lifecycles/scaffold.js
```diff
@@ -84,24 +96,24 @@ const writeListImportFile = async (compilation) => {
   const appDir = path.join(compilation.context.scratchDir, 'app');
 
   if (!fs.existsSync(appDir)) {
-    await fs.mkdirSync(appDir);
+    await fs.mkdirs(appDir);
```
thescientist13: did we miss a `Sync` on L89?

hutchgrant:
This PR was in regards to stable tests which was a problem due to multiple
bugs, so multiple bugs were solved and tested at once. It's easy to say
after the fact that it should be more narrow, but part of the bug was with
the polyfills plugin so I tested against your most recent solution for it.
I argue it's causing more work to seperate it only to add it again.

thescientist13:
>  This PR was in regards to stable tests which was a problem due to multiple
bugs, so multiple bugs were solved and tested at once.

I understand the goal.  But what does code _not part of the bug_ in question have to do with fixing the bug in question?

>  It's easy to say after the fact that it should be more narrow, but part of the bug was with
the polyfills plugin so I tested against your most recent solution for it.

which again, does _not_ include the changes submitted as part of #212 so I'm not sure the validity of this argument


> I argue it's causing more work to seperate it only to add it again.

that's fine, I will do it myself since the value is not in the convenience to us, as has always been my point each time this topic (continues) to come up.

hutchgrant:
Well each bug issue has an automatic reference link to this PR since you
cited them. I don't see the value in separating, removing, and redoing it
for a different PR just for the sake of having one issue per PR.

It's possible, while not encouraged, that a single PR can tackle multiple
issues at once epecially one that was definetly contained related
issues(polyfill implementation).

thescientist13:
I understand these nuanced discussions never seem to translate well over GitHub unfortunately but I know we've talked about them at length over calls and Slack so I'll just defer to all the same reasons I've given before for why I care about efficient PRs and git histories. 

Just try and understand beyond the surface level why I care about things like avoiding unnecessary formatting and mixing issues into single PRs.  In this case, the scope was trying to bring stability to our codebase, in the current version (0.3.x) and the WIP version for 0.4.0.  I appreciate your enthusiasm for wanting to take on all the things but please try and exercise restraint and discipline when submitting work that _others have to read, review, and test_ and be willing to work towards a commit history for the project that is clean and easy to manage and more importantly, easy to go back over.  Linking all the issues isn't the first thing you see when reviewing git history, it's the branch names, which is the whole point of putting issues in the branch name, and thus why it's my preference to keep 1 issue:1 PR so changes aren't just magically showing up in the codebase without a clean history.

Anyway, this PR is shaping up nice now, thank you for making the PR review feedback suggestions.  I'll get to doing a final review / test on those  and back out the changes to #212 so that it can stand as its own discrete change into our release branch and overall project history.

0.4.0 here we come!!

thescientist13:
Also, to be clear, #9 is just an enhancement on _top_ of the polyfills plugin, to use the loader _instead_ of the single bundle file.  So everything about the polyfills plugin is already in the release branch.   We could ship without #212 and still have a polyfills plugin, it would just be less efficient at runtime.

thescientist13:
@hutchgrant 
Did you have any thoughts on either of these comments I left by chance?  I think these should be considered as items for tracking.

### [_.greenwood/_ directory](https://github.com/ProjectEvergreen/greenwood/pull/217#pullrequestreview-300385867)
> One thought: Is the output to .greenwood/ anything we need to worry about between builds? Should we be blowing that out between each run? 🤔


### [browser + renderer](https://github.com/ProjectEvergreen/greenwood/pull/217#discussion_r333762162)
> Also, what would your thoughts be on folding renderer right in browser? Doesn't seem all that useful keeping them apart and then you don't have to pass down instances of browser? Just a thought.

PR #223, opened 2019-10-15
Website/issue 222 Shelf List Items

### LINKED ISSUE #222: website pages with a shelf component are showing console errors when there are no subitems
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Several bugs related to the items[] array within our shelf if no items present.  Solution is to make sure array is initialized and to use a map.  These errors are viewable at https://www.greenwoodjs.io/getting-started/

## Details
shelf-console-error

```
Uncaught TypeError: Cannot read property 'length' of undefined
Uncaught TypeError: Cannot read property 'map' of undefined
Uncaught TypeError: Cannot read property 'findIndex' of undefined
```

hutchgrant:
## Related Issue
Resolves #222 

## Summary of Changes

* Added shelf expand and render conditions if shelf contains items


www/components/shelf/shelf.js
```diff
@@ -51,21 +51,25 @@ class shelf extends LitElement {
...
+        this.requestUpdate();
+      }
     }
   }
 
   collapseAll() {
-    for (let i = 0; i < this.shelfList.length; i = i + 1) {
-      this.shelfList[i].selected = false;
+    if (this.shelfList) {
+      for (let i = 0; i < this.shelfList.length; i = i + 1) {
```
thescientist13: I would recommend a `forEach` here.  

Also, could we just initialize `shelfList` to an empty array?  Then we could eliminate all the multiple checks
```javascript
if (this.shelfList) {
  // do stuff
}
```

hutchgrant: lol yes.

thescientist13: I only use `map` if i need to return a new array.  Otherwise, `forEach` is my goto for non mutating changes.

www/components/shelf/shelf.js
```diff
@@ -120,11 +121,11 @@ class shelf extends LitElement {
 
       return listItems;
     };
-    /* eslint-enable */
 
+    /* eslint-enable */
```
thescientist13: Do we actually need this `/* eslint */`?  What was it disabling?  I think I should make another issue because we might have a few of these we don't need anymore.

Plus, I would prefer we only enable it per rule, if we do need an exception.  Like I said, maybe another issue, but curious since it's showing in the diff.

www/components/shelf/shelf.js
```diff
@@ -64,9 +64,10 @@ class shelf extends LitElement {
   }
 
   collapseAll() {
-    for (let i = 0; i < this.shelfList.length; i = i + 1) {
-      this.shelfList[i].selected = false;
-    }
+    this.shelfList = this.shelfList.map(item => {
+      item.selected = false;
```
thescientist13: ah, maybe a `map` is needed then...

www/components/shelf/shelf.js
hutchgrant: Yes it's required, you'll note the section:

```
        listItems = html`
          <ul>
            ${list.items.map(item => {
              return html`
                <li id="index_${item.index}" class="${list.selected ? '' : 'hidden'}"><a @click=${()=> this.goTo(`#${item.id}`)}">${item.name}</a></li>
              `;
            })}
          </ul>
        `;
```

Requires it to be disabled because of indentation.  Then below that function we re-enable eslint

thescientist13: AH, right, that thing...

PR #225, opened 2019-10-16
technical/issue 176 Workspace 404 not building and serializing

### LINKED ISSUE #176: 404 page not building and serializing correctly
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Observed a couple issues with 404 page, the default anyway.  Observed from [my own project](https://github.com/thegreenhouseio/www.thegreenhouse.io/issues/112) but also in getting started companion repo.  Probably exists in _www/_ too.

**develop**
1. `<title>` is default value, _My App_.  Not sure if there could be more missing config items, like `meta` or others?
Screen Shot 2019-08-19 at 5 29 38 PM

**build**
1. It doesn't seem to exist at all?
Screen Shot 2019-08-19 at 5 29 33 PM

## Details
Maybe this is more of a need around documentation as well?  Not 100% clear what the expectations are.  Do users make their own 404.md?  

> **Would be better to do this instead > #200**

### LINKED ISSUE #210: hook plugins not working when using greenwood develop
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Observed when working on #208 that hook plugins don't seem to be showing up when running `greenwood develop` because `getOutboundLink` was undefined.
Screen Shot 2019-09-24 at 8 47 34 PM

## Details
Comparing the output, it's clear these custom hooks are not getting included.  However our custom fallback handling script is there.  🤔 

#### develop
Screen Shot 2019-09-24 at 8 52 07 PM

#### build
Screen Shot 2019-09-24 at 8 53 28 PM

hutchgrant:
## Related Issue
Resolves #176 

Side Effect, Fixes #210 

## Summary of Changes

* removed default development 404 redirect scripts 
* added `http-server` instead of `ws` for testing production builds
* added additional custom 404 route detection and scaffold
* added `postSerialize` lifecycle specifically for copying and overriding our default 404.html with a custom 404 route
* added a default `404.md` to website workspace.

With this change, you can now override the default 404 by placing a `404.md` in your /pages directory within your workspace.  Previously you could do the same with regular .html file, but it wouldn't render in development. This way the 404 is testable in development and production.  Maybe there's a more elegant solution but I like the idea of adding a 404 to any of our page templates as we do with regular markdown pages.

As a happy addition, this also fixes #210.

If we move forward with this method, we should add tests and documentation for this feature.


packages/cli/package.json
```diff
@@ -32,6 +32,7 @@
     "fs-extra": "^8.1.0",
     "glob-promise": "^3.4.0",
     "html-webpack-plugin": "^3.2.0",
+    "http-server": "^0.11.1",
```
thescientist13: technically it would only be a `devDependency`

hutchgrant:
develop_210

#210 fixed, in development, probably related to the removal of those redirect scripts. Not positive.

thescientist13:
closing in favor of #240

hutchgrant:
no as I said before it cannot use ws for testing production builds locally

hutchgrant:
There is no specific fix for #210 again as discussed, it's liekly due to to the removal of the redirect scripts that we're inserting in development via webpack.

PR #237, opened 2019-10-16
Remove Peer Dependency Warnings For Greenwood Plugins

### LINKED ISSUE #228: peerDependencies warning from plugins packages
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
When doing a fresh install in [my blog project]() with these dependencies
```json
"@greenwood/cli": "0.4.0",
"@greenwood/plugin-google-analytics": "0.4.0",
"@greenwood/plugin-polyfills": "0.4.0",
```

Getting these warning on install
```shell
warning " > @greenwood/plugin-google-analytics@0.4.0" has incorrect peer dependency "@greenwood/cli@^0.3.6".
warning " > @greenwood/plugin-polyfills@0.4.0" has incorrect peer dependency "@greenwood/cli@^0.3.6".
```

## Details
Not sure if something is pinned or if this is a lerna setting?  I would think the usage of `^` in our packages would support anything <= 1.0.0

hutchgrant:
## Related Issue
Resolves #228 

## Summary of Changes
* Change peer dependency for @greenwood/cli to be `>= 0.3.6`


thescientist13:
Did some googling, looks like this supported by `lerna` automatically, in that it will automatically bump the version on each release, but only if the `peerDependency` is also a `devDependency`
https://github.com/lerna/lerna/issues/955#issuecomment-323260292.

So maybe we could try reverting this change and just adding the same dependency to `devDependecnies` too?  This would preferable since ideally I would like the version to change with each new released published.

PR #238, opened 2019-10-18
Remove Redux From Shelf

### LINKED ISSUE #142: remove redux store code and dependencies (where applicable) from shelf component
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
In the shelf component for the website, there is [unused code connecting the component to a redux store](https://github.com/ProjectEvergreen/greenwood/blob/master/www/components/shelf/shelf.js#L18) that can be removed.

## Details
1. Remove the store connection code
1. Remove `redux` related dependencies (thunk?)
1. Remove other reference to `store.dispatch`
1. Should also be able to remove **pwa-helpers** as well (confirm is it can be removed from the `cli` package)

hutchgrant:
## Related Issue
Resolves #142 

## Summary of Changes

* Remove Redux From Shelf
* Fix tiny broken link in getting-started list

### DIFF
--- a/www/components/shelf/getting-started-list.json
+++ b/www/components/shelf/getting-started-list.json
@@ -17,7 +17,7 @@
       "id": "configuring-workflows"
     }, {
       "index": 2,
-      "name": "Project Structures",
+      "name": "Project Structure",
       "id": "project-structure"
     }]
   }, {
--- a/www/components/shelf/getting-started-list.json
+++ b/www/components/shelf/getting-started-list.json
@@ -18,7 +18,7 @@
     }, {
       "index": 2,
       "name": "Project Structures",
-      "id": "project-structures"
+      "id": "project-structure"
     }]
   }, {
     "index": 2,
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -1,9 +1,5 @@
 import { html, LitElement } from 'lit-element';
 import css from './shelf.css';
-import { applyMiddleware, createStore, compose as origCompose, combineReducers } from 'redux';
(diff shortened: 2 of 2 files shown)


www/components/shelf/getting-started-list.json
```diff
@@ -18,7 +18,7 @@
     }, {
       "index": 2,
       "name": "Project Structures",
-      "id": "project-structures"
+      "id": "project-structure"
```
thescientist13: was it intentional to change the `id`?  Doesn't the `name` need to match to the `id` for in page scrolling?

hutchgrant: the id has to match the header yes. That's why I changed it. It doesn't work currently.

thescientist13: oh, the heading in the document.  got it.  So what is the `name` for?  If we're making everything else singular, shouldn't we change `name` too?

hutchgrant: Name is for the text of the shelf link. We can also have be structure(singular) if you'd like but it has no affect on the link.

thescientist13: I don't think it makes sense to have them be different then?  The left nav is one way, but the heading is another?  I would see that as a typo / grammar error.  

Can we make them all the same?

hutchgrant: Singular or plural? Which is the more accurate heading name?  If plural, we can revert this change and fix the getting-started page for project-structure(s).  If singular, I need to add singular to the name of this shelf link.

thescientist13: i think it was intended to be _Project Structure_ (singular)

Issue #143, opened 2019-08-07
make index optional for shelf component


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
The [shelf component](https://github.com/ProjectEvergreen/greenwood/blob/master/www/components/shelf/shelf.js) takes "shelf" objects like [seen in the _.json_ files](https://github.com/ProjectEvergreen/greenwood/tree/master/www/components/shelf) also in the shelf component directory, that each define an `index` property, [for example](https://github.com/ProjectEvergreen/greenwood/blob/master/www/components/shelf/about.json)
```json
[
  {
    "index": 0,
    "name": "Goals",
    "path": "/about/goals"
  },
  {
    "index": 1,
    "name": "Features",
    "path": "/about/features"
  },
  {
    "index": 2,
    "name": "Community",
    "path": "/about/community"
  }
]

```
## Details
Given the above data structure, it would be great if index was optional, such that shelf component could just go by the order of the objects passed in, instead.

PR #239, opened 2019-10-22
resolve website issue 143, remove index dependence for the shelf rend…

aholtzman:
resolve website issue 143, remove index dependence for the shelf render from the json data.

## Related Issue

https://github.com/ProjectEvergreen/greenwood/issues/143


aholtzman:
Yeah @thescientist13 I use this technique all the time, it does not use an index or id number from the json (if they are there they are ignored). We can confirm it by removing it from the json and printing the index out...

thescientist13:
I think I see what you mean now, clever!   I wasn't sure if the `index` was part of some sort of sorting logic.  @hutchgrant can you confirm what `index` is used for?

thescientist13:
@aholtzman 
Can you edit all those _.json_ files too?  If `index` isn't being used by any business logic for sorting the shelf items in the component itself, and just for this output, then we should remove it altogether.  (including in the _.json_ files too)

aholtzman:
will do @thescientist13

hutchgrant:
I'm noticing another issue with shelf, I don't think it's related to this PR and I don't know when it was introduced. The shelf will open and collapse, but refuses to open again.

Issue #124, opened 2019-07-13
create reusable components for chevron icons in the shelf


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
In the [`Shelf`](https://github.com/ProjectEvergreen/greenwood/blob/master/www/components/shelf/shelf.js#L8) component, the expand / collapse arrow icons (chevrons) are SVGs but are inline with the code, ex.
```javascript
const chevronRt = html`<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 512">
<path d="M285.476 272.971L91.132 467.314c-9.373 9.373-24.569 9.373-33.941 0l-22.667-22.667c-9.357-9.357-9.375-24.522-.04-33.901L188.505
256 34.484 101.255c-9.335-9.379-9.317-24.544.04-33.901l22.667-22.667c9.373-9.373 24.569-9.373 33.941 0L285.475 239.03c9.373 9.372 9.373
24.568.001 33.941z"/></svg>`;
```

## Details
It would be nice to move each of these chevrons into their own standalone component that can then be instead `import`ed into the `Shelf` component.

PR #243, opened 2019-10-23
ISSUE #124 - refactor chevron svgs by removing them from the shelf component into…

aholtzman:
… a svg component for importing

## Related Issue

https://github.com/ProjectEvergreen/greenwood/issues/124


aholtzman:
easy enough, just thought it might keep the components from getting to cluttered as we intend to do the same with the social icons.

thescientist13:
> easy enough, just thought it might keep the components from getting to cluttered as we intend to do the same with the social icons.

Good point... maybe something like
```
www/components/icons/
   chevron-down/chevron-dowm.js
   chevron-right/chevron-left.js
```

So at least they are all grouped, but still individual to group individual styles / tests / etc needed on a a case by case basis?

aholtzman:
that makes sense  - TK shortly

thescientist13:
Nice @aholtzman , thanks!!

Issue #245, opened 2019-10-23
scollIntoView tossing error on the console when clicking sub menu link


aholtzman:
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary

scollIntoView tossing error on the console when clicking sub menu link

Steps:
1. Click _Getting Started_
1. Click _Project Setup_
1. Click _Installing Greenwood_

## Details

const elements = this.querySelectorAll(selector); is a NodeList, needs to be an Array
hash = hash.replace('-', ' '); only works on the first one, problem for multiple instances of the -

PR #246, opened 2019-10-23
fix erros in handleHashChange() and add page load listener to scroll component

aholtzman:
## Related Issue

https://github.com/ProjectEvergreen/greenwood/issues/245


www/components/scroll/scroll.js
```diff
@@ -19,18 +19,21 @@ class scroll extends LitElement {
     // on hash change child event, scroll content
     const contains = (selector, text) => {
       const elements = this.querySelectorAll(selector);
-
-      return Array.prototype.filter.call(elements, (element) =>{
+      //convert NodeList to Array
+      const elementsArray = Array.from(elements);
+      return Array.prototype.filter.call(elementsArray, (element) =>{
```
thescientist13: I think if we are going to use `.from`, then we also don't need to use `Array.prototype` since i think they were aiming to do the same thing?

So I think we could simply to this maybe?
```javascript
return Array.from(elements).filter(element) => {
  // do stuff
}
```

aholtzman:
Screenshot 2019-10-24 09 31 51
there seems to be an issue when switching between sub menus - I will look into this further when I can. Let me know any hints or ideas if you are aware of them.

www/components/scroll/scroll.js
```diff
@@ -19,22 +19,29 @@ class scroll extends LitElement {
...
 
     let { hash } = window.location;
 
     // clean hash string, remove # and replace - with spaces
-    hash = hash.replace('#', '');
-    hash = hash.replace('-', ' ').toLowerCase();
-
+    hash = hash.replace('#', '').toLowerCase();
+    // for cases of multiple hyphens
+    while (hash.indexOf('-') !== -1) {
```
hutchgrant: str.replace(a, b) will replace all instances of a with b. There is no need to loop it.

www/components/scroll/scroll.js
```diff
@@ -19,22 +19,29 @@ class scroll extends LitElement {
...
+      hash = hash.replace('-', ' ');
```
hutchgrant: str.replace(/-/g, "");  will replace all instances of '-'

www/components/scroll/scroll.js
```diff
@@ -8,33 +8,40 @@ class scroll extends LitElement {
   connectedCallback() {
     super.connectedCallback();
     window.addEventListener('hashchange', this.handleHashChange.bind(this), false);
+    window.addEventListener('load', this.handleHashChange.bind(this), false);
```
hutchgrant: :+1:

www/components/scroll/scroll.js
```diff
@@ -8,33 +8,40 @@ class scroll extends LitElement {
...
 
   handleHashChange() {
     // on hash change child event, scroll content
     const contains = (selector, text) => {
       const elements = this.querySelectorAll(selector);
+      // convert NodeList to Array
+      const elementsArray = Array.from(elements);
 
-      return Array.prototype.filter.call(elements, (element) =>{
+      return Array.prototype.filter.call(elementsArray, (element) => {
```
thescientist13: Do we still think we can make this refactor possible?
https://github.com/ProjectEvergreen/greenwood/pull/246#discussion_r338226468

PR #247, opened 2019-10-24
website bug/issue 244 sub-nav expand/collapse

### LINKED ISSUE #244: sub nav not opening in shelf component after open/close
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Bug

## Summary

when a menu item with a sub nav on the shelf is clicked it opens the sub nav (as expected), when clicked again it closes (as expected) but if clicked a third time it does nothing (expected to open sub menu). If after opening and closing a different menu item is clicked, it will open when clicked again (as expected).
## Details

no errors present in the console

hutchgrant:
## Related Issue
Resolves #244

## Summary of Changes
* Fixed selectedIndex on initial page load
* Fixed immediate expand on page load if shelf item has sub-nav

Page was refreshing when we clicked a shelf item that contained a submenu, this was resetting the list index which caused an issue with lifecycle logic.  We now initialize the selectedIndex when menu is expanded on page load.

Page wasn't comparing the path correctly. Locally, when I click a shelf item it resolves to `http://localhost:8000/getting-started/project-setup` for example.  In production, on netlify, the path is being modified to end with a `/` for example: `http://localhost:8000/getting-started/project-setup/`.

You can manually test this by clicking a menu shelf link, refresh page, seeing it stayed expanded, then collapse and re-expand and refresh page.


www/components/shelf/shelf.js
```diff
@@ -32,13 +30,21 @@ class shelf extends LitElement {
   }
 
   expandRoute(path) {
+    // remove / if at the end of the path in order to match our shelf paths
+    const lastPathCharacter = path.length - 1;
+
+    if (path.lastIndexOf('/') === lastPathCharacter) {
+      path = path.substring(0, lastPathCharacter);
```
thescientist13: actually, is there any way to do something like this instead of the `lastPathCharacter` check?
```
    let routeShelfListIndex = this.shelfList.findIndex(list => {
      return list.path.indexOf(path) >= 0;
    });
```

www/components/shelf/shelf.js
```diff
@@ -15,8 +15,6 @@ class shelf extends LitElement {
 
   constructor() {
     super();
-    this.selectedIndex = 0;
```
thescientist13: would still want to keep a default value, or is that part of the bug?  I see below we set it to empty string
```
if (this.selectedIndex === previousSelected) {
  this.selectedIndex = '';
}
```

hutchgrant: Yea it's redundant apparently

thescientist13: My comment was more towards keep the initialization though, for `this.selectedIndex`.  Did it need to be removed to fix the bug?  Otherwise, it's generally best practice to create / initialize your members.

Issue #197, opened 2019-09-17
development mode not working with Microsoft Edge


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
When running the `develop` task and loading a project up with Edge, there is an error in the console saying `customElements is not defined`.
<img width="1217" alt="edge-develop-custom-elements-not-define" src="https://user-images.githubusercontent.com/895923/65084659-ddfb1280-d979-11e9-9405-4b251649106e.png">

## Details
Something to do with polyfills I'm sure and dev vs prod serialization.  Lower priority though.

thescientist13:
Awesome, it looks like that by solving #210 and having **HtmlWebpackPlugin** working correctly in develop and production mode, Edge will now work in `develop` mode.
<img width="1243" alt="Screen Shot 2019-11-10 at 4 54 34 PM" src="https://user-images.githubusercontent.com/895923/68551439-fa945380-03da-11ea-9d2a-b89fc4d92b40.png">

Issue #199, opened 2019-09-20
Enable a development / production mode flag


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As might be the case with certain plugins are other parts of Greenwood's build pipeline, having a `production` flag that is set during the `config` lifecycle based on whether it is a `develop` or `build` CLI command.

## Details
A good use case would be disabling the Analytics plugin when developing, and thus only enabling it when building for production.
```js
module.exports = (options = {}) => {
  const { analyticsId, anonymous } = options;

  const validId = analyticsId && typeof analyticsId === 'string';
  const trackAnon = typeof anonymous === 'boolean' ? anonymous : true;

  if (!validId) {
    throw new Error(`Error: analyticsId should be of type string.  get "${typeof analyticsId}" instead.`);
  }

  return [{
    type: 'index',
    provider: (compilation) => {
      const hook = !compilation.config.production
        ? ''
        : `
            <script async src="https://www.googletagmanager.com/gtag/js?id=${analyticsId}"></script>
          
            <script>
              window.dataLayer = window.dataLayer || [];
              function gtag(){dataLayer.push(arguments);}
              gtag('js', new Date());

              gtag('config', '${analyticsId}', { 'anonymize_ip': ${trackAnon} });
              gtag('config', '${analyticsId}');
            </script>
          `
      };
     
     return {
       hookGreenwoodAnalytics: hook
     }
  }];
};
```

thescientist13:
As part of #355 this was simply done as an environment variable, but ideally this should be a little more ergonomic / less mutable.

thescientist13:
We could build off this after we do #1393

Issue #47, opened 2019-04-18
add tests for develop and serve commands


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
Although the `develop` tasks does start `webpackDevServer`, it would be good to see how much of it could actually be testable.
![Screen Shot 2019-07-13 at 5 28 17 PM](https://user-images.githubusercontent.com/895923/61176712-ab0c5b00-a593-11e9-90c4-5cd5fa4a89cd.png)

## Details
If we can figure out a way to work around the dev server without influencing the actual running of `greenwood` too much, we should try, if at least for the basic `generateCompilation` step and that the command works as expected.

thescientist13:
Should try and expand coverage as much as possible to at least get everything within the thresholds.

thescientist13:
So although we've moved away from webpack, the challenges presented still apply in trying to do the testing in that starting a server will hang the node / mocha process.  Luckily I have added support for this workflow in https://github.com/thescientist13/gallinago/issues/19 !

So now we can test a server and make sure command specific behaviors actually work as expected, like
- import maps
- live reload
- etc

thescientist13:
Made some good progress in #670 , but still a couple TODOs so going to leave this open to put into the next project to try and tack another crack at the `server` command + `mode` specs that I would like to add (and GraphQL client!)
https://github.com/ProjectEvergreen/greenwood/pull/670#issuecomment-885908945

thescientist13:
Also make sure to better handle 404 routes from develop, as per the test logs you'll see this
```sh
Develop command with default 404 behavior
(node:17870) UnhandledPromiseRejectionWarning:
  Error: ENOENT: no such file or directory, open '/abc.js'


(Use `node --trace-warnings ...` to show where the warning was created)
(node:17870) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

PR #250, opened 2019-10-30
Bug/issue 210 hooks plugins not working

### LINKED ISSUE #210: hook plugins not working when using greenwood develop

thescientist13:
## Related Issue
resolves #210  (and #197 🌟 )

## Summary of Changes
1. Instances of `HtmlWebpackPlugin` were overwriting each other.  Instead, just explicitly instantiating the instances that we need in **webpack** config common, and then we can override the configuration options per **webpack** config as needed.

~~Note: As expected, there is some duplication but not sure the best way to abstract it, without it being clunky.  Maybe in [_cli/lifecycles/config.js_ for `plugins`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/config.js#L14)?~~

### DIFF
--- a/packages/cli/src/config/webpack.config.common.js
+++ b/packages/cli/src/config/webpack.config.common.js
@@ -39,18 +39,6 @@ module.exports = ({ config, context }) => {
     from: context.assetDir,
     to: path.join(context.publicDir, 'assets')
   }] : [];
-
-  // gets Index Hooks to pass as options to HtmlWebpackPlugin
-  const customOptions = Object.assign({}, ...config.plugins
-    .filter((plugin) => plugin.type === 'index')
-    .map((plugin) => plugin.provider({ config, context }))
-    .filter((providerResult) => {
-      return Object.keys(providerResult).map((key) => {
-        if (key !== 'type') {
-          return providerResult[key];
-        }
-      });
-    }));
       
   const commonCssLoaders = [
     { loader: 'css-loader' },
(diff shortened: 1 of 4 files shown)


hutchgrant:
I agree the plugin filter + map function should be exported from a single file less duplication.

packages/cli/src/config/webpack.config.develop.js
```diff
@@ -29,6 +28,30 @@ module.exports = ({ config, context, graph }) => {
   const { devServer, publicPath } = config;
   const { host, port } = devServer;
 
+  // decorate HtmlWebpackPlugin instance with devServer specific SPA handling for index.html
+  configWithContext.plugins[0].options.hookGreenwoodSpaIndexFallback = `
```
thescientist13: Another ➕ 1 for trying to figure out #47

packages/cli/src/config/webpack.config.common.js
```diff
@@ -114,6 +114,20 @@ module.exports = ({ config, context }) => {
     },
 
     plugins: [
+      new HtmlWebpackPlugin({
+        filename: path.join(context.publicDir, context.indexPageTemplate),
+        template: path.join(context.scratchDir, context.indexPageTemplate),
+        chunksSortMode: 'dependency',
+        ...customOptions
```
thescientist13: Would like to look into something like #199 so that for plugins like Analytics, it could be disabled for development, and only enabled for production builds.

Issue #115, opened 2019-07-07
[RFC] Build Time Data Injection


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As a precursor to something like #21 , being able to pass data from build time to runtime is a critical API for Greenwood to provide.

## Details
A good example of where this could be useful now is in how the items and subitem are configured for the Shelf component.  

Currently the pages and links are determined by hardcoding a list of items and subitems in a JSON file in _www/templates/page-template.js_
```javascript
setupShelf() {
  // based on path, display selected list
  const path = window.location.pathname;

  if (path.substring(0, 5) === '/docs') {
    shelfList = require('../components/shelf/documentation-list.json');
  } else {
    shelfList = require('../components/shelf/getting-started-list.json');
  }
}

...

<eve-shelf .shelfList="${shelfList}"></eve-shelf>
```

So an example of how Greenwood could help with this is by being able to pass information from the build, like the graph.  

```javascript
import { graph } from 'xxx';

setupShelf() {
  // map over the graph and create the shelfList object dynamically
  graph.pages.map((page) => {
    // xxx
  } 
}
```

Crossing this membrane is essentially how eventually external data and plugins can operate and provide dynamic information at build time, for the run time.

thescientist13:
Would also be good to be able to define data via frontmatter, e.g.
```md
---
template: 'blog'
date: '2019-09-01'
title: 'Blog Post Title'
---
```

Would help a lot in this project and would allow more (all?) of the logic to happen in the template
https://raw.githubusercontent.com/thegreenhouseio/www.thegreenhouse.io/technical/issue-96-greenwood-migration/src/pages/blog/2018/01/24/index.md

PR #252, opened 2019-11-10
commit filesystem change git keeps setting

thescientist13:
## Related Issue
keep getting this from `git` to change the file permissions.  is anyone else getting this?  Seems like it showed up #115 too

## Summary of Changes
1. Accept `git`s preferred file permissions for this file

I should compare settings for other files, just to confirm though.  Some more reading on the topic
https://stackoverflow.com/questions/14557106/git-changes-my-file-permissions-upon-checkout


hutchgrant:
I too have been seeing this.

thescientist13:
I just ran all the build tasks, `yarn lint`, `yarn build`, `yarn serve` and `yarn lint` and none of them changed the file permissions.

I'm not sure what causes this then but it must be something we're both doing.  I don't think merging this really helps us understand this though, so will leave it open a little bit more to see if we can figure it out.

thescientist13:
@hutchgrant 
Found the culprit, it's `yarn install`
```shell
$ rm -rf node_modules/
Owens-MBP-2:greenwood owenbuckley$ git status
On branch bug/issue-256-fix-babel-configuration

nothing added to commit but untracked files present (use "git add" to track)

$ yarn install
yarn install v1.12.3
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning "workspace-aggregator-898da660-27e5-480c-9d0e-4238686e8733 > @greenwood/cli > css-to-string-loader@0.1.3" has incorrect peer dependency "css-loader@>=0.25.0 <1.0.0".
warning "workspace-aggregator-898da660-27e5-480c-9d0e-4238686e8733 > @greenwood/cli > glob-promise@3.4.0" has unmet peer dependency "glob@*".
warning "workspace-aggregator-898da660-27e5-480c-9d0e-4238686e8733 > @greenwood/cli > lit-redux-router@0.9.13" has unmet peer dependency "lit-html@1.x.x".
[4/4] 📃  Building fresh packages...
✨  Done in 16.57s.

$ git status
On branch bug/issue-256-fix-babel-configuration
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   packages/cli/src/index.js

no changes added to commit (use "git add" and/or "git commit -a")
```

wonder if it has something to do with this configuration in _package.json_?
```json
"main": "./packages/cli/src/index.js",
```

thescientist13:
Looks like this actually ["snuck in"](https://github.com/ProjectEvergreen/greenwood/commit/c4be9a4ffc03856ec9fbd14fea77f61567b345bd#diff-11b04b2d23165cdb1a5aec0369a0b8d9) as part of #257.

In reflecting on it some more now, I think it actually makes sense that it should be treated as an executable since we are actually [using it that way](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/package.json#L18), to be fair.  For that reason, it would make sense it should be flagged as [`0755`](http://permissions-calculator.org/decode/0755/), instead of [`0644`](http://permissions-calculator.org/decode/0644/).

In fact, GitHub helps attribute that distinction as well, fwiw.
Screen Shot 2020-02-04 at 8 30 41 PM

Given that, let's close this then, I'm not sure it's worth forcing this to 0644, to be honest and it makes more sense the way it is.

PR #253, opened 2019-11-11
conslidate browser and render lib files

### LINKED ISSUE #231: consolidate browser and renderer into one file
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Just posing this as question coming out of our work in #213 
https://github.com/ProjectEvergreen/greenwood/pull/217#issuecomment-541325788

## Details
It seems that they share the same responsibilities and now that we're only using one instance of `puppeteer` as part of #16 , having them be in two files seems to granular.

thescientist13:
## Related Issue
resolves #231 

## Summary of Changes
1. Consolidate _lib/renderer.js_ into _lib/browser.js_ to keep related concerns together


packages/cli/src/lib/browser.js
```diff
@@ -1,5 +1,14 @@
+/*
+* Rendertron - Modified
+* Repo: https://github.com/Goo./rendererndertron
```
hutchgrant: should fix this with the correct repo https://github.com/GoogleChrome/rendertron

thescientist13: oh right!  good catch

PR #255, opened 2019-11-12
Technical/issue 229 docker documentation for local development

thescientist13:
## Related Issue
resolves #229 

## Summary of Changes
1. Created a _Dockerfile_ for local development 
1. Documented in _CONTRIBUTING.md_

Question: I observed in some of our specs we have different ways to do top level `describe`.  
1. [`describe('Build Greenwood With: ', async function() {`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.default/build.config.default.spec.js#L20)
1. [`describe('Build Greenwood With: ', () => {`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.error-public-path/build.config.error-public-path.spec.js#L22)
1. [`describe('Build Greenwood With: ', function() {`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.workspace-custom/build.config.workspace-custom.spec.js#L29)

Is there a "canonical" way we should do it?  Seems like given our history here, having a single way or documented pattern would be useful.  I can make them all the same as part of this PR if that helps.

PR #259, opened 2019-11-15
have all specs use same describe callback convention

thescientist13:
## Related Issue
related to #213 , but observed as part of #255 

## Summary of Changes
Noticed in the PR for #255 that our tests weren't consistent in `describe` conventions for specs
1. ~~[`describe('Build Greenwood With: ', async function() {`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.default/build.config.default.spec.js#L20)~~
1. ~~[`describe('Build Greenwood With: ', () => {`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.error-public-path/build.config.error-public-path.spec.js#L22)~~
1. [`describe('Build Greenwood With: ', function() {`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.workspace-custom/build.config.workspace-custom.spec.js#L29)

I don't think they need `async` and should all probably use `function()` since mocha has issues with `() => `

----

Went with option 3 and also made other usages of `async` consistent by removing them if not used with an `await`.

Updated Contributing docs.


packages/cli/test/cases/build.default.workspace-getting-started/build.default.workspace-getting-started.spec.js
```diff
@@ -34,7 +34,7 @@ const { JSDOM } = require('jsdom');
 const path = require('path');
 const TestBed = require('../../../../../test/test-bed');
 
-describe('Build Greenwood With: ', function() {
+describe('Build Greenwood With: ', async function() {
```
thescientist13: this one should remove the `aync`

thescientist13:
> and I recall mocha had issues with lambda functions.

Ah yes, good call, that one definitely caught us off guard in the beginning.  

Essentially, it's just a matter of knowing when (not) to use it, which is basically if your test relies on needing access to `this` or not.  I will add a note about this and link to (it in) [the mocha docs](https://mochajs.org/#arrow-functions) in CONTRIBUTING.md

thescientist13:
Hmm, this is a new one
```
  1) Build Greenwood With: 
       Default Greenwood Configuration and Workspace w/Custom Page Template
         "before all" hook in "Default Greenwood Configuration and Workspace w/Custom Page Template":
     Error: the string "(node:1075) UnhandledPromiseRejectionWarning: #<ErrorEvent>\n" was thrown, throw an Error :)
      at process._tickCallback (internal/process/next_tick.js:68:7)
```

Failed on v10, but not v12.  Let me test locally.

thescientist13:
Removing all instances of arrow functions from specs.  Everything will just `function` now.  See if that makes a difference.

PR #263, opened 2019-11-21
setting nodejs engines in npm

### LINKED ISSUE #180: set minimum node version in package.json
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
_package.json_ can support identifying a minimum version of Node required when installing the project, e.g.
```json
"engines": {
  "node": ">= 6.9.0 < 13.0.0"
},
```

## Details
Would be good to set this to help provide clear feedback to users at install time.  Recommendation would be to go with the current LTS 10.x, though I think 12 as LTS may be coming soon?

thescientist13:
## Related Issue
resolves #180 

## Summary of Changes
On the home page for NodeJS we say we support latest LTS NodeJS, but CI was only running against 10.x.
1. Set CI Action to run against 10.x and 12.x versions
1. Set minimum NodeJS version for `engines` in _package.json_ for `@greenwood/cli`
1. Set Master Action to build using 12.x (LTS)

**Question**
Should we set an upper bound, or just a minimum?  I think minimum is better since CI will tell us if we introduce a breaking change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,14 +10,14 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x]
+        node: [10, 12]
 
     steps:
     - uses: actions/checkout@v1
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v1
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: ${{ matrix.node }}
     - name: Installing project dependencies
       run: |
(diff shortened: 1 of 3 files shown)


thescientist13:
Yup, now it is showing multiple builds per change 🎉 
Screen Shot 2019-11-20 at 10 09 02 PM

thescientist13:
oh, there was an issue for this!  😄 
https://github.com/ProjectEvergreen/greenwood/issues/180

PR #264, opened 2019-11-24
fix filename and directory name overlap breaking the build

### LINKED ISSUE #132: filenames that include existing directory name in the filename causes compilation error
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
If you name a file `creating-pages.md` in the pages directory, it will throw an error that the file wasn't found.

## Details
I think this error is related to a regex when we're looking for files within the `pages` directory.

edit: Same bug also we cannot use the word `templates` (with an s) within a filename.  It seems `pages` and `templates` are reserved words.

thescientist13:
## Related Issue
resolves #132  (see https://github.com/ProjectEvergreen/greenwood/issues/132#issuecomment-557913939 for additional insights on the issue)

## Summary of Changes
1. Added a `/` to the regex pattern used with `webpack.NormalModuleReplacementPlugin` to distinguish files from directories
1. Added specs 
1. Little refactoring / comments

## TODO
~~The main issue is that while the problem is solved for _www/_  (yay) I can't get the tests to fail though for the same use case (and without the fix in place) 😕~~
<details>
To reproduce, remove the `\/` added to the Regex pattern in _webpack.config.commons.js_
- `yarn build` should fail, and does ✅ 
- `yarn test` should fail, and does _not_ ❌
</details>

---

1. [x] Figure out how to get tests to reproduce the use case from the _www/_ folder
1. [x] Handle double overlap case, e.g.
    ```javascript
    import service from '../../services/pages/pages.js';
    ```
1. [x] Handle _pages/pages.md_ case
1. [x] Clean up comments / logging / files / tests
1. [x] Misc refactor items
1. [x] Documentation update (e.g. undo https://github.com/ProjectEvergreen/greenwood/pull/254)


thescientist13:
>  I can't get the tests to fail though for the same use case (and without the fix in place) 

Oh, yeah.... probably because I don't have an actual _pages/_ directory in my workspace 🤦‍♂

hutchgrant:
If I rename the about.md within the directory mapping test case to `pages.md` it will fail.

If I rename the about.md within the directory mapping test case to `creating-pages.md`, as per the initial issue, it will succeed.

Almost there.

thescientist13:
Cool, thanks for helping to test!  Probably just need to make a small tweak to the `Regex` constructor.

hutchgrant:
Finally users can add a `pages/templates/page-templates.md` without the dang regex breaking.

Well done.

PR #265, opened 2019-11-27
Route Based Code Splitting

### LINKED ISSUE #136: [RFC] Route Based Code Splitting
## Type of Change
- [x] New Feature Request (RFC)
- [ ] Documentation / Website
- [ ] Bug

## Summary
As it stands now it looks like Greenwood only creates on _index.js_.  Would like to look into how we can possibly do route based code splitting, or I guess in the context of this project, template based code splitting?
Screen Shot 2019-09-18 at 7 59 54 AM

## Details
In particular this is a potential concern until larger projects can be tested because if ALL the content / code ends up in _index.js_ then it will absolutely crush larger projects even though currently the Greenwood website scores well.  

Ideally we should be able to configure this all through webpack.
https://webpack.js.org/plugins/split-chunks-plugin/

We should also want to handle / consider splitting out vendor code into it's own chunk too.

hutchgrant:
## Related Issue
Resolves #136 

## Summary of Changes
1. Modify scaffold to build lit-redux routes using lazy loaded dynamic imports
2. ~~Add babel-eslint~~
3. ~~Enable `@babel/plugin-syntax-dynamic-import` plugin in babel config~~
4. Modify webpack config to set chunk file names
5. Fix app-template test case
6. Add loading component and spinner


hutchgrant:
You'll notice in the www example that since we're placing our header and footer in our page-template, we lose the SPA feel on route change with lazy loaded routes. Perhaps it might be a better idea to implement a custom app-template as part of the www as well.

Thoughts?

hutchgrant:
Added a basic loading component w/spinner

loading

hutchgrant:
Just a note. Given the size of our pages, I don't know if this is necessary at the moment. Though it might be a nice to have feature.

The bundle size on master is 61kb and with this PR it drops 27kb, but it has to make an extra request for an additional 5-7kb for each page that isn't the home page.  While you do reduce the bundle size by 45%, assuming you only view one single page, is it really that necessary given the small size of our pages?  And if you view more than 5-6 pages, it will actually be slower for you, than without chunks.

package.json
```diff
@@ -24,12 +24,14 @@
     "test:tdd": "yarn clean && mocha --watch ./packages/**/test/**/**/*.spec.js --timeout 15000"
   },
   "devDependencies": {
+    "babel-eslint": "^10.0.3",
```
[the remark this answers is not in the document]
hutchgrant: It's required so that lint doesn't throw an error immediately within the editor. Yes I should move it to the package specifically.

packages/cli/package.json
```diff
@@ -29,6 +29,7 @@
     "@babel/plugin-transform-runtime": "^7.6.2",
     "@babel/preset-env": "^7.6.0",
     "@babel/runtime": "^7.7.2",
+    "@greenwood/cli": "^0.4.2",
```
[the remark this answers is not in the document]
hutchgrant: Yea I removed some packages and added them, don't know how this got in there.

packages/cli/src/config/webpack.config.common.js
```diff
@@ -87,6 +87,7 @@ module.exports = ({ config, context }) => {
     output: {
       path: path.join(context.publicDir, '.', config.publicPath),
       filename: '[name].[hash].bundle.js',
+      chunkFilename: '[name].bundle.js',
```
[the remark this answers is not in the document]
hutchgrant: of course.

packages/cli/src/lifecycles/scaffold.js
```diff
@@ -89,30 +89,21 @@ const writePageComponentsFromTemplate = async (compilation) => {
...
-};
-
 const writeRoutes = async(compilation) => {
   return new Promise(async (resolve, reject) => {
     try {
       let data = await fs.readFile(compilation.context.appTemplatePath, 'utf8');
 
       const routes = compilation.graph.map(file => {
-        return `<lit-route path="${file.route}" component="eve-${file.label}"></lit-route>\n\t\t\t\t`;
+        return `<lit-route path="${file.route}" component="eve-${file.label}" .resolve="\${() => import(${file.relativeExpectedPath})}" loading="eve-loading"></lit-route>\n\t\t\t\t`;
```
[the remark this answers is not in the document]
hutchgrant: That would be cool, good point. Why would a developer just want my basic spinner right?

package.json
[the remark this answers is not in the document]
thescientist13: > It's required so that lint doesn't throw an error immediately within the editor. 

Not sure I understand what the editor has to do with our code though?  Does it actually cause  `eslint` to fail?

Maybe it would be better to provide a screenshot or the error message so I can see what the actual issue is, not sure this dependency is needed, at least not for the CLI since linting really shouldn't have any impact on our source code per se.

www/components/loading/spinner.js
```diff
@@ -0,0 +1,38 @@
...
+
+  static get properties() {
+    return {
+      size: {
+        type: String
+      }
+    };
+  }
+
+  render() {
```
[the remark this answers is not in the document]
hutchgrant: Yea that's a custom feature of lit-redux-router you're talking about. It doesn't do progressive loading.  It displays the loading component while its waiting to resolve. You could just put the loading component on a timer if you want. That's the simplest way.

hutchgrant:
The lint errors were related to the dynamic-import-plugin which appparently is already included in our dependencies. The only way I found to get around that problem was to add a .eslintignore file and specified all node_modules/ files.

The app-template header/footer methodology is similar to any one page app. You want to keep the header/footer consistent, while the content is the only thing that lazyily loads. This is better UX IMO. Otherwise the entire page goes blank every single click and it defeats the purpose of an SPA.

package.json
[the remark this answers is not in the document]
hutchgrant: yea it was highlighting the dynamic import, but I forgot I needed to make that string anyway so it doesn't matter actually, that entire dep is now not needed.

thescientist13:
> The app-template header/footer methodology is similar to any one page app. You want to keep the header/footer consistent, while the content is the only thing that lazyily loads. This is better UX IMO. Otherwise the entire page goes blank every single click and it defeats the purpose of an SPA.

For sure, and totally understand.  I was just under the impression we were already doing something like that with this, re: `<entry>`
```
<div class='wrapper'>
  <eve-header></eve-header>

    <div class="content">
      <entry></entry>
    </div>

  <eve-footer></eve-footer>
</div>
```

But I do see now that is more of mechanism of our scaffolding, and not so much the runtime (e.g. routing), where as you point out, in a SPA, you would have something like:
```html
<header></header>
    
  <router-outlet></router-outlet>

<footer></footer>
```
With the router outlet being more a fragment / encapsulated component's HTML for _just_ page content.

An interesting distinction / challenge I see now.  Will think about some of this myself as I continue to familiarize myself with working on #115 as well, since we are now seeing similar architectural needs converging at once on this.  (as you anticipated).

hutchgrant:
What if we set code splitting as toggleable from the greenwood config, include a default loading component, allow that to be overriden the same way we do app-template? We'd have to add logic to the context and config lifecycles for that, but it's doable.

hutchgrant:
I've modified it to split into 4 main chunks, 1 for each of our main categories, as well as an additional vendor.js chunk. Obviously this only works for our website but I've done this as an example because you mentioned it would be cool to combine the pages based on the route.  

The tests failed because now there is an additional script element. Do you want to keep the vendor.js bundle?

packages/cli/src/config/webpack.config.common.js
```diff
@@ -125,6 +126,60 @@ module.exports = ({ config, context }) => {
...
+        maxAsyncRequests: 6,
+        maxInitialRequests: 4,
+        automaticNameDelimiter: '~',
+        automaticNameMaxLength: 30,
+        cacheGroups: {
+          about: {
+            test(module, chunks) {
+              const regexp = /about/i;
+              const result = `${module.name || ''}`.match(regexp) || chunks.some(chunk => `${chunk.name || ''}`.match(regexp));
+
```
hutchgrant: Yes it's hardcoded into 4 cache groups, but ideally we could generate `cacheGroups` based on the graph.  This is utilizing our newly generated chunk names `page--[mainroute]--[page-name].js` which we create when the graph is filled.

hutchgrant: Or better yet, why don't we just move this to the `greenwood.config.js` and allow people to set their own chunks based on our generated graph chunk names.

greenwood.config.js
```diff
@@ -20,6 +20,61 @@ module.exports = {
...
+    splitChunks: {
+      chunks: 'all',
+      minSize: 30000,
+      maxSize: 0,
+      minChunks: 1,
+      maxAsyncRequests: 6,
+      maxInitialRequests: 4,
+      automaticNameDelimiter: '~',
+      automaticNameMaxLength: 30,
+      cacheGroups: {
```
thescientist13: yeah, We should be doing this in _webpack.config.common.js_ for the user by default.

And also do some sort of dynamic creation of `cacheGroups` here like we do in _webpack.config.common.js_ with `NormalModuleReplaclementPlugin`.

packages/cli/src/lifecycles/config.js
```diff
@@ -11,6 +11,7 @@ let defaultConfig = {
   publicPath: '/',
   title: 'Greenwood App',
   meta: [],
+  optimization: {},
```
thescientist13: this should not be a public API (yet), we'll think of something but I want to keep this under the hood

greenwood.config.js
hutchgrant: I was thinking along the lines of similar to a plugin, if a developer wants custom split routes they can configure it however they please.

packages/cli/src/lifecycles/scaffold.js
```diff
@@ -89,30 +89,21 @@ const writePageComponentsFromTemplate = async (compilation) => {
 
 };
 
-const writeListImportFile = async (compilation) => {
```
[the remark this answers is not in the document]
thescientist13: Does this [get rid of `MDIMPORT`](https://github.com/ProjectEvergreen/greenwood/issues/322)!?

www/components/loading/spinner.js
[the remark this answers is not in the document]
thescientist13: I think let's back this work out then and open it as a separate PR for us to discuss further in, just so we can evaluate on its own and then it makes this PR a bit "cleaner" for review purposes.  I think there is something interesting here, but let's take it to another thread.  👍

www/templates/app-template.js
```diff
@@ -0,0 +1,106 @@
...
+        header.replaceChild(meta, oldmeta);
+      } else {
+        header.appendChild(meta);
+      }
+    });
+  }
+
+  render() {
+    return html`
+      <div class='wrapper'>
```
thescientist13, line 96: Seems a shame we need a custom _app-template.js_ just for a few lines of HTML.  🤔 

I wonder if we could inherit from Greenwoood's `AppComponent` and just override just the `render` method somehow???  That would be pretty slick...

www/templates/app-template.js
```diff
@@ -0,0 +1,106 @@
...
+        <eve-header></eve-header>
+        MYROUTES
```
thescientist13, line 98: Is `MYROUTES` new?  Would be a shame to have add another "magic constants" when we're trying to move away from that (`MDIMPORT`, `META****`... any other thoughts on how to approach this?

packages/cli/src/config/webpack.config.common.js
```diff
@@ -138,6 +138,8 @@ module.exports = ({ config, context }) => {
       }]
     },
 
+    optimization: config.optimization,
```
thescientist13, line 141: let's back out exposing this to the user for now
https://github.com/ProjectEvergreen/greenwood/pull/265/files#r355764479

packages/cli/src/lifecycles/graph.js
```diff
@@ -76,6 +76,16 @@ const createGraphFromPages = async (pagesDir, config) => {
                 // set <title></title> element text, override with markdown title
                 title = title || '';
 
+                // create webpack chunk name based on route and page name
+                const routes = route.lastIndexOf('/') === route.length - 1 && route.lastIndexOf('/') > 0
```
thescientist13, line 80: what is this conditional expressing?  Maybe better to put it into a variable to improve readability?

packages/cli/src/lifecycles/graph.js
```diff
@@ -76,6 +76,16 @@ const createGraphFromPages = async (pagesDir, config) => {
...
+                  ? route.substring(1, route.length - 1).split('/')
+                  : route.substring(1, route.length).split('/');
+                let chunkName = 'page';
+
+                routes.map(subDir => {
+                  chunkName += '--' + subDir;
```
thescientist13, line 86: if we're not actually mapping anything, a `forEach` is probably more appropriate.  And we can also use a template string literal as well
```js
let chunkName = '';

routes.forEach(subDir => {
  chunkName = `page--${subDir}`;
});
```

Also, if this generating a lit-route (presumably) how come we are getting an array of routes?

packages/cli/src/templates/app-template.js
```diff
@@ -59,7 +58,7 @@ class AppComponent extends LitElement {
     meta.forEach(metaItem => {
       const metaType = metaItem.rel // type of meta
         ? 'rel'
-        : metaItem.name 
+        : metaItem.name
```
thescientist13, line 61: i get the page was updated, but unnecessary formatting.

www/templates/app-template.js
hutchgrant: MYROUTES isn't new

thescientist13: ah yes, i missed that one.  thanks for catching me on that 👍

thescientist13: I will add it tracking on that issue then.  Hopefully we can find a "clean" way around those as well.

packages/cli/src/lifecycles/scaffold.js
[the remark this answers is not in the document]
hutchgrant: No it gets rid of the ./list.js which is a list of components to import in the app-template

packages/cli/src/lifecycles/graph.js
hutchgrant: No weren't getting an array, we're stepping through a route and getting each directory to put it as part of the chunk name.  We're splitting the route by `/` and then placing it into one string with `chunk += 'page--' + route`

e.g. /docs/markdown  ===   page--docs--markdown

packages/cli/src/lifecycles/graph.js
hutchgrant: becausewedwouldhavevariablenamesthislong

thescientist13: I only asked a question 🤷

packages/cli/src/lifecycles/graph.js
thescientist13: Gotcha, thanks for clarifying.

And just to follow up, is there anything meaningful to this this string format, per se?
```
page--docs--markdown
```

Or just a convention you came up with?

packages/cli/src/lifecycles/graph.js
hutchgrant: I mean I originally wrote that 6 months ago and was still able to understand it fairly quickly.

packages/cli/src/lifecycles/graph.js
hutchgrant: just a convention I came up with, I think I've seen something similar elsewhere.

packages/cli/src/lifecycles/graph.js
thescientist13: Yeah, I'm just trying to think more of how this would be read / understood by someone else coming to the project fresh who hasn't spent a year+ working on it, or how you would explain the code to someone else as you walked through it.  Length or mild verboseness shouldn't really be considered as some sort of gatekeeping metric, to be fair, but rather the value it adds to the overall expressiveness of the code.  Mapping logic to context (variables) can thus help reinforce the intent / readability of the code so much so that it almost comes out as natural language (or at least is always my goal).

When the code itself reads like a book, then you never really need comments, unless it's maybe capture the why (specific business rule / case, environment constraint, etc), as opposed to the what.

Another factor to my comment is also more that we have a lot of similar logic like this around our code [[1](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/graph.js#L45)], [[2](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/scaffold.js#L45)], like in our resolvers and serialize code that I think we have come to accept over time.  I have a card in Trello where I had wanted to make a more technical document for ourselves and contributing developers that would explain the architecture of the project, and how it is organized overall.  I think a key outcome of that exercise will be a better understood taxonomy (or in Domain Driven Design, they would call it a ubiquitous language) which would then allow us to better express these (at first glance) arbitrary bits of logic, and / or could lead to a more refined implementation of Greenwood itself as we get more awareness of what we're trying to achieve.

I'll add this convo as notes to that card.

thescientist13: hah.  there was already a note for this exact topic in the card already 🤓

packages/cli/src/lifecycles/scaffold.js
```diff
@@ -71,30 +71,25 @@ const writePageComponentsFromTemplate = async (compilation) => {
...
     try {
       let data = await fs.readFile(compilation.context.appTemplatePath, 'utf8');
 
       const routes = compilation.graph.map(file => {
-        return `<lit-route path="${file.route}" component="eve-${file.label}"></lit-route>\n\t\t\t\t`;
+        return `<lit-route
+          path="${file.route}"
+          component="eve-${file.label}"
+          .resolve="\${() => import(/* webpackChunkName: "${file.chunkName}" */ ${file.relativeExpectedPath})}"
+          ></lit-route>\n\t\t\t\t`;
```
thescientist13, line 84: what's the `\n\t\t\t\t` for?  And foe the first `$` need to be escaped?

hutchgrant: just for code formatting, not necessary to function.

thescientist13: Not sure I follow.  Formatting in the rendered HTML?   Not sure that has ever really been a concern?

www/templates/app-template.js
```diff
@@ -23,6 +23,13 @@ connectRouter(store);
 
 class AppComponent extends LitElement {
 
+  static get styles() {
```
thescientist13, line 26: 👌

packages/cli/src/lifecycles/scaffold.js
thescientist13: Ah, I see.  


Screen Shot 2020-06-03 at 8 49 40 PM
It is the difference between this (with)


and this (without)
Screen Shot 2020-06-03 at 8 49 49 PM

I don't think it's a big deal, and we can probably omit it.  I think if anything, we should / could use a html minfier on the final output just save even more bytes.  I'll handle removing the formatting and tracking html minification.

PR #266, opened 2019-12-20
testing refactoring for build matrix

### LINKED ISSUE #180: set minimum node version in package.json

thescientist13:
## Related Issue
resolves #180 

## Summary of Changes
1. Refactor GitHub Actions to appropriately take advantage of Node Matrix based on [this example](https://github.com/ianwalter/puppeteer). 🤞

### DIFF
--- a/.github/workflows/chromium-lib-install.sh
+++ b/.github/workflows/chromium-lib-install.sh
@@ -3,7 +3,7 @@
 sudo apt-get update \
   && sudo apt-get install -yq libgconf-2-4 \
   && sudo apt-get install -y wget --no-install-recommends \
-  && sudo wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
+  && sudo wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add - \
   && sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
   && sudo apt-get update \
   && sudo apt-get install -y google-chrome-unstable --no-install-recommends \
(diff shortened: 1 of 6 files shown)


thescientist13:
Interesting.  Noticed an odd failure for a spec, which had been refactored as part of #259 
https://github.com/ProjectEvergreen/greenwood/runs/357556860
```
 331 passing (3m)
  1 failing

  1) Build Greenwood With: 
       Empty Configuration and Default Workspace
         "before all" hook in "Empty Configuration and Default Workspace":
     Error: the string "TypeError: this.browser.newPage is not a function\n" was thrown, throw an Error :)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

But even removing `async` didn't get it working.  Not sure why the `puppeteer` call is failing just for that one spec?  🤔

hutchgrant:
I'm going out on a limb here but here's my hypothesis.

`this.browser` was not initialized because in our puppeteer render library we're call async init() from constructor. So this.browser, within init() doesn't execute in time before the rest of the thread. 

*lib/browser.js*
```js
  constructor() {
    this.browser = {}, this.renderer = {};
    this.init();
  }
  async init() {
    this.browser = await puppeteer.launch({
      headless: true,
      args: ['--no-sandbox']
    });
    // this.renderer = new Renderer(this.browser);
  }

```

thescientist13:
> I believe this is the result of the merge between the 2 puppeteer render libraries we were previously using

good call, will look back to see if this is the issue, but either way your point about `async` usage here seems a little clunky, so a tweak here would seem worthwhile either way.

thescientist13:
Good call @hutchgrant , looks like our matrix is running tests successfully now!  👏 👏 👏 

Gonna start calling you the Async King 👑

Issue #271, opened 2020-01-17
query results are not deterministic


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
One of the current drawbacks of the implementation in #115 is that data is returned in the order it is on disk, which is essentially alpha-sorted.  However, like for the Getting Started docs on our website, this navigation needs to be in a certain order.

## Details
And also observed is that order is not guaranteed either, [possibly due to `JSON.stringify`](https://github.com/ProjectEvergreen/greenwood/pull/269#pullrequestreview-347305483).

For that reason, it should get fixed and put under test, before adding the sorting and filtering.

thescientist13:
[Observed](https://github.com/ProjectEvergreen/greenwood/pull/269#pullrequestreview-347305483) as part of #115 that not only should user's be able to provide a filter for results, but that at the very least, the default order should _always_ be deterministic each and every time, at minimum.

So as part of this, we should make sure that we can guarantee basis results order (like the blog posts test cases) and then we can add additional enhancements on top of that if needed.

hutchgrant:
The correct way to do this IMO is through the graphql schema and resolver.
https://www.howtographql.com/graphql-js/8-filtering-pagination-and-sorting/

For our built in schemas, we can add predetermined sorting/filtering capabilities.

thescientist13:
I think I should just fold #280 into this, since this is such a critical expectation of this feature, it should include unit tests too for not only default determinism, but also put it under test.

hutchgrant:
I would prefer to merge navigation query with shelf query into one single menu query which uses links and submenus.  Navigation is basically a menu, the only difference is it has no submenus.  We can create a map of mensu(with recursive menus and links) and then filter it based on the menu request.  Currently within markdown we can get the menus based on front-matter during the graph lifecycle(#274).

e.g.

Some page.md
```
---
label: 'getting-started'
menu: navigation
title: 'Getting Started'
---
## Introduction
First off, thank you for taking the time to check out Greenwood! 

```

schema
```
  type Menu {
    self: Link,
    links: [Link]
    children: [Menu]
  }

  type Link {
    label: String,
    link: String
  }
```

resolver:
```
  type Query {
    menu(filter: String): Menu
  }
```

our query from our components(both header and shelf):

```
query($menu: String!) {
  menu(filter: $menu) {
[...]

PR #285, opened 2020-02-07
[WIP] task/issue-275 Create Extensible Menus Query

### LINKED ISSUE #275: add support for dynamic imports
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Currently the shelf uses `require` to load JSON files.
```javascript
setupShelf() {
  let list = [];

  if (url.indexOf('/about') >= 0) {
    list = require('../components/shelf/about.json');
  } else if (url.indexOf('/docs') >= 0) {
    list = require('../components/shelf/documentation-list.json');
  } else if (url.indexOf('/getting-started') >= 0) {
    list = require('../components/shelf/getting-started-list.json');
  } else if (url.indexOf('/plugins') >= 0) {
    list = require('../components/shelf/plugins.json');
  }

  this.shelfList = list;
}
```

## Details
This is only permitted because **webpack** allows it, but browser code should only be using browser APIs, like `import` / `export`.

The project should be made to support `import()` syntax and remove any usage of NodeJS specific code, like `require` in the shelf.

hutchgrant:
## Related Issue
Resolves #275

## Summary of Changes

1. Begins to address (#271) which requires deterministic queries through a single menu based query
2. Removes need for shelfList to use json file
3. Uses [markdown-toc](https://github.com/jonschlinkert/markdown-toc) to generate a table of contents from every page flagged with front-matter variable `linkheadings: true`
4. Added markdown front-matter globals for `menu, index, linkheadings` which are added to our graph via the graph lifecycle
5. if `linkheadings` is flagged true, during graph lifecycle(as markdown file is already being read) we will parse the file for all headings, add it to our graph for that specific page, into an array we will use for our shelf later.
6. the `menu` front-matter is used to specifically define a menu to use with any given markdown file e.g. `navigation` or `side`
7. I've modified our graphql  type defs to add `link` and `menu` 
8. I've added filtering to the query to only display listing for specific menus. e.g. `side` or `navigation`
9. I've modified our resolver to instead look through graph for any pages with the corresponding menu name
10. as per 9, I've also made the resolver check if a file has been flagged `listheadings` if so, it will loop through the `tableOfContents` variable for that page, and create a list of children for that specific link.

## TODO
1. shelf is not rendering the results correctly at the moment, this is important and is the next step, but the data however is being received correctly for all menus
2. fix broken collapse/render that was dependent on previous json file

## Note
At the moment this WIP. Data is received correctly, for all menus, only need to render it.

### DIFF
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -1,6 +1,6 @@
 import { LitElement, html } from 'lit-element';
 import client from '@greenwood/cli/data/client';
-import ChildrenQuery from '@greenwood/cli/data/queries/children';
+import MenuQuery from '@greenwood/cli/data/queries/menu';
 import css from './shelf.css';
 import chevronRt from '../icons/chevron-right/chevron-right';
 import chevronDwn from '../icons/chevron-down/chevron-down';
@@ -29,21 +29,17 @@ class Shelf extends LitElement {
   }
 
   async setupShelf(page) {
-    console.log('setupShelf for page =>', page);
-
     if (page && page !== '' && page !== '/') {
-      // TODO remove require call - #275
-      this.shelfList = require(`./${page}.json`);
-
-      // TODO not actually integrated, still using .json files - #271
       const response = await client.query({
-        query: ChildrenQuery,
+        query: MenuQuery,
         variables: {
-          parent: page
+          menu: 'side',
+          route: window.location.pathname
         }
       });
 
-      console.log('response from the shelf (data.children)', response.data.children);
+      console.log('shelf =>', response.data.menu.children);
+      this.shelfList = response.data.menu.children;
     }
   }
 
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -22,14 +22,9 @@ class Shelf extends LitElement {
     this.page = '';
   }
 
-  connectedCallback() {
+  async connectedCallback() {
     super.connectedCallback();
-    this.collapseAll();
-    this.expandRoute(window.location.pathname);
-  }
-
-  async setupShelf(page) {
-    if (page && page !== '' && page !== '/') {
+    if (this.page !== '' && this.page !== '/') {
       const response = await client.query({
         query: MenuQuery,
         variables: {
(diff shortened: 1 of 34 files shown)


hutchgrant:
I thought we were removing need for the menu json file altogether. This way it's done on a page by page basis determines what page should be in what menu, and if necessary, it will add all headings as a table of contents within the graph for the page.

www/components/shelf/shelf.js
```diff
@@ -22,29 +22,24 @@ class Shelf extends LitElement {
...
-  }
-
-  async setupShelf(page) {
-    console.log('setupShelf for page =>', page);
-
-    if (page && page !== '' && page !== '/') {
-      // TODO remove require call - #275
-      this.shelfList = require(`./${page}.json`);
-
-      // TODO not actually integrated, still using .json files - #271
```
hutchgrant: This line is what I'm referring to when I resolved 271, using my method.  You knew we were going to use a query here.

thescientist13:
Well the details of #275 are pretty clear I think, in that specifically we shouldn't be using NodeJS APIs (`require`) in browser code, where we should be using Browser APIs (`import()`).
> This is only permitted because webpack allows it, but browser code should only be using browser APIs, like import / export

I think per something I saw in #265 , this would require additional Babel configuration.  I can rename the issue if it helps clarify the intent but that was the outcome I was looking for from this issue.

For this PR then and as with most things, I still recommend that it all new ideas / enhancements goes through an issue workflow first ideally, to allow for technical design review / decision and share code snippets and APIs.  Given the amount of moving parts here, I think this warrants it and then the issue links and branch names can be reflective of the intended scope / issue.

hutchgrant:
You specifically put a query, because the idea is to query our menu through graphql not import or require. There's no need to import or require that json file because its redundant with a working grpahql query.  That's the point I think you're missing here. Read the rest of your code for the component.

thescientist13:
I understand the point, I wrote the issue.  :)

like I said, I'll just rename the issue **because** the main point was supporting `import()`, regardless of the use case.  You make it seem like no one could ever need to use it, ever which is shortsighted.  Also because `require` never should have been used in the first place.

My opinion on issue preceding PRs still stands and would / should apply here though, it is the way.

hutchgrant:
closing this issue and i'll resubmit it

PR #286, opened 2020-02-12
issue/275 Add Support For Dynamic Imports

### LINKED ISSUE #275: add support for dynamic imports

hutchgrant:
## Related Issue
Resolves #275 

## Summary of Changes
* Adds `@babel/plugin-syntax-dynamic-import` dep
* Modified `babel.config.js` with additional plugin
* Modified `webpack.config.common.js` to split bundles
* Modified `page-template,js` to use dynamic imports and use a function attribute instead of array
* Modified `shelf.js` to await the promise from the JSON data import
* Updated babel to 7.8.3, added `babel-eslint` dependency
* Updated tests for the additional script element for dynamically imported chunk

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -31,7 +31,6 @@
     "@babel/preset-env": "^7.8.3",
     "@babel/runtime": "^7.8.3",
     "@webcomponents/webcomponentsjs": "^2.3.0",
-    "babel-eslint": "^10.0.3",
     "babel-loader": "^8.0.5",
     "chalk": "^2.4.2",
     "colors": "^1.3.3",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -25,8 +25,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@babel/core": "^7.8.4",
-    "@babel/parser": "^7.8.4",
+    "@babel/core": "^7.8.3",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@babel/plugin-transform-runtime": "^7.8.3",
     "@babel/preset-env": "^7.8.3",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -25,12 +25,14 @@
     "access": "public"
   },
   "dependencies": {
-    "@babel/core": "^7.6.0",
+    "@babel/core": "^7.8.4",
+    "@babel/parser": "^7.8.4",
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
-    "@babel/plugin-transform-runtime": "^7.6.2",
-    "@babel/preset-env": "^7.6.0",
-    "@babel/runtime": "^7.7.2",
+    "@babel/plugin-transform-runtime": "^7.8.3",
+    "@babel/preset-env": "^7.8.3",
+    "@babel/runtime": "^7.8.3",
     "@webcomponents/webcomponentsjs": "^2.3.0",
+    "babel-eslint": "^10.0.3",
     "babel-loader": "^8.0.5",
     "chalk": "^2.4.2",
     "colors": "^1.3.3",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@babel/core": "^7.6.0",
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
     "@babel/plugin-transform-runtime": "^7.6.2",
     "@babel/preset-env": "^7.6.0",
     "@babel/runtime": "^7.7.2",
--- a/packages/cli/src/config/webpack.config.common.js
+++ b/packages/cli/src/config/webpack.config.common.js
@@ -12,12 +12,12 @@ const getUserWorkspaceDirectories = (source) => {
 const mapUserWorkspaceDirectories = (directoryPath, userWorkspaceDirectory) => {
   const directoryName = directoryPath.replace(`${userWorkspaceDirectory}/`, '');
   const userWorkspaceDirectoryRoot = userWorkspaceDirectory.split('/').slice(-1);
-  
+
   return new webpack.NormalModuleReplacementPlugin(
     // https://github.com/ProjectEvergreen/greenwood/issues/132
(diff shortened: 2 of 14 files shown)


packages/cli/src/config/webpack.config.common.js
```diff
@@ -87,6 +87,7 @@ module.exports = ({ config, context }) => {
     output: {
       path: path.join(context.publicDir, '.', config.publicPath),
       filename: '[name].[hash].bundle.js',
+      chunkFilename: '[name].bundle.js',
```
thescientist13: will this work with `[hash]` too, or can they only be named?  e.g.
```javascript
chunkFilename: '[name].[hash].bundle.js',
```

hutchgrant: yea we can add hash too.

thescientist13:
Hmm, I can reproduce this locally.
```
$ yarn clean:deps && rm -rf node_modules
$ yarn install && yarn lerna bootstrap
$ yarn lint
```

This new package is getting installed into a node_modules "local" to the package, not the project, where everything else is.  

```bash
$ yarn why @babel/plugin-syntax-dynamic-import
yarn why v1.12.3
[1/4] 🤔  Why do we have the module "@babel/plugin-syntax-dynamic-import"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@babel/plugin-syntax-dynamic-import@7.2.0"
info Has been hoisted to "@babel/plugin-syntax-dynamic-import"
info Reasons this module exists
   - "workspace-aggregator-569b6660-214f-4be2-8d6e-14a9df5db7e2" depends on it
   - Hoisted from "_project_#@greenwood#cli#@babel#preset-env#@babel#plugin-syntax-dynamic-import"
   - Hoisted from "_project_#@greenwood#cli#@babel#preset-env#@babel#plugin-proposal-dynamic-import#@babel#plugin-syntax-dynamic-import"
info Disk size without dependencies: "20KB"
info Disk size with unique dependencies: "48KB"
info Disk size with transitive dependencies: "48KB"
info Number of shared dependencies: 1
=> Found "@greenwood/cli#@babel/plugin-syntax-dynamic-import@7.8.3"
info This module exists because "_project_#@greenwood#cli" depends on it.
info Disk size without dependencies: "20KB"
info Disk size with unique dependencies: "48KB"
info Disk size with transitive dependencies: "48KB"
info Number of shared dependencies: 1
✨  Done in 0.93s.
```

I think maybe we should install using closer version to our existing `@babel` dependencies, e.g. 7.2.x?  Or upgrade all the `@babel` deps to the same minor versions?

That said...

----

I would expect ESLint to ignore these files though.  It "should" ignore nested _node_modules_?  Or least we can add support for it.
https://github.com/eslint/eslint/issues/3788

Let me play around with this some more

hutchgrant:
I'm in favour of using the latest.

thescientist13:
Sounds good to me. 👍 

Let's just confirm either way if `@babel` recommends consistent versioning.  (and maybe something worth documenting in _CONTRIBUTING.md_)

hutchgrant:
With updated babel lint is now passing. However now we have the new problem which is the additional script tag causing 33 tests to fail:
```
  313 passing (3m)
  35 failing

  1) Build Greenwood With: 
       Empty Configuration and Default Workspace
         Running Smoke Tests: Empty Configuration and Default Workspace
           Index (Home) page
             should have one <script> tag for loading main JavaScript bundle at the end of the <body> tag:

      AssertionError: expected 2 to equal 1
      + expected - actual

      -2
      +1
      
      at Context.<anonymous> (test/smoke-test.js:97:44)
```

packages/cli/package.json
```diff
@@ -25,11 +25,13 @@
...
-    "@babel/plugin-transform-runtime": "^7.6.2",
-    "@babel/preset-env": "^7.6.0",
-    "@babel/runtime": "^7.7.2",
+    "@babel/core": "^7.8.3",
+    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
+    "@babel/plugin-transform-runtime": "^7.8.3",
+    "@babel/preset-env": "^7.8.3",
+    "@babel/runtime": "^7.8.3",
     "@webcomponents/webcomponentsjs": "^2.3.0",
+    "babel-eslint": "^10.0.3",
```
thescientist13: Oh, my bad, this shouldn't be a dependency (dev or otherwise) of the CLI.  ESLint is a dev tool, but also not used by the CLI, so should be moved to the root instead.  (where **eslint** currently lives)

thescientist13: Also, I was able to remove this and not have any issues
```
    "padding-line-between-statements": [
      "error",
      { "blankLine": "always", "prev": ["const", "let", "var"], "next": "*"},
      { "blankLine": "any",    "prev": ["const", "let", "var"], "next": ["const", "let", "var"]}
    ],
```

Also, based on **babel-eslint**, it is recommended to pass a _babel.config.js_.

thescientist13:
OK, I guess there is a question now on if we want to enable the `newline-after-var` rule?  I'm for it, I like some spacing in between variable declarations and any code after it.  (note: I had no problem with keeping it in _.eslintrc_ or any deprecation warnings)

hutchgrant:
If you were to re-enable `"newline-after-var": 2` how it is on master, you will see 38 errors which were not present before adding babel-eslint and upgrading babel.

thescientist13:
Ah, deprecated but replaced by this one instead; [https://eslint.org/docs/rules/padding-line-between-statements](padding-line-between-statements).

It looks more complicated to configure so I'm happy enough leave this as is since it's not / has been a big deal so far.

PR #292, opened 2020-02-25
Bug/issue 271 determinism

### LINKED ISSUE #271: query results are not deterministic

thescientist13:
## Related Issue
resolves #271 

## Summary of Changes
1. Pages are tracked as read from the filesystem, and inserted into `pages` Array to that index, instead of asynchronously being pushed .  Was trying to do something fancy [like this](https://decembersoft.com/posts/promises-in-serial-with-array-reduce/) but couldn't get it to work.
1. Documented graph return order
1. Added unit testing for _data/graph.js_ and updated test cases to expect a specific order each time

## Notes
1. ~~had to extend Mocha glob configuration, which is making the `test` and `test:tdd` tasks a little cumbersome to maintain now.  Will create a new issue for refactoring this configuration to use a _mocharc.js_ file~~ - see #294 
1. Could probably use some additional unit testing around other data files.  Should probably make a new issue for that.  Could also some up with a better way to capture `MOCK_GRAPH`


thescientist13:
Hmm.. test is failing because _second-post.md_ is still returned before _first-post.md_  😞  😠 
```
             {
               "generated": false
      -        "id": "Page:cc38e1248273d99"
      +        "id": "Page:f3f1bb94286324a"
               "type": "id"
               "typename": "Page"
             }
             {
               "generated": false
      -        "id": "Page:6d2fe63686567be"
      +        "id": "Page:9a4ce9aeeb41a94"
               "type": "id"
               "typename": "Page"
             }
           ]
      
```

thescientist13:
Merging per consensus reached during the last project meeting and no standing objections / comments on this PR.

PR #298, opened 2020-03-07
refactor shelf property from function to array

### LINKED ISSUE #297: website shelf not showing when loading a deep link with more than one route segment (from the shelf)
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [x] Bug

## Summary
When loading a route directly with a routing depth of greater than one segment, e.g. _/docs/configuration, where that second segment is coming from the shelf, the shelf doesn't show at all and there are no console errors.

Screen Shot 2020-03-07 at 4 52 06 PM

### Steps to reproduce
1. Got to the docs page
1. Expand configuration
1. The URL should now have show this: https://www.greenwoodjs.io/docs/configuration
1. Copy / paste that URL into a new browser tab

### Expected
The shelf would show on the left, expanded to the correct spot

### Actual
The shelf does not show.  (you can see it flash for a second)

### Observations
- This doesn't seem to happen in `develop` or `serve` mode, only when using `build` on Netlify
- Likely related to #275 but just need to confirm first
- This all happens in the `release/0.5.0` branch this happens too, but the loading of _cache.json_ is also breaking
Screen Shot 2020-03-07 at 4 59 40 PM

thescientist13:
## Related Issue
resolve #297 

## Summary of Changes
1. Refactored `shelfList` from `Function` type to Array

## TODO
1. remove console debugging after verifying with Netlify deploy preview

> In the issue, there are also 404's loading cache.json, I think that is something different and related to how our [client wrapper loads _cache.json_](https://github.com/ProjectEvergreen/greenwood/blob/release/0.5.0/packages/cli/src/data/client.js#L18) which is relative, so will need to fix that secondarily on the `release/0.5.0` branch

### DIFF
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -7,8 +7,8 @@ class shelf extends LitElement {
 
   static get properties() {
     return {
-      list: {
-        type: Function
+      shelfList: {
+        type: Array
       }
     };
   }
@@ -16,13 +16,11 @@ class shelf extends LitElement {
   constructor() {
     super();
     this.selectedIndex = '';
-    this.list = () => [];
     this.shelfList = [];
   }
 
   async connectedCallback() {
     super.connectedCallback();
-    this.shelfList = await this.list();
     this.collapseAll();
     this.expandRoute(window.location.pathname);
   }
@@ -123,6 +121,7 @@ class shelf extends LitElement {
   }
 
   render() {
+    console.log('render()', this.shelfList);
     return html`
     <style>
       ${css}
--- a/www/templates/page-template.js
+++ b/www/templates/page-template.js
@@ -14,28 +14,42 @@ METADATA;
 
 class PageTemplate extends LitElement {
 
+  static get properties() {
+    return {
+      shelfList: {
+        type: Array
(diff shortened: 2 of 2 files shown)


www/templates/page-template.js
```diff
@@ -14,28 +14,42 @@ METADATA;
...
     this.setupShelf();
   }
 
-  setupShelf() {
+  async setupShelf() {
     // based on path, display selected list
     const url = window.location.pathname;
     let list = [];
 
+    console.log('url', url);
```
hutchgrant: ''

www/components/shelf/shelf.js
```diff
@@ -123,6 +121,7 @@ class shelf extends LitElement {
   }
 
   render() {
+    console.log('render()', this.shelfList);
```
hutchgrant: is there a reason to keep this?

thescientist13: > TODO
> remove console debugging after verifying with Netlify deploy preview

just giving everyone a chance to review first 👍

thescientist13: :D

PR #300, opened 2020-03-10
Enhancement/issue 128 title from data

### LINKED ISSUE #128: <title></title> element is static
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
The <title></title> element is static and should potentially change on each page.

We should also remove `<title>` from _index.html_

## Details
In our meta component we're using a static title for each page, but perhaps we want the title to be static + a dynamic page name.  This would require using front-matter variables from each markdown page, and somehow through our API for other non-markdown pages to alter titles for each page's header.

e.g.  Greenwood | Getting Started

edit:

Also `og:title` and `og:description` are also static. I'd prefer them to be dynamic for each page as well.

thescientist13:
## Related Issue
resolves #128 

1. Set default `config.title` to _My App_
1. Move `<title>` setting to _app-template.js_
1. Pull `title` from data
1. Added logic / support for per page titles
1. Added unit tests and test cases
1. Update documentation

## TODO
1. [x] Handle per page specific `<title>`
1. [x] Update unit tests
1. [x] Documentation
1. [x] `<title>` doesn't change on sub sequent page change (need to hook into the router, or better use of lifecycles?) - see #306


packages/cli/src/data/schema/graph.js
```diff
@@ -69,19 +69,19 @@ const getChildrenFromParentRoute = async (root, query, context) => {
...
-        const { label } = getDeriveMetaFromRoute(route);
         const id = page.label;
 
         pages.push({
           id,
           filePath: mdFile,
           fileName,
           template,
-          title: label,
+          title: title !== '' ? title : label,
```
hutchgrant: :+1:

packages/cli/src/lifecycles/graph.js
```diff
@@ -72,7 +72,7 @@ const createGraphFromPages = async (pagesDir, config) => {
                 label = label || generateLabelHash(filePath);
 
                 // set <title></title> element text, override with markdown title
-                title = title || config.title;
+                title = title || '';
```
hutchgrant: :+1:

packages/cli/src/templates/app-template.js
```diff
@@ -17,6 +20,37 @@ import './list';
...
 
 class AppComponent extends LitElement {
+
+  async connectedCallback() {
+    super.connectedCallback();
+    const route = window.location.pathname;
+    const response = await Promise.all([
+      await client.query({
+        query: ConfigQuery
+      }),
```
hutchgrant: Is it easier to instead include this data as part of the graph, from within the graph lifecycle, then queried as part of the graph query? Just spitballing here.

thescientist13: Yeah, two queries here isn't ideal, that said, I think `config` should stay it's own schema / structure.  Mixing it up with the graph may not be super intuitive / maintainable.  

However, it would be great if **apollo-client** could support sending multiple queries for us, then we wouldn't have to use `Promise.all` but I suppose it could just be a limitation.  Would be good to think of another more elegant way if we can think of one.

PR #303, opened 2020-03-16
Enhancement/issue 125 meta from config

### LINKED ISSUE #125: improve unit test coverage for the meta plugin (get meta from data)
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
In coverage reporting the meta plugin is entirely being entirely overlooked and is uncovered.  
Screen Shot 2019-07-13 at 5 26 43 PM

## Details
This is likely down to its usage within the project though, in that it is used at runtime, not at build time.  At build time Greenwood makes user / default configuration available at build time through the scaffolding lifecycle.

As a result, this  might require a couple things to completed first in order to resolve this
- [Build Time Data Injection RFC](#115 )
- [Meta Plugin](#17 )

thescientist13:
## Related Issue
resolves #125 and builds off of / should be merged after #300 

> See the direct diff between [this PR and #300](https://github.com/ProjectEvergreen/greenwood/compare/enhancement/issue-128-title-from-data...enhancement/issue-125-meta-from-config)

## Summary of Changes
1. Moves `<meta>` data to _app-template.js_, like with `<title>` and pull from GraphQL
1. Remove instances of `METAELEMENT`, `METADATA`, `METAIMPORT`
1. Add missing `href` from config schema
1. Update documentation
1. Update code coverage thresholds 💯 

## TODO (nice to have)
1. Consider a meta specific `qql` query?


packages/cli/src/templates/app-template.js
```diff
@@ -17,6 +20,76 @@ import './list';
...
+    const currentPageTitleSuffix = currentPage.link === '/'
+      ? ''
+      : ` - ${currentPage.title}`;
+    const fullTitle = `${config.title}${currentPageTitleSuffix}`;
+
+    this.setDocoumentTitle(fullTitle);
+    this.setMeta(config.meta, currentPage);
+  }
+
+  setDocoumentTitle(title = '') {
```
hutchgrant: sp.   setDocument()

packages/cli/src/templates/app-template.js
```diff
@@ -17,6 +20,76 @@ import './list';
...
+    const head = document.head;
+    const titleElement = head.getElementsByTagName('title')[0];
+
+    titleElement.innerHTML = title;
+  }
+
+  setMeta(meta = [], currentPage = {}) {
```
hutchgrant: I really liked this in a separate component as people will want to do custom app-templates, it would be easier to simply import the metaComponent.  Thoughts?

packages/cli/test/cases/build.default.workspace-template-app/src/templates/app-template.js
```diff
@@ -18,6 +20,25 @@ import './list';
...
+
+    const response = await client.query({
+      query: ConfigQuery
+    });
+    const { config } = response.data;
+
+    this.setDocoumentTitle(config.title);
+  }
+
+  setDocoumentTitle(title) {
```
hutchgrant: sp.  this.setDocument()

thescientist13:
> Spelling mistake on function, I don't know if that was intentional. 
😬 

Good catch.  👍 

> Is it not easier for other developers looking to make their own app-templates if we can refactor the meta query and dom changes into a separate component ?

I think that's definitely a good idea and so I think you are thinking of having something equivalent to [**react-helmet**](https://www.npmjs.com/package/react-helmet).  I think that's a good idea so unless there's any strong objection, I think I would like to make an issue for that, primarily to determine if
1. It should aliased within the CLI package like we do with data, e.g. `@greenwood/cli/components/meta`
1. Or as its package, but as a dependency, e.g. `@greenwood/component-meta`

Issue #313, opened 2020-03-27
Documentation for Netlify CMS 


hutchgrant:
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Netlify CMS has an [excellent guide](https://www.netlifycms.org/docs/add-to-your-site/#configuration) on how to setup Netlify with most the popular static generators around. I figured, why not add and document how to use Netlify with Greenwood.  This began as an experiment but I'm thinking this definitely something useful that's worth documenting and sharing.

## Details

We need a page dedicated to configuring netlify CMS with a greenwood project. I suggest a guide that uses the [greenwood-getting-started](https://github.com/ProjectEvergreen/greenwood-getting-started) repo as a starting off point.

PR #314, opened 2020-03-27
Task/issue 313 add netlify cms docs

hutchgrant:
## Related Issue
#313 

## Summary of Changes
* updates navigation adding a new "guides" category
* adds new "guides" shelf
* adds new "guides" section index page
* adds netlify-cms page
* adds screenshot assets

edit: updated to add new "guides" category, shelf, and index page.


www/pages/getting-started/netlify-cms.md
```diff
@@ -0,0 +1,144 @@
...
+<img src="/assets/netlify-git-gateway.png" alt="netlify--git-gateway" style="max-width:800px;"/>
+```
+
+To automatically generate an access token from and for the GitHub API
+
+### Access Netlify CMS
+
+Your site is now deployed and configured, you can now login and access your Netlify CMS from:
+ `https://yourgeneratedurl-ads6387.netlify.com/admin`
+
```
thescientist13, line 144: Is there anything next add here?  Not sure if you covered an "authoring" content step that would verify everything is working where in the reader would add a test page, see it commit to GitHub, and build in Netlify.  Maybe I just missed it though?

hutchgrant: No, I didn't add that step because I guess I thought it was self expalanatory. The guide was just to setup netlify cms not using it or testing it out.

hutchgrant:
If you're in agreement with a new guides section, as we discussed, then I'll certainly add one.

www/pages/getting-started/netlify-cms.md
thescientist13: Yeah, it might be nice to do one of two things to close things out:
1. Provide a couple next steps and walk the reader a couple basic steps

or

1. Share a link with how to get started publishing content with Netlify CMS

Either way, a nice way to tie things up and bring the guide to a nice resolution.  Your call.

www/components/shelf/guides.json
```diff
@@ -0,0 +1,22 @@
+[
```
thescientist13, line 1: shouldn't this not be needed anymore?

www/templates/page-template.js
```diff
@@ -37,10 +37,10 @@ class PageTemplate extends LitElement {
       <style>
         ${pageCss}
       </style>
-     
+
       <div class='wrapper'>
         <eve-header></eve-header>
-        
+
```
thescientist13, line 43: nitpick, but unnecessary formatting changes here.

www/components/shelf/guides.json
hutchgrant: yes removed

www/templates/page-template.js
hutchgrant: removed

hutchgrant:
Owen we added a guides section, that changed the mock graph and the tests

thescientist13:
> Owen we added a guides section, that changed the mock graph and the tests. Some of our tests are the number of pages in the graph, as well as the number of links in the navigation, etc.

I understand, but just because we add content, doesn't mean it was (also) a change to the schema.  To be clear, the mock graph / unit tests don't care about the content really.  It could be all Lorum Ipsum and two navigation sections for all the tests care.

And so it is precisely the reason why I don't want to have to change the tests each time content changes, it would be cumbersome and wouldn't add any real value anyway.

> Also front matter has changed on index pages because if its sorted alphabetically within navigation menu, the "Guides" section would be the fourth item instead of fifth. This way we've given each index page an index within the sort so that they sort in the order we want(with guides as last item)

Gotcha, that makes sense, I didn't think of that.

----

So just rollback the testing changes and then this should be good, and then it will pave the way for the other good content Adam has lined up as PRs.  💪

hutchgrant:
Owen if a single change happens later and you need to regenerate the graph
you will have to reads these test changes though.

thescientist13:
For sure, but I really don't hope that is something that happens a lot, since those would likely be considered breaking changes to users (our schema changing all the time).  In the meantime, it really only takes a second to tweak the graph manually for when it actually needs to happen.

But more to the point, a content change shouldn't require all our tests to change.  If there really is an exceptional case we need to test in the graph (certain number of pages, route depth, etc) then it would be better to just keep mocking it out.

hutchgrant:
I agree, then perhaps that's something should be part of seperate issue
that should be addressed so that the test is agnostic to current site
content.

thescientist13:
> I agree, then perhaps that's something should be part of separate issue
that should be addressed so that the test is agnostic to current site
content.

for sure.  I created #327 for tracking something similar.  I think this conversation is in a similar vein so we can continue to discuss there.

PR #338, opened 2020-04-20
render keyword not needed for code formatting in markdown (wc-markdown-loader to 0.2.0)

### LINKED ISSUE #337: remove render keyword needed for code formatting in markdown
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
There's a new release of [wc-markdown-loader](https://github.com/hutchgrant/wc-markdown-loader/releases/tag/v0.1.3) we should update dep `"wc-markdown-loader": "^0.1.4"`

## Details
wc-markdown-loader 0.1.4 has removed the `render` fencing because it's now supporting use of unified js plugins such as remark and rehype-prism.

This means html will now render within markdown without any additional work. The `render` string is no longer necessary within code blocks we want to highlight.  We can also add additional plugins via the **preset** option of the loader within our webpack config. An example would be `rehype-slug` and `rehype-autolink-headings`.

hutchgrant:
## Related Issue
Resolves #337 

## Summary of Changes

* Changed dep `wc-markdown-loader: ^0.2.0`
* Updated all docs removing all render fencing


packages/cli/package.json
```diff
@@ -57,7 +57,7 @@
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
     "style-loader": "^0.23.1",
-    "wc-markdown-loader": "^0.1.2",
+    "wc-markdown-loader": "^0.1.4",
```
thescientist13, line 60: is there any way that **wc-markdown-loader** can follow [semver](https://semver.org/)?  I don't think it's intuitive for features to be getting added and published as patches, when typically that would happen as part of a minor.  (or major if breaking).

It's easy enough for Greenwood to pin to minors using `~` for pre 1.0.0, so as to not impact downstream users when new releases of **wc-markdown-loader** are published.

thescientist13:
Awesome!  (and just adjusted the title a lit bit to capture the higher level feature here).

www/pages/docs/component-model.md
```diff
@@ -22,7 +23,7 @@ class FooterComponent extends HTMLElement {
   getTemplate() {
     const year = new Date().getFullYear();
 
-    return \`<header>This is the header component.  &copy; ${year}</header>\`;
+    return \`<header>This is the header component.  &copy; \${year}</header>\`;
```
thescientist13, line 26: Is this expected, that now code blocks inside code fences need to be escaped?  If it's markdown, it should treat it all as a string, right?

Or Is it because this is getting processed as JavaScript instead?

hutchgrant: Yea it has to be escaped, no it's not treating it as a string, might have something to do with syntax highlighting.

thescientist13: Would be cool to get something like GitHub flavored markdown
Screen Shot 2020-04-28 at 7 42 29 PM

Screen Shot 2020-04-28 at 7 42 34 PM

hutchgrant:
https://www.greenwoodjs.io/about/community  it's already linked in the community docs

www/pages/guides/netlify-cms.md
```diff
@@ -155,16 +145,13 @@ If you've been following along, you should now be viewing the netlify CMS admin
 
 From this page, you can see the default `blog` content type that we added in our netlify cms configuration file. When we create and publish a new `blog` page it will publish to the `/public/blog` folder in your repository.
 
-```render
 <img src="/assets/netlify-admin.png" alt="netlify-cms" style="max-width:800px;"/>
-```
```
thescientist13, line 160: whoa... so this means HTML "just works" now, right?  ™️

thescientist13:
> https://www.greenwoodjs.io/about/community it's already linked in the community docs

Cool, thanks. 

Yeah, I was more just making sure it was clear to users that they would know the "rules" of the tool.  Not sure if there is a markdown spec that would establish the default behavior of what would happen in this scenario with the template literal syntax needing to be escaped, and so would a user know to escape by default?  I'm not sure the error message that happens (if any). I think the page just goes white, or that part of the page does.  

Anyway, this is what someone might refer to as ["institutional knowledge"](https://en.wikipedia.org/wiki/Institutional_memory), and so just wanted to know if it made since to document this specific behavior / expectations somewhere.  (e.g. Greenwood's markdown support)

www/pages/docs/markdown.md
```diff
@@ -9,29 +9,29 @@ linkheadings: 3
...
-
-```render js
+```js
 const hello = "world";
 
-<p>${hello}</p>
+<p>\${hello}</p>
 ```
 
+> **Note:** As demonstrated in the above example, backticks and `\$` characters require `\\` to escape correctly.
```
hutchgrant, line 33: As pointed out, we've added this caveat to the documentation, to clarify the only exception to regular markdown that we're aware of.

PR #344, opened 2020-04-30
bug/issue 340 Shelf's First Nav Item Expanding By Default

### LINKED ISSUE #340: website shelf menu gets stuck on a page where first nav item has sub headings
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Bug

## Summary
If you go to the guides page right now, you will find that while the content displayed is that of the _guides/index.md_ page, the shelf on the left is already auto-expanded to show the link heading children of the first shelf item.
Screen Shot 2020-04-29 at 9 00 54 AM

However, clicking any of those links or the nav item itself does nothing.  The only way to access the content is to click the downward facing carrot, which is not very intuitive.

## Details
It is likely due to [this logic](https://github.com/ProjectEvergreen/greenwood/blob/master/www/components/shelf/shelf.js#L56), which auto expands the first selected item automatically, but I don't think it checks to see if it's on a top level / index.md route, e.g. _/guides/_ to know if it _should_ expand or not.

```js
  expandRoute(path) {
    // find list item containing current window.location.pathname
    let routeShelfListIndex = this.shelfList.findIndex(list => {
      return list.item.link.indexOf(path) >= 0;
    });

    if (routeShelfListIndex > -1) {
      this.shelfList[routeShelfListIndex].selected = true;
      this.selectedIndex = routeShelfListIndex;
      // force re-render
      this.requestUpdate();
    }
  }
```

This gets called during `onConnectedCallback` so will happen each time.  This isn't an issue for other pages because so far the guides page is the only page that has linkHeadings for it's first shelf child.

hutchgrant:
## Related Issue
Resolves #340 

## Summary of Changes

* fixed the shelf list index condition for when to expand a menu based on whether the current route is in shelf list

### DIFF
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -48,9 +48,11 @@ class Shelf extends LitElement {
   }
 
   expandRoute(path) {
+
     // find list item containing current window.location.pathname
     let routeShelfListIndex = this.shelfList.findIndex(list => {
-      return list.item.link.indexOf(path) >= 0;
+      let expRoute = new RegExp(`${path}$`);
+      return expRoute.test(list.item.link);
     });
 
     if (routeShelfListIndex > -1) {
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -48,10 +48,9 @@ class Shelf extends LitElement {
   }
 
   expandRoute(path) {
-
     // find list item containing current window.location.pathname
     let routeShelfListIndex = this.shelfList.findIndex(list => {
-      let expRoute = new RegExp(`${path}$`);
+      let expRoute = new RegExp(`^${path}$`);
       return expRoute.test(list.item.link);
     });
 


www/components/shelf/shelf.js
```diff
@@ -50,7 +50,8 @@ class Shelf extends LitElement {
   expandRoute(path) {
     // find list item containing current window.location.pathname
     let routeShelfListIndex = this.shelfList.findIndex(list => {
-      return list.item.link.indexOf(path) >= 0;
+      let expRoute = new RegExp(`^${path}$`);
```
thescientist13, line 53: ah... I think this has been a solution i've been looking for, wherever we have to do this sort of pattern matching for current route like in [app-template.js ](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/templates/app-template.js#L26) and could help fix our [missing forward `/` issue](https://trello.com/c/AuzJwHkZ)?

Basically anywhere else we are doing something like this
```
const route = window.location.pathname;

const currentPage = response[1].data.graph.filter((page) => {
  return route === page.link;
})[0];
```

Issue #331, opened 2020-04-18
routes with and without trailing forward slash causes route "misses"


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
As evidenced in PR #308, it looks like if there is or isn't a forward slash at the end of a route. e.g. 

http://domain.com/about 

vs.

http://domain.com/about/

Can lead to issues with matching against active routes in GraphQL / coming out of the Greenwood graph and even cause errors in the console, although I think Netlify will add / forward to the `/` so then it "catches".

<img width="1647" alt="77268328-ea8b5b00-6c7b-11ea-8cff-a53442cc83f5" src="https://user-images.githubusercontent.com/895923/79667701-12d77e00-8184-11ea-9b19-f9823a18088c.png">

## Details
I think anywhere where we are matching against an active route or using `window.location.hostname` or similar to find out the current page, we will need to evaluate how it is being implemented and try and build a fairly robust / non fragile solution.

Also see #310 for reference.

thescientist13:
Well, now adding the forward slash removes the `undefined` error, but adds this error now instead 😞 🤷 
<img width="1647" alt="Screen Shot 2020-04-22 at 6 28 41 PM" src="https://user-images.githubusercontent.com/895923/80040131-6c62e400-84c7-11ea-8e0f-604658bffec3.png">
```
index.b37c464fdfd3e49a05d7.bundle.js:1 Uncaught (in promise) Invariant Violation: Invariant Violation: 8 (see https://github.com/apollographql/invariant-packages)
    at new e (https://deploy-preview-284--elastic-blackwell-3aef44.netlify.app/index.b37c464fdfd3e49a05d7.bundle.js:1:16081)
    at https://deploy-preview-284--elastic-blackwell-3aef44.netlify.app/index.b37c464fdfd3e49a05d7.bundle.js:117:58147
    at Array.forEach (<anonymous>)
    at n.diffQueryAgainstStore (https://deploy-preview-284--elastic-blackwell-3aef44.netlify.app/index.b37c464fdfd3e49a05d7.bundle.js:117:58105)
    at n.readQueryFromStore (https://deploy-preview-284--elastic-blackwell-3aef44.netlify.app/index.b37c464fdfd3e49a05d7.bundle.js:117:57465)
[...]

PR #346, opened 2020-05-07
make cache file writing sync and hashed by query

### LINKED ISSUE #345: cache.json is large due to containing multiples copies of the same data
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Moving the discussion from https://github.com/ProjectEvergreen/greenwood/issues/317#issuecomment-619227081, but after the `0.5.0` release, there are some serious concerns with _cache.json_ generation.  Docs page is transfering over 2.6MB of JSON data!

1. It's being over `fetch`ed (why!?)
1. It's huge!  Docs page
image

This also means all _index.html_ pages are huge too...
Screen Shot 2020-05-03 at 11 48 19 AM

## Details
Thoughts on quick wins here:
1. Fix over`fetch`ing by caching in the client. (but would also be good to know why this is happening)
1. Reduce duplication of query results within a single cache.json file (e.g. _about/cache.json_) because **deepmerge** just keeps doubling up the previous file contents, resulting in duplicates over and over again?

For something like _docs/cache.json_, this could bring the size down to 35k! We should also try and add a "budget" spec for this to make sure the size can't explode again without it failing at least.

> Later down the road, we should look to reduce duplication of query results across all _cache.json_ files - see #317

thescientist13:
## Related Issue
resolves #345 

## Summary of Changes
1. Write files by query, not by page
1. And write them synchronously to avoid overwriting

> For other mentioned issues #345, i will split them off into their own issues

### DIFF
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -86,7 +86,7 @@ class Shelf extends LitElement {
       query: MenuQuery,
       variables: {
         name: 'side',
-        route: this.page,
+        route: `/${this.page}/`,
         order: 'index_asc'
       }
     });
--- a/www/components/shelf/shelf.js
+++ b/www/components/shelf/shelf.js
@@ -25,24 +25,6 @@ class Shelf extends LitElement {
   async connectedCallback() {
     super.connectedCallback();
     this.collapseAll();
-    //   console.log('shelf page', this.page);
-    //   console.log('shelf window.location.pathname', window.location.pathname);
-    //   if (this.page && this.page !== '' && this.page !== '/') {
-    //     const response = await client.query({
-    //       query: MenuQuery,
-    //       variables: {
-    //         name: 'side',
-    //         route: window.location.pathname,
-    //         order: 'index_asc'
-    //       }
-    //     });
-
-    //     this.shelfList = response.data.menu.children;
-    //     this.requestUpdate();
-    //   }
-
-    //   this.collapseAll();
-    //   this.expandRoute(window.location.pathname);
   }
 
   goTo(path) {
@@ -51,26 +33,18 @@ class Shelf extends LitElement {
   }
 
   expandRoute(path) {
-    console.log('expandRoute', path);
-    console.log('this.shelfList', this.shelfList);
-    // find list item containing current window.location.pathname
     let routeShelfListIndex = this.shelfList.findIndex(list => {
       let expRoute = new RegExp(`^${path}$`);
-      console.log('list.item.link', list.item.link);
       return expRoute.test(list.item.link);
     });
-    console.log('routeShelfListIndex', routeShelfListIndex);
 
     if (routeShelfListIndex > -1) {
       this.shelfList[routeShelfListIndex].selected = true;
       this.selectedIndex = routeShelfListIndex;
-      // force re-render
-      // this.requestUpdate();
     }
   }
 
   collapseAll() {
-    console.log('collapseAll');
     this.shelfList = this.shelfList.map(item => {
       item.selected = false;
       return item;
@@ -108,9 +82,6 @@ class Shelf extends LitElement {
   }
 
   async fetchShelfData() {
-    console.log('shelf page', this.page);
-    console.log('shelf window.location.pathname', window.location.pathname);
-    // if (this.page && this.page !== '' && this.page !== '/') {
     const response = await client.query({
       query: MenuQuery,
       variables: {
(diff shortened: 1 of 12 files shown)


thescientist13:
Ugh... getting these stupid invariant errors in the Netlify preview environment. :/
Screen Shot 2020-05-06 at 8 29 42 PM

I don't know if it's related to this change though...?


Had a thought on it here https://github.com/ProjectEvergreen/greenwood/issues/331#issuecomment-624951466

hutchgrant:
> Ugh... getting these stupid invariant errors in the Netlify preview environment. :/

Your Menu Query for the side menu is not being cached. Check cached files and you'll see theres only a navigation menu query, config, and graph queries.

hutchgrant:
When using: 

```js
const md5 = crypto.createHash('md5').update(query).digest('hex');
```

I saw the following error:

"SyntaxError: /media/skynet/DATA/workspace/evergreen/greenwood/public/plugins/2bc8e256a25844b37c22af93673c67e3-cache.json: Unexpected token d in JSON at position 2830"

It's not being caught. I was only able to catch it by switching back to async fs.writeFile and fs.mkdirs.


If you switch back to:

```js
const md5 = crypto.createHash('md5').update(cache).digest('hex');
```

While still using synchronous as per this PR, it works fine, no errors.

thescientist13:
> Your Menu Query for the side menu is not being cached. Check cached files and you'll see theres only a navigation menu query, config, and graph queries.

Oof...

> Notice how our testing didn't catch this? Because it relies on mock graph.

Yeah, noticed that too.  Will see if I can do something about that somehow.

> It's not being caught. I was only able to catch it by switching back to async fs.writeFile and fs.mkdirs.

Not quite sure I understand, but you are saying saw the JSON parse error when testing this branch and using `query`?  That's a bummer, I ran it a bunch of times on my local machine.  

So this might be a bit more complex of an issue, but I then since every _page_ gets serialized in parallel, all these GraphQL queries are happening as well.  I think it would be similar to what we did here for #271 ?

So the most sure fire way would be to make it all synchronous then.  Or i guess try the file locking approach?  Other options?

hutchgrant:
The reason you aren't creating a side menu cache is simple: you're using the same data to generate a hash and therefore it only writes one file for two different queries!

menu query for side and navigation:
```js
query ($name: String, $route: String, $order: MenuOrderBy) {
  menu(name: $name, pathname: $route, orderBy: $order) {
    item {
      label
      link
      __typename
    }
    children {
      item {
        label
        link
        __typename
      }
      children {
        item {
          label
          link
          __typename
        }
        __typename
      }
      __typename
    }
    __typename
  }
}

```


The fix is to include the query's variables to make it unique:
```js
const md5 = crypto.createHash('md5').update(JSON.stringify(query) + JSON.stringify(variables)).digest('hex');
```

www/components/shelf/shelf.js
```diff
@@ -99,6 +83,32 @@ class Shelf extends LitElement {
...
+
+    this.shelfList = response.data.menu.children;
+    console.log('fetchShelfData - ', this.shelfList);
+  }
+
+  async updated(changedProperties) {
+    console.log('ENTER updated - changedProperties', changedProperties);
+    console.log('updated - this.page', this.page);
+    if (changedProperties.has('page') && this.page !== '' && this.page !== '/') {
+      await this.fetchShelfData();
```
hutchgrant, line 105: You're expecting to return a promise that doesn't exist.  Maybe just return the query(which is a promise) like this:

```js
const response = await this.fetchShelfData();
this.shelfList = response.data.menu.children;
```

www/templates/page-template.js
```diff
@@ -26,12 +26,14 @@ class PageTemplate extends LitElement {
   }
 
   updated() {
```
hutchgrant, line 28: My understanding is this lifecycle is run after the component has been rendered. I tested using the connectedCallback() to set the this.page variable from the page-template and had better results (I wasn't able to break it). Maybe that's just chance, I don't know.

www/templates/page-template.js
```diff
@@ -26,12 +26,14 @@ class PageTemplate extends LitElement {
...
+    console.log('ENTER page template updated', window.location.pathname);
     this.route = window.location.pathname;
```
hutchgrant, line 30: My understanding is this lifecycle is run after the component has been rendered. I tested using the connectedCallback() to set the this.page variable from the page-template and had better results (I wasn't able to break it). Maybe that's just chance, I don't know.

thescientist13: `this.page` was actually undefined at that point, so `updated` seemed like a good way to make the sure the value is "ready".

www/components/shelf/shelf.js
thescientist13: still happening, I guess I should just revert back the `connectedCallback` / `window.location.pathname` approach then? 

Not sure if we'll need to start recommending / documenting certain patterns around this stuff then, though I would hope we don't have to be any different than SPA development.  Like if having unpredictable lifecycles could be a concern, or just needing to guard initial values of your variables.

PR #361, opened 2020-06-04
task/issue 360 update app-template documentation

### LINKED ISSUE #360: App Template documentation example needs to be updated
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [x] Bug

## Summary
For one, the [app-template documentation page](https://www.greenwoodjs.io/docs/layouts#app-template) example suggests defining an element name which will not work with the default index.html we use. We also shouldnt include the magic keyword string *MYROUTES* within a comment in case someone copies the example. We don't want the routes to be written during scaffold lifecycles to a comment within the template!

## Details

1.  `<eve-app></eve-app>` is hardcoded to the default index.html.  
1. `MYROUTES` should not be in a comment in the example
1. `import './list'; `is no longer needed because of recent code splitting changes

hutchgrant:
## Related Issue
Resolves #360 

## Summary of Changes

1. Place comment about eve-app hardcoded defined app-template element name
2. Remove need for `../list.js`
3. Replace `MYROUTES` with `MY-ROUTES(without dash)`

### DIFF
--- a/www/pages/docs/layouts.md
+++ b/www/pages/docs/layouts.md
@@ -83,13 +83,13 @@ const store = createStore(
 
 // Next we need to import a list of files that will be generated when Greenwood is run
 import '../index/index.js';
-import './list';
 
 // Finally we can connect to our store and define our component.
 connectRouter(store);
 
-// Add a \`MYROUTES\` predefined hook. This is where all your routes will be loaded.
+// Add the MY-ROUTES(without dash) predefined hook. This is where all your routes will be loaded.
 // You may also opt to define a custom 404 route here.
+// You must define the app-template with the element name eve-app
 class AppComponent extends LitElement {
   render() {
     return html\`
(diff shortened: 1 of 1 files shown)


www/pages/docs/layouts.md
```diff
@@ -83,13 +83,13 @@ const store = createStore(
 
 // Next we need to import a list of files that will be generated when Greenwood is run
 import '../index/index.js';
-import './list';
 
 // Finally we can connect to our store and define our component.
 connectRouter(store);
 
-// Add a \`MYROUTES\` predefined hook. This is where all your routes will be loaded.
+// Add the MY-ROUTES(without dash) predefined hook. This is where all your routes will be loaded.
```
thescientist13, line 90: Why not just keep without the dash in the comment too?  Also, maybe ton confuse hook with our plugins, let's call it a constant?

so all together would be:
> `// Add the `MYROUTES` predefined constant. This is where all your routes will be loaded.`

hutchgrant: I had to add dash intentionally otherwise it will be caught by our find and replace within scaffold lifecycle which breaks compilation.

Issue #363, opened 2020-06-08
broken custom-media queries


hutchgrant:
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Currently, our site banner and page-template utilise the [evergreen web component library](https://github.com/hutchgrant/evergreen-web-components) which was previously using postcss-next.   With that deprecation, and with using postcss-preset-env instead, our `@custom-media` queries don't work correctly. With the updated component library, along with postcss-preset-env set to `stage: 1`,  we can utilize these components correctly responsively.  

## Details

Current `<eve-container></eve-container>` element is not breaking correctly on different size windows. That's because @custom-media wasn't being loaded properly by postcss.  We need to set postcss-preset-env to `stage: 1` in addition to updating the `@greeenwood/www package.json` with the new updated `@evergreen-wc/eve-container: ^0.0.10`

thescientist13:
@hutchgrant 
Was reviewing the deploy preview of #366 but wasn't able to see the breakpoint issue you are referring to?  (essentially greenwoodjs.io/docs/configuration and  https://deploy-preview-366--elastic-blackwell-3aef44.netlify.app/docs/configuration).

Could you maybe post a screenshot in your issue description and detail / document what is broken, and so what could be verified to confirm it was fixed?

hutchgrant:
`<eve-container></eve-container>` in banner has broken media query

![broken_media_query](https://user-images.githubusercontent.com/1429612/84599580-08aec480-ae41-11ea-97b3-4b493efdbc13.png)

thescientist13:
So are you saying it should be broken down even further at lower break points?  

As I mentioned, comparing this against prod website it looked / behaved the same, so from a user perspective was anything fixed?  Not sure then what is "broken" here then, per se?

hutchgrant:
Here's is the fixed @custom-media query from #366 

![fix_media_query](https://user-images.githubusercontent.com/1429612/84599635-73600000-ae41-11ea-9824-b576c217d774.png)

hutchgrant:
@custom-media is basically like a css variable for media queries.

When those aren't being compiled correctly because in evergreen-wc it was compiled with postcss-cssnext which is deprecated.  So the component itself had to be updated and the postcss-preset-env stage 1 had to be used for the media query to function properly.

hutchgrant:
[...]

PR #366, opened 2020-06-13
task/issue 365 add support for custom babel/postcss configs in user workspace

### LINKED ISSUE #365: support custom PostCSS / Babel configurations
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Currently there is no way to extend Greenwood's built in Babel and PostCSS configuration, to either add custom plugins or additional stages, as seen in #363 .

## Details
In _webpack.config.common.js_ the relative config files for both are provided and loaded.  We should update this to
1. Detect if equivalent file is found in users project directory
1. Use that version instead
1. Document in the website the new behavior
1. Add a test (I suppose detecting transpiled features in the output would have to the be test?)

Only immediate thought that comes to mind that might make this tricky is that both assume the relative location of _.browserslistrc_ which may differ now when the Babel and PostCSS files are loaded from a different location now.  Not sure if _.browserslistrc_ path can be hardcoded or configured on the fly in that case?  Seems like this enhancement could / should be done along with #224 ?

hutchgrant:
## Related Issue
Resolves #365 #363 

## Summary of Changes
* Modifies context to check for babel/postcss config in workspace, otherwise defaults to built in configs
* adds custom `postcss.config.js` for website in `www` directory
* updates `evergeen-wc/container` to `0.0.10`
* add new docs section titled: build


packages/cli/src/lifecycles/context.js
```diff
@@ -24,6 +27,8 @@ module.exports = initContexts = async({ config }) => {
       const userHasWorkspaceAppTemplate = await fs.exists(userAppTemplate);
       const userHasWorkspaceIndexTemplate = await fs.exists(path.join(userTemplatesDir, 'index.html'));
       const userHasWorkspaceNotFoundTemplate = await fs.exists(path.join(userTemplatesDir, '404.html'));
+      const userHasWorkspaceBabel = await fs.exists(path.join(config.workspace, babelConfig));
```
thescientist13, line 30: I think default configuration / expectation for these kinds of files is that they are typically kept at the project root.

So instead of
```
<workspace>/
  babel.config.js
  pages/
  ...
```

It would make more sense to support
```
babel.config.js
<workspace>/
  pages/
  ...
```

www/package.json
```diff
@@ -8,6 +8,6 @@
   "license": "MIT",
   "dependencies": {
     "@evergreen-wc/eve-button": "^0.0.9",
-    "@evergreen-wc/eve-container": "^0.0.9"
+    "@evergreen-wc/eve-container": "^0.0.10"
```
thescientist13, line 11: I think I would still be good to keep #366 open and have that handle just to isolate the dependency related changes from the new enhancement work.

www/postcss.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
thescientist13, line 1: same goes for this, re: #366

www/package.json
thescientist13: Actually, this is probably small enough and closely related enough that it should be fine.  I had to do something similar for #367 and so probably makes sense and takes the most advantage of a monorepo.  Up to you.

hutchgrant:
[browserslist will default to current directory](https://github.com/browserslist/browserslist#queries)  I assume of the babel.config.js

packages/cli/test/cases/build.config.babel/build.config.babel.spec.js
```diff
@@ -0,0 +1,58 @@
...
+
+    describe('index page should not compile properly', function() {
+      let dom;
+
+      beforeEach(async function() {
+        dom = await JSDOM.fromFile(path.resolve(this.context.publicDir, './index.html'));
+      });
+  
+      it('should not contain any components within eve-app', function() {
+        // prove that our custom broken babel config is being used
```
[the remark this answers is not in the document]
hutchgrant: It's broken because I removed the babel plugins. It's just enough to get it to compile but it won't compile properly, which is testable.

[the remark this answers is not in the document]
hutchgrant: We of course know that it's using our custom babel.config.js if the compilation isn't accurate to what it should be.

www/pages/docs/css-and-images.md
```diff
@@ -2,7 +2,7 @@
 label: 'styles-assets'
 menu: side
 title: 'Styles and Assets'
-index: 5
+index: 6
```
thescientist13, line 5: Were some of these out of order?

hutchgrant:
Yes good point.

www/pages/docs/build.md
```diff
@@ -0,0 +1,36 @@
...
+To override the default **postcss.config.js** with your own configuration, create a new postcss.config.js file within your project's root directory.
+
+By default, [this is the postcss.config.js](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/config/postcss.config.js) being used.
+
+
+### Babel
+
+To override the default **babel.config.js** with your own configuration, create a new babel.config.js file within your project root directory.
+
+> Note: If you use your own **babel.config.js** you need to include your own [.browserslist](#browserslist) file within the project's root directory
```
thescientist13, line 25: Actually, this isn't entirely the case is it?
> @babel/preset-env also has an option to set the config path by default we have it set to use the current directory of the babel.config.js

So in theory, they could still use our _browserslistrc_ by passing the path _./node_modules/@greenwood/cli/src/config/.browserslistrc_.  Not sure if that is worth mentioning?

www/pages/docs/build.md
```diff
@@ -0,0 +1,36 @@
...
+
+By default, [this is the babel.config.js](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/config/babel.config.js) being used.
+
+
+### Browserslist
+
+By default, the **.browserslistrc** file is found in the same directory as the `babel.config.js`.  If you want to override it, include a **.browserslistrc** file within your project's root directory. You are also required to include your own `babel.config.js` file if you're providing a custom **.browserslistrc**, also within your project's root directory. You can specify an alternative path to the browserslist using the `configPath` setting of your `babel-preset-env` options within your `babel.config.js` file.
```
thescientist13, line 32: I think it should be noted that _.browserslistrc_ is used by both Babel and PostCSS preset-env packages.

www/pages/docs/build.md
hutchgrant: Well just in case I added it. Can only use ENV variables or modify package.json to set path to browserlist elsewhere for postcss

PR #369, opened 2020-06-14
task/issue 329 Auto link headings and scroll all headings

### LINKED ISSUE #329: [website] Scroll component only scrolls to h3 elements
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Bug

## Summary
With the addition of the MenuQuery in #295 we will have the ability to link any headings from a page e.g. h1, h2, h3 etc, into a single menu.  Currently, the scroll component is hardcoded to only scroll to h3 elements containing a hash. It would be ideal if we could scroll to any type of heading element.

## Details

scroll.js
```js
    // clean hash string, remove # and replace - with spaces
    hash = hash.replace('#', '').toLowerCase();
    // for cases of multiple hyphens
    hash = hash.replace(/-/g, ' ');
    // query text
    const heading = contains('h3', hash)[0];
```

hutchgrant:
## Related Issue
Resolves #329

## Summary of Changes

* adds `wc-markdown-loader` preset with `rehype-slug` and `rehype-autolink-headings` dependencies
* Updates scroll component to scroll to any anchor element hash
* adds a link button to every heading(could be styled fancier if anyone wants to have at it).


hutchgrant:
We would have to allow override of webpack config by user. Just like how CRA can provide a a copy of webpack and other config it uses, which the user can then modify.

hutchgrant:
created a new issue for supporting custom webpack configurations #370

thescientist13:
Down with this, let's do it through #373 first though.  Then this can land into the website through the new API.  💯

Issue #372, opened 2020-06-17
support ejecting webpack configuration (and friends)


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Related to #370, for users that want the most control over their project configuration, Greenwood should support ["ejecting"](https://create-react-app.dev/docs/available-scripts/#npm-run-eject) its Greenwood configuration, and defaulting to that if it's there.  With that, user's can do any sort of loader / plugin configurations they want.

## Details
1. This should be exposed as a new task, called `eject`, e.g. `greenwood eject` as _webpack.config.js_
1. Greeenwood CLI will need to know to know to use the user's webpack configuration
1. It should also support an `--all` flag that will also copy the remaining config files to the user's root
    - _babel.config.js_
    - _postcss.config.js_
    - _.browserslistrc_
1. Add documentation to the CLI section of the [Docs -> Overview](https://www.greenwoodjs.io/docs/) page
1. Add test case

PR #375, opened 2020-06-19
task/issue 372 Add support for ejecting config files including webpack, babel, postcss

hutchgrant:
## Related Issue
Resolves #372

## Summary of Changes
* Add new `greenwood eject` command with `--all` option
* Eject task that copies config files to project root directory

## TODO

* [x] Test case
* [x] Documentation

PR #374, opened 2020-06-18
task/issue 373 add unified preset extensibility through greenwood config

hutchgrant:
## Related Issue
Resolves #373

## Summary of Changes

* Modified `config.js` lifecycle to accept `markdown` object with `settings` and `plugins` key values.
* Modified `webpack.common.js` to utilize the config objects markdown properties
* Added 2 test cases: one for settings, one for preset plugins
* Updated markdown documentation for markdown loader customization configuration information
* Add deps `rehype-slug, rehype-autolink-headings, remark-github` and cite them in greenwood website config

Issue #373, opened 2020-06-17
expose unified presets through greenwood configuration


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
In addition to the principal web languages (HTML, CSS, JavaScript) Greenwood should also support markdown out of the box.  In particular, being able to take advantage of passing [**unified**](https://unifiedjs.com/) compatible presets to [**wc-markdown-loader**](https://github.com/hutchgrant/wc-markdown-loader#unified-presets).

## Details
1.  Add a configuration option called `unified` that should take an object with `settings` and `presets` keys
1. It should "forward" that config to **wc-markdown-loader**
1. Add documentation to [Docs -> Markdown](https://www.greenwoodjs.io/docs/markdown) section detailing unified support and best practices / tips for how to use this in the context of the loader
1. Add test case

PR #376, opened 2020-06-20
task/issue 370 support webpack customization

### LINKED ISSUE #370: Support custom webpack loader configurations
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Currently, if a user wants to add their own webpack plugins, they can do so [through the greenwood.config.js](https://www.greenwoodjs.io/plugins/webpack).  But what if a user wants to modify an existing loader?  I propose allowing user overrides of webpack loaders.  ~~production, development, and common configs. As well as an ability to eject current configs to user's workspace(maybe a separate issue/PR).~~

## Details

Users need to be able to modify webpack configurations for a variety of things.  For example, if I want to add rehype plugins to wc-markdown-loader, I would need access to the webpack config so that I can add:

```js
{
        test: /\.md$/,
        loader: 'wc-markdown-loader',
        options: {
          defaultStyle: false,
          shadowRoot: false,
          preset: {
            settings: { commonmark: true, bullet: '*', emphasis: '*', fences: true },
            plugins: [
              require('rehype-slug'),
              require('rehype-autolink-headings')
            ]
          }
        }
      }, 
```

My Suggestion for accomplishing this is:
- ~~modify context lifecycle to check for if a webpack config exists~~
- ~~modify build/develop tasks to use the compilation.context to decide what webpack config to use~~
- add support for custom webpack loaders (as experimental) as we have support [custom **webpack** plugins](https://www.greenwoodjs.io/plugins/).

A future idea would be to have a CLI command to eject the configs into the user's project directory.

hutchgrant:
## Related Issue
~~Resolves #370~~
Supports the read part of #372 

## Summary of Changes
* Add support for ability to write your own `webpack.config.develop.js` and `webpack.config.prod.js` from within project's directory

## TODO

* [x] Documentation
* [x] Test Case


packages/cli/src/lifecycles/context.js
```diff
@@ -16,6 +17,8 @@ module.exports = initContexts = async({ config }) => {
       const userPageTemplate = path.join(userTemplatesDir, 'page-template.js');
       const indexPageTemplate = 'index.html';
       const notFoundPageTemplate = '404.html';
+      const webpackProd = 'webpack.config.prod.js';
```
thescientist13, line 20: Per my comment here, would it make more sense to pick this up as **webpack**'s default support for _webpack.config.js_?

Or maybe it doesn't matter since we do it programmatically.  Also, I think **webpack** [recommends breaking it up](https://webpack.js.org/guides/production/) actually, so maybe this is fine as is.

www/pages/docs/build.md
```diff
@@ -0,0 +1,13 @@
...
+title: 'Build'
+index: 4
+linkheadings: 3
+---
+
+# Build Configurations
+
+### Webpack
+
+The webpack config for production(`webpack.config.prod.js`) and development(`webpack.config.develop.js`) can be overridden by providing your own custom configuration within your project's directory.  If you wish to revert back to the default provided configuration, simply delete these 2 files from your project's root directory.
```
thescientist13, line 13: Maybe a link to eject docs (when they become available)?

hutchgrant:
I've provided both?

thescientist13:
> I've provided both in 2 differ PRs See #374

Right, which is the source of my question / confusing.  

So looking to the objectives of #370 as defined in the issue, I was thinking a PR for that would have been to support for webpack _loaders_ (similar to our experimental support for webpack _plugins_).  From our conversation, we wanted to provide a progression of support so that if I just want to add TypeScript, a user wouldn't have to eject and override the entire webpack configuration.

I would have expected #372 to handle ejecting _and_ reading custom webpack configurations.  In theory, the tests for ejecting should be to run the `eject` command and then run the build using that ejected config and everything should work the same.  Not sure it makes sense to split up the writing (eject) from the reading.

So should probably scale this PR back to just fit the issue description then there isn't so much overlap between PRs.

hutchgrant:
No we discussed adding a custom configuration in greenwood config specifically for markdown loader only because we are supporting it by default. I believe you mentioned its easier that way then ejecting all webpack, so you suggested enabling both was a good idea and I agreed.

If someone wants to use additional loaders, they can use the eject command.  If they only want to modify the unified presets, they can do that through greenwood config.  I supported both options and adding the same thing as plugins in greenwood config for custom loaders is a third option that I don't recall discussing.  You explained that markdown was supported so we should open up just that loader to the greenwood config.

`Add a configuration option called unified that should take an object with settings and presets keys`

#373

packages/cli/src/lifecycles/context.js
hutchgrant: we do it programmatically, it won't work that way.

thescientist13:
We can review this on our call, but it is all available in the [meeting notes](https://docs.google.com/document/d/1fZWEGsMZ2a58agZIPn74Iu3_w8fFuNRmUH_6C-XuEXA).  The general consensus was
- make markdown (vja unified) a first class citizen since we aim to treat it as core along with HTML, CSS, JS
- expose webpack loaders, like with webpack plugins, for ease of overriding for simple file processing
- if full control is desired, provide an `eject` option

So I did the following regarding traking
1. I chose to adapted #370 to just support loader configurations
1. I created #372 and included reading custom configurations in the details
1. I created #373 for first class unified / markdown support

If it's just that you don't want to do #370 then that's fine.  I can just close this PR and we can just focus on #375 since we don't really need two PRs doing the same thing.  Hopefully we circle back to #370 soon.

PR #381, opened 2020-07-08
style headers and link icon to be consistent throughout pages and sizes

aholtzman:
## Related Issue
was discussed during 7/7/2020, not sure if there is an issue yet.

## Summary of Changes
updated the styling of h2 through h5 elements to align with link icon and consistent sizing of h2 to h6 elements 

[ex.](https://www.greenwoodjs.io/docs/data)
Screen Shot 2020-07-07 at 9 10 43 PM

### DIFF
--- a/www/styles/page.css
+++ b/www/styles/page.css
@@ -6,11 +6,12 @@
   & h3 {
     color: #2D3D3A;
     padding-top: 4rem;
+    font-size: 1.8rem !important;
   }
 
   & h2 {
     color: #2D3D3A;
-    font-size: 2.2rem;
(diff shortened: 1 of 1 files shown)


www/styles/page.css
```diff
@@ -8,7 +8,18 @@
     padding-top: 4rem;
   }
 
+  & h2 {
+    color: #2D3D3A;
+    font-size: 2.2rem;
```
hutchgrant, line 13: I'd like to reduce the size a bit for h2 being `2rem` and h3 being `1.8rem`

Issue #377, opened 2020-07-01
greenwood internal class names are conflicting with vendor (markdown plugin) classnames


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
If you go to certain pages on our [website documentation](https://www.greenwoodjs.io/getting-started/creating-content#development-server), you'll see certain markdown snippets showing some unexpected syntax highlighting.
<img width="1653" alt="Screen Shot 2020-06-30 at 8 03 15 PM" src="https://user-images.githubusercontent.com/895923/86188480-c5c14200-bb0c-11ea-8ed7-bef9d2917a52.png">

## Details
Initial thoughts are that the class names used by the markdown parsers could be conflicting with Greenwood generated classnames in our templates.
<img width="1608" alt="Screen Shot 2020-06-30 at 8 00 07 PM" src="https://user-images.githubusercontent.com/895923/86188530-ea1d1e80-bb0c-11ea-9dc5-38efebd623cd.png">

Maybe we should namespace all our class names within our default [page template](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/templates/page-template.js)
```html
// before
<div class='page-template content'>
  <entry></entry>
</div>

// adfter
<div class='gwd-page-template gwd-content'>
  <entry></entry>
</div>
```

thescientist13:
We should also update our template examples in the docs.

PR #382, opened 2020-07-09
task/issue 356 Extendable App Templates

### LINKED ISSUE #356: extendable app templates
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
As evidenced in the [PR for code splitting](https://github.com/ProjectEvergreen/greenwood/pull/265#issuecomment-559246951) there are some rough developer ergonomics around having to extend [_app-template.js_](), in that you have recreate (copy / paste) the entire file.  Even if all you want to make is a little change like this
```html
<!-- current template -->
<div class='wrapper'>
  MYROUTES
  <lit-route><h1>404 Not found</h1></lit-route>
</div>

<!-- just to add a heater and a footer -->
<div class='wrapper'>
  <eve-header></eve-header>
  MYROUTES
  <lit-route><h1>404 Not found</h1></lit-route>
  <eve-footer></eve-footer>
</div>
```

## Details
I wonder if there's a way to do one of the following?

#### Extend
Simple extend the component and override the `render` method
```js
import { html } from 'lit-html';
import { AppComponent } from `@greenwood/cli/src/templates/app-component';

class GreenwoodAppComponent extends AppComponent {

  render() {
    return html`
      <div class='wrapper'>
        <eve-header></eve-header>
        MYROUTES
        <lit-route><h1>404 Not found</h1></lit-route>
        <eve-footer></eve-footer>
      </div>
    `;
  }
}
```

> Note I tried this and was getting errors  that _/index.index.js_ couldn't be resolved, likely because the app-template is in the user directory, not CLI directory?

#### Slots
Maybe allow users to project / [slot](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot) content into the `<app-component/>?

```js
import { html } from 'lit-html';
import `@greenwood/cli/src/templates/app-component';

class GreenwoodAppComponent {

  render() {
    return html`
      <app-template>
        <div class='wrapper'>
          <eve-header></eve-header>
          MYROUTES
          <lit-route><h1>404 Not found</h1></lit-route>
          <eve-footer></eve-footer>
        </div>
      </app-template>
    `;
  }
}
```

hutchgrant:
## Related Issue
Resolves #356 

## Summary of Changes

* Move app-template _guts_ to base-template.js
* Alias appTemplatePath with @greenwood/app-template within webpack
* Add writeBaseTemplate to scaffold lifecycle - write scratch '.greenwood/app/app.js' with base-template.js
* Modify scaffold lifecycle - write scratch '.greenwood/app/app-template.js' with app-template.js
* modify index.html to use `<base-app></base-app>` instead of `<eve-app></eve-app>`

Now you can simply write an app-template.js without needing all the other underlying logic.

## Note

* `app-template.js` still require MYROUTES magic string (we can change it later to `<routes></routes>`)
* `app-template.js` must have a component name `eve-app` for now
* `app-template.js` must maintain filename and path `<workspace>/templates/app-template.js`

## TODO

* ~~Documentation~~

## Future

* grep the app-template.js component name and insert it into base-template.js
* Allow developer to override `base-template.js` with their own custom one

## Example app-template.js

```js
import { html, LitElement } from 'lit-element';

  render() {
    return html`
      MYROUTES
      <lit-route><h1>404 Not found</h1></lit-route>
    `;
  }
}

customElements.define('eve-app', AppComponent);
```

### DIFF
--- a/packages/cli/src/templates/app-template.js
+++ b/packages/cli/src/templates/app-template.js
@@ -1,99 +1,11 @@
-import { html, LitElement } from 'lit-element';
-import { connectRouter } from 'lit-redux-router';
-import { applyMiddleware, createStore, compose as origCompose, combineReducers } from 'redux';
-import { lazyReducerEnhancer } from 'pwa-helpers/lazy-reducer-enhancer.js';
-import thunk from 'redux-thunk';
-import client from '@greenwood/cli/data/client';
-import ConfigQuery from '@greenwood/cli/data/queries/config';
-import GraphQuery from '@greenwood/cli/data/queries/graph';
-
-// eslint-disable-next-line no-underscore-dangle
-const compose = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || origCompose;
-
-const store = createStore((state) => state,
-  compose(lazyReducerEnhancer(combineReducers), applyMiddleware(thunk))
-);
-
-import '../index/index';
-
-connectRouter(store);
+import { html, css, LitElement } from 'lit-element';
 
 class AppComponent extends LitElement {
 
-  async connectedCallback() {
-    super.connectedCallback();
-    const route = window.location.pathname;
-    const response = await Promise.all([
-      await client.query({
-        query: ConfigQuery
-      }),
-      await client.query({
-        query: GraphQuery
-      })
-    ]);
-    const { config } = response[0].data;
-    const currentPage = response[1].data.graph.filter((page) => {
-      return route === page.link;
-    })[0];
-
-    const currentPageTitleSuffix = !currentPage || currentPage.link === '/'
-      ? ''
-      : ` - ${currentPage.title}`;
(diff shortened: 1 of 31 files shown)


packages/cli/src/templates/app-template.js
```diff
@@ -1,101 +1,13 @@
...
-        <lit-route><h1>404 Not found</h1></lit-route>
+      MYROUTES
+      <lit-route><h1>404 Not found</h1></lit-route>
     `;
   }
 }
 
-customElements.define('eve-app', AppComponent);
\ No newline at end of file
+customElements.define('eve-app', AppComponent);
```
thescientist13, line 13: Maybe this should be `gwd-app` to maintain our naming conventions set in #377 ?  (as a separate PR related to that issue is fine?)

Issue #349, opened 2020-05-07
data client is not rehydrating from existing WINDOW.APOLLO_STATE on initial load


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
As observed in #345, when first loading a production build, where data is set in `WINDOW.APOLLO_STATE` there should be no need for _cache.json_ to be requested via `fetch`, but it is. 😞 
<img width="1656" alt="Screen Shot 2020-05-06 at 8 38 13 PM" src="https://user-images.githubusercontent.com/895923/81241875-8fb68480-8fd9-11ea-841d-8efca0979bf0.png">

## Details
I think the crux of the issue likely due to the fact that although we configure a real `ApolloClient` in [_client.js_](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/data/client.js#L6) with the initial `WINDOW.APOLLO_STATE`
```js
return fetch(`/${root}${rootSuffix}cache.json`)
  .then(response => response.json())
  .then((response) => {
    // mock client.query response
    return {
      data: new InMemoryCache().restore(response).readQuery(params)
    };
  });
```

We don't actually use that client in "SSG" mode so that would make sense why it wouldn't do that.  That said, if this can be solved using an in cache memory, then this might also solve #347 .

thescientist13:
Could we try this?
```js
import { ApolloClient } from 'apollo-client';
import { InMemoryCache } from 'apollo-cache-inmemory';
import { HttpLink } from 'apollo-link-http';

const APOLLO_STATE = window.__APOLLO_STATE__; // eslint-disable-line no-underscore-dangle
const client = new ApolloClient({
  cache: new InMemoryCache().restore(APOLLO_STATE),
  link: new HttpLink({
    uri: 'http://localhost:4000'
  })
});
const backupQuery = client.query;
window.backupCache = {};

client.query = (params) => {

  if (APOLLO_STATE) {
    // __APOLLO_STATE__ defined, in "SSG" mode...
    const root = window.location.pathname.split('/')[1];
    const rootSuffix = root === ''
      ? ''
      : '/';
   const path = `/${root}${rootSuffix}cache.json`;

   if(!window.backupCache[path])
      return fetch(path)
        .then(response => response.json())
        .then((response) => {
           // mock client.query response
           window.backupCache[path] = response;
           return {
[...]

Issue #317, opened 2020-04-11
code splitting / chunking of cache.json files


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
While [code splitting](#265 ) helps with chunking horizontally across routes, for a full graph with hundreds or thousands of pages, a call to the graph would result in huge _cache.json_ file.

We need to think about how to split this up, so _cache.json_ file can be chunked too.

thescientist13:
@hutchgrant 
I think it would be good to edit the description here with what you had shared in Slack.  Added this issue to project 5.

hutchgrant:
Yes I agree. This is a concern we will need to address or scaling larger than 300 pages or using too much data on <100 could slow page load times.

thescientist13:
I wonder if something like [**unified**](https://unifiedjs.com/) would be helpful with either this or #305 ?

thescientist13:
[Tweet from Dan Abramov](https://twitter.com/dan_abramov/status/1251927220030976000)

> _But about a typical app architecture. Such as a fully client side app. **Or a server-rendered app with a poor approach (like a giant JSON blob in HTML blocking interactions)**_.

thescientist13:
Docs page transfers 2.6MB of JSON from _cache.json_ 😞 
<img width="1672" alt="Screen Shot 2020-04-24 at 4 34 53 PM" src="https://user-images.githubusercontent.com/895923/80254670-980dd780-8649-11ea-984c-5af41bf596fa.png">

hutchgrant:
The deepmerge within serialize.js is where we're getting this large cache from.  There are 2 queries in the app-template (one isn't necessary), one in the header, and one in the shelf. That's how you have so many requests being made. Also I don't know if our hydration logic is correct either.

In any event, the deepmerge approach can be tweaked so that maybe we dont include 8 pages worth of queries into a single file?

hutchgrant:
```js
    const response = await Promise.all([
      await client.query({
        query: ConfigQuery
      }),
      await client.query({
        query: GraphQuery
      })
    ]);

```

That graph query is massive and when each page is serialized that will be added to every single cache.  It's only needed once on initial page load for SPA.

thescientist13:
Some thoughts on quick wins here
[...]

PR #383, opened 2020-07-09
Bug/issue 371 similar query calls invariant error

### LINKED ISSUE #371: similar query calls breaking production build with Invariant Violation 8 error
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
In my [website project](https://github.com/thegreenhouseio/www.thegreenhouse.io/pull/137), I have a [page template](https://github.com/thegreenhouseio/www.thegreenhouse.io/blob/technical/issue-133-markdown-for-blog-posts-bak/src/templates/post-template.js) template that is making its own `query` call.
```js
import { html, LitElement } from 'lit-element';
import client from '@greenwood/cli/data/client';
import gql from 'graphql-tag';

import '../components/blog-post/blog-post';
import '../components/footer/footer';
import '../components/header/header';
import '../components/navigation/navigation';
import pageCss from '../styles/page.css';
import '../styles/theme.css';

MDIMPORT;

class PostTemplate extends LitElement {
  ...

  async connectedCallback() {
    super.connectedCallback();
    let route = window.location.pathname;
      query: gql`query {
        graph {
          title,
          link,
          data {
            date,
            image
          }
        }
      }`
    });

    ...
  }

  ...
}

customElements.define('page-template', PostTemplate);
```

Results in an ApolloClient error
```
index.c10096742db3b816782b.bundle.js:1 Uncaught (in promise) Invariant Violation: Invariant Violation: 8 (see https://github.com/apollographql/invariant-packages)
    at new t (http://localhost:8000/index.c10096742db3b816782b.bundle.js:1:17113)
    at l (http://localhost:8000/index.c10096742db3b816782b.bundle.js:1:17334)
    at R (http://localhost:8000/index.c10096742db3b816782b.bundle.js:1:10323)
    at L (http://localhost:8000/index.c10096742db3b816782b.bundle.js:1:13986)
    at t.transformDocument (http://localhost:8000/index.c10096742db3b816782b.bundle.js:98:67607)
    at t.read (http://localhost:8000/index.c10096742db3b816782b.bundle.js:98:65761)
    at t.e.readQuery (http://localhost:8000/index.c10096742db3b816782b.bundle.js:98:45766)
    at http://localhost:8000/index.c10096742db3b816782b.bundle.js:98:72246
```

Screen Shot 2020-06-16 at 8 38 43 PM

## Details
What I think correlation here is that since app-template is also making a `query` call for the graph, and this page template an adapted version of it, when logging _cache.json_, it looks like there are _two_ entries for the graph, one per call.  
Screen Shot 2020-06-16 at 8 40 56 PM

Since it only happens in production, this is likely an issue with our [mock Apollo client](https://spectrum.chat/apollo/apollo-client/optimistic-response-failed-with-invariant-violation-8~9995b7e0-4cdd-42d1-9a50-2b3983079d2d).

thescientist13:
## Related Issue
resolves #371 

1. Created a common utility that will generate a unique hash per query, that can in turn be used to write / read cache file paths in _cache.js_ and _client.js_, respectively
1. Added unit tests and test cases, observed #386 that may result in intermittent test failures on this PR perhaps
1. Removed _cache.json_, left `window.__APOLLO_STATE__`, cleaned up deps.  
    - This effectively invalidates #349 at this point, IMO
    - Would also take care of #317 too then?

~~**Note**: You can test on the home page now and see multiple `query` calls now returning.   However, do keep in mind that pages with a menu will break right now until I add support for variables.  (right now both `menu` queries are using the same keys and so filenames are not unique)~~
Screen Shot 2020-07-08 at 10 21 37 PM

## TODOs
1. [x] Add support for variables (to fix pages with `menu` queries)
1. [x] Add a test case
1. [x] remove `deepmerge` dependency / unified _cache.json_ in **serialize** lifecycle.  All _-cache.json_ files can now stay at the root as well
1. [x] return an actual hash
1. [x] Perf comparison - need to swap out with something other than `crypto`, much too large!
1. [x] remove `console` logs / example in _home-template.js_

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -44,12 +44,10 @@
     "css-loader": "^2.1.1",
     "css-to-string-loader": "^0.1.3",
     "cssnano": "^4.1.10",
-    "deepmerge": "^4.2.2",
     "file-loader": "^3.0.1",
     "filewatcher-webpack-plugin": "^1.2.0",
     "front-matter": "^3.0.1",
     "fs-extra": "^8.1.0",
-    "glob-promise": "^3.4.0",
     "graphql": "^14.5.8",
     "graphql-tag": "^2.10.1",
     "html-webpack-plugin": "^3.2.0",
@@ -74,6 +72,7 @@
     "webpack-merge": "^4.2.1"
   },
   "devDependencies": {
+    "glob-promise": "^3.4.0",
     "rehype-autolink-headings": "^4.0.0",
     "rehype-slug": "^3.0.0"
   }
--- a/packages/cli/src/data/cache.js
+++ b/packages/cli/src/data/cache.js
@@ -24,8 +24,6 @@ module.exports = async (req, context) => {
       const { query, variables } = req.body;
       const queryObj = gql`${query}`;
 
-      console.log('cache query', query);
-
       const { data } = await client.query({
         query: queryObj,
         variables
--- a/packages/cli/src/data/cache.js
+++ b/packages/cli/src/data/cache.js
@@ -1,11 +1,11 @@
 const { ApolloClient } = require('apollo-client');
 const createHttpLink = require('apollo-link-http').createHttpLink;
-const crypto = require('crypto');
 const fetch = require('node-fetch');
 const fs = require('fs-extra');
 const { gql } = require('apollo-server');
 const InMemoryCache = require('apollo-cache-inmemory').InMemoryCache;
 const path = require('path');
+const { getQueryKeysHash } = require('./common');
 
 /* Extract cache server-side */
 module.exports = async (req, context) => {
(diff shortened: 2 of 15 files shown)


hutchgrant:
Does this also partially address #317 with deepmerge gone?

thescientist13:
> Does this also partially address #317 with deepmerge gone

I believe so!

hutchgrant:
Excellent a 2 for 1.  This provide valuable information now on why how we're overfetching. I noticed we're fetching side menu cache query 3 times, navigation twice, greenwood config once, and graph once.  Note the graph and config queries are in the `app-template.js` whereas the others are in the `page-template.js`.  I haven't reviewed this yet just cursory look at the results and it looks promising.

thescientist13:
OK, did some quick performance checks and looks like i need to swap out `crypto` with [something else](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa), or figure out a way to slim it down.  It almost doubles the bundle size!  Nothing fancy, if we even want to obfuscate the name?  🤷‍♂️ 

With
```
-rw-r--r--   1 owenbuckley  staff  551585 Jul 14 09:53 index.c69a6e52fffa6a3bec0f.bundle.js
```

Without
```
-rw-r--r--   1 owenbuckley  staff  244633 Jul 14 09:56 index.8b04ed729af8e894acb4.bundle.js
```

Not a big deal, maybe can use btoa inxtead.

hutchgrant:
Just use https://www.npmjs.com/package/md5  I just tested it on your build its much smaller.

thescientist13:
Giving [this solution](https://gist.github.com/hyamamoto/fd435505d29ebfa3d9716fd2be8d42f0#gistcomment-2775538) a try, with some additional tweaks, using `Math.abs` to ensure unsigned values (the third test for variables was returning a negative value) and then cast `.toString()`.

PR #385, opened 2020-07-11
task/issue 384 add support for custom loading component

### LINKED ISSUE #384: lazy loaded routes have no loading component in SPA
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary

When we lazy load routes in SPA, we have no placeholder component between route changes. This causes as an ugly bug on greenwood's current site where the content disappears, while the header and footer get smashed together. In the initial PR there was a loading circle that was distracting so we removed it. But we never filled that placeholder with something else.  Furthermore, all developer apps will also wish to use this feature(which [lit-redux-router merged](https://github.com/fernandopasik/lit-redux-router/pull/14)). 

## Details

I suggest an overrideable default loading component. It can be overridden the same way page templates are overridden.  It must have the same filename and use the same component element definition. e.g. `<eve-loading></eve-loading>`

hutchgrant:
## Related Issue
Resolves #384

## Summary of Changes

* Create default loading component used as a placeholder between lazy loaded routes 
* Modify scaffold to enable loading component in lazy loaded routes
* Modify scaffold to replace instances of predefined `LDIMPORT` with path to loading component
* Modify context to detect if custom loading component is present

## Notes

* custom loading component must be in `<workspace>/templates/loading.js`
* custom loading component element must be defined as `eve-loading`
* custom app-template must import the component using predefined hook `LDIMPORT`

## TODO

* Documentation
* Test custom/default loading component does indeed render

## Future

* remove need for LDIMPORT hook

### DIFF
--- a/packages/cli/src/config/webpack.config.common.js
+++ b/packages/cli/src/config/webpack.config.common.js
@@ -84,7 +84,8 @@ module.exports = ({ config, context }) => {
     resolve: {
       extensions: ['.js', '.json', '.gql', '.graphql'],
       alias: {
-        '@greenwood/cli/data': context.dataDir
+        '@greenwood/cli/data': context.dataDir,
+        '@greenwood/cli/templates/loading': context.loadingTemplatePath
       }
     },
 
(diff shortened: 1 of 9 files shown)


hutchgrant:
it turns out we had a [style already present in the app-template.js]9https://github.com/ProjectEvergreen/greenwood/blob/master/www/templates/app-template.js#L28) that already had minimum height but we didn't add the `gwd-` prefix to it.

Now that doesn't mean custom loading components shouldn't be added. Just that its not necessary to fix the initial problem.

thescientist13:
closing this for now since a big refactor is underway.  This issue can definitely be re-visited post the hitting of the No Bundle Feature Parity milestone `v0.10.0` release.

PR #388, opened 2020-07-12
filter out index page from children query

### LINKED ISSUE #332: filter out parent page from ChildrenQuery response
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Right now `ChildrenQuery` is returning the [index.md](https://github.com/ProjectEvergreen/greenwood/blob/release/0.5.0/packages/cli/test/unit/data/schema/graph.spec.js#L68) as part of the array of pages returned, which is effectively the parent route.  I think in this case, it doesn't make sense to return the parent if you are asking for the children, or else to render the children as a list, you'll have to exclude the parent, e.g. _${parent}/index.md_

So effectively the resolver should just filter out _index.md_ then.

## Details
Given a structure like this
```shell
pages/
  blog/
    first-post.md
    index.md
    second-post.md
    third-post.md
```

And render it like this
```js
render() {
  return html`
    <ul>
      ${posts.map((post) => {
        return html`
          <li>
            <a href="${post.link}" title="Click to read my ${post.title} blog post">${post.title}</a>
          </li>
        `;
       })}
     </ul>
  `
}
```

You'll have to something like this in your component
```js
class BlogPostsTemplate extends LitElement {
  ...
  
  async connectedCallback() {
    super.connectedCallback();
      query: ChildrenQuery,
      variables: {
        parent: 'blog'
      }
    });

    this.posts = response.data.children.filter(post => {
      return post.filePath.indexOf('/blog/index') < 0;
    });
  }
```

Or else you'll have a list like this, when you probably only want the posts themselves
1. First Post
1. Blog
1. Second Post
1. Third Post

thescientist13:
## Related Issue
resolves #332 

## Summary of Changes
1. Exclude  parent page from `getChildrenFromParentRoute` call
1. Update Specs

## TODO
1. [x] Wait until #383 is merged, then remove need to [filter here](https://github.com/ProjectEvergreen/greenwood/blob/v0.7.2/packages/cli/test/cases/build.data.graph/src/templates/blog-template.js) (which should still allow the test to pass and further confirm the implementation here)

### DIFF
--- a/.github/workflows/chromium-lib-install.sh
+++ b/.github/workflows/chromium-lib-install.sh
@@ -6,5 +6,6 @@ sudo apt-get update \
   && sudo wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add - \
   && sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
   && sudo apt-get update \
-  && sudo apt-get install -y google-chrome-unstable --no-install-recommends \
+  && sudo apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst fonts-freefont-ttf \
+      --no-install-recommends \
   && sudo rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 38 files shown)


thescientist13:
Hmm I wonder if there's a bug here as this is failing now after getting upgraded with master.  
```
  557 passing (7m)
  1 failing

  1) Build Greenwood With: 
       Data from GraphQL
         Blog Page (Template) w/ Navigation and Children Query
           should have expected blog posts links in the <body> tag when using ChildrenQuery:

      AssertionError: expected 0 to equal 2
      + expected - actual

      -0
      +2
      
      at Context.<anonymous> (packages/cli/test/cases/build.data.graph/build.data.graph.spec.js:169:40)

```

Might be related to `ChildrenQuery` and variables since `Graph` query in the header seems to be fine.

thescientist13:
All set now 👍

PR #395, opened 2020-07-21
fix/issue 395 missing style prefix

### LINKED ISSUE #394: missing gwd prefix in style affecting page minimum height
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Bug
- [ ] Other (please clarixy below)

## Summary
As noted in #384 the page currently jumps when content disappears due to the content wrapper not having a minimum height.  The style is already present in app-template but we forgot to rename it when we added gwd prefixes in #378

## Details
* Fix remaining missing style prefix in app template

hutchgrant:
## Related Issue
Resolves #394 

## Summary of Changes

* fix missing style prefix in app-template.js which sets minimum height for website content

### DIFF
--- a/.github/workflows/chromium-lib-install.sh
+++ b/.github/workflows/chromium-lib-install.sh
@@ -6,5 +6,6 @@ sudo apt-get update \
   && sudo wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add - \
   && sudo sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
   && sudo apt-get update \
-  && sudo apt-get install -y google-chrome-unstable --no-install-recommends \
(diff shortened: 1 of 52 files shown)


hutchgrant:
Well we can't/shouldn't assume the style a developer uses on their page-templates.  They either have to know the page content will collapse due to route changes, or we simply implement the default loading component which does in contain the minimum height and does not necessitate anything from the developer.

thescientist13:
> They either have to know the page content will collapse due to route changes

But I think that is also part of the issue, is that they don't know, and there's nothing to let them know.  So if we take the path of not assuming any default styles, then we should at least document, IMO.

hutchgrant:
Not that documenting it is bad, but maybe both solutions are ideal(loading component + documentation).

At least with the loading component, developer doesn't need to read the docs until they want to change the loading component behaviour.

thescientist13:
Looks like this was fixed as part of #382 
https://github.com/ProjectEvergreen/greenwood/commit/7bbdb175c099de9f51237cc471241601c2bdde9c#diff-24f3d821c68ad700464a69cc9c93aef8

hutchgrant:
Good eye, I forget I added that to the new app-template

Issue #357, opened 2020-06-04
html minification


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
To help save even more bytes by reducing overall _index.html_ size, we should minify all HTML.  Related to the conversation [here](https://github.com/ProjectEvergreen/greenwood/pull/265#discussion_r433007702)

Also, anywhere we are using `\n` inside (tagged) template literals, we should remove them since template literals already preserve line breaks, so adding the `\n` just adds more space.

## Details

Or is this really worth it?

thescientist13:
Not sure if this is something we need to do?  Seems like it would be better handled by the CDN or hosting provider.  May end up closing this.

thescientist13:
actually, this one could still be useful, but probably not critical for 1.0

jstockdi:
Proposing a conservative, opt-in take on this that I think addresses the earlier hesitation, with a PR to follow shortly.

## Design

A new package, `@greenwood/plugin-minify-html`, as a standard resource plugin using the existing `shouldOptimize` / `optimize` lifecycle (the same seam `plugin-standard-html` already uses), applying two transforms to each built HTML page:

1. **Strip HTML comments** — except comments matching a preserve list, which defaults to `[/^!/, /^\/?lit-/]` so that Lit SSR hydration markers (`<!--lit-part ...-->` / `<!--/lit-part-->`) and license-style `<!--!` comments survive. Overridable via an `ignoreCustomComments` option.
2. **Collapse whitespace runs** (`[ \t\r\n\f]+` → single space) in text nodes — skipping `<pre>`, `<textarea>`, `<script>`, `<style>`, and `<noscript>` subtrees entirely. Runs always collapse to *one space*, never removed outright, so inline-element gaps (`<span>a</span> <span>b</span>`) and rendering semantics are untouched — this is exactly the collapsing CSS already performs, applied at build time instead of paint time.

Deliberately **not** doing the aggressive tricks (attribute quote removal, optional tag omission, inline JS/CSS minification — the bundler already runs terser over scripts). That keeps the transform ~50 lines and hard to get wrong.

## Why this shape

- **Zero new dependencies in the tree** — implemented with `node-html-parser`, already a `@greenwood/cli` dependency.
[...]

Issue #354, opened 2020-05-31
Progressive Enhancement ("RedactJS")


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of exploration into https://github.com/ProjectEvergreen/greenwood/issues/305 and conversations during our recently weekly meetings, it seems like there was something we might have overlooked along the way on this project, and that we designed it SPA first, rather than SSG first.  (IMO)

![76916456-d8ce4000-6896-11ea-8818-7ec52aaea1e2](https://user-images.githubusercontent.com/895923/83361024-8fe53e00-a353-11ea-91cc-58dd3450e0a5.png)

I would like to open this issue to evaluate how efficiently we are shipping / bundling the output of Greenwood.  I would like to see if there are ways the minimum bundle size could be mitigated even further.  A website shouldn't really _need_ all this [fancy SPA magic](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/templates/app-template.js), at least not by default and in the final bundle.
```js
import { html, LitElement } from 'lit-element';
import { connectRouter } from 'lit-redux-router';
import { applyMiddleware, createStore, compose as origCompose, combineReducers } from 'redux';
import { lazyReducerEnhancer } from 'pwa-helpers/lazy-reducer-enhancer.js';
import thunk from 'redux-thunk';

// eslint-disable-next-line no-underscore-dangle
const compose = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || origCompose;

const store = createStore((state) => state,
  compose(lazyReducerEnhancer(combineReducers), applyMiddleware(thunk))
);
```

As much as possible a Greenwood site by default should work without all the JS needed to _render_ the site, and otherwise make it fall back gracefully.  JavaScript should be seen as a progressive enhancement opted-into by the user.

> It should be interest to note that if you disable the _index.js_ bundle from loading on the main site, which is 528 KB (granted about half of that is due to #305 ) the whole site still works except for the fading in / out text on the home page and the shelf sub menu clicks.  At least the shelf would probably still work if using native click handler and `HTMLElement`.... Just some food for thought 🍖 🤔 💭  
[...]

Issue #321, opened 2020-04-11
improvements and optimizations for webpack CLI configuration


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Just wanted to track a general collection of enhancements / improvements I've been wanting to make

## Details
Kind of grab bag at the moment, but hopefully these are quick one liners implementations.  Or else if they take too long, they can spun off into their own issues.
1. IE11 support not needed by default now that custom configs can be provided?
1. add [`output.futureEmitAssets`](https://github.com/ProjectEvergreen/greenwood/issues/317#issuecomment-612520691)
1. **babel [preset-modules](https://babeljs.io/blog/2020/03/16/7.9.0)**
1. Make sure webpack is emitting / configured specifically for ES2015+ by default
1. Use [TerserPlugin](https://webpack.js.org/plugins/terser-webpack-plugin/) instead of UglifyJS (if not already enabled)
1. Review browserslist config for best modern browser support + [`Promise`](https://twitter.com/_developit/status/1229174005233569797)?
1. [polyfills on demand?](https://twitter.com/smashingmag/status/1229343311426719744)
1. evaluate [**mini-html-webpack-plugin**](https://www.npmjs.com/package/mini-html-webpack-plugin)
1. Check setting for [`ModuleConcatPlugin`](https://twitter.com/acutmore/status/1233716201144057856)
1. [**Brotli**](https://github.com/netlify/ask-netlify/issues/24#issuecomment-633256071) compression available in Netlify

thescientist13:
linking an article for further review
https://github.com/iamakulov/awesome-webpack-perf

thescientist13:
As part of #355 / #417 webpack wont be used in the project anymore, so this issue will become invalid once this gets merged and release.

thescientist13:
v0.10.0 released, webpack related issues no longer apply to the project

Issue #305, opened 2020-03-18
Dedup content in JavaScript


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
One [concern with a pre-rendering approach](https://twitter.com/benawad/status/1239615944298815488) is that even though the HTML is pre-generated, it often still ends up with an architecture that ships the same code via JavaScript as well.  
![ETP_TAQXsAIcldW](https://user-images.githubusercontent.com/895923/76916456-d8ce4000-6896-11ea-8818-7ec52aaea1e2.png)

If the project is generating static content it really shouldn't _also_ be delivered via JavaScript too, ideally. 

> This should also be taken into context with #299 and #265 

## Details
For example, in our project, since all pages are generated by JavaScript, you'll see that even static the page content is in the [index.js bundle](https://www.greenwoodjs.io/index.bbee30c9e4fd3879072a.bundle.js).
```js
        <h2>Community</h2>
<p>Greenwood understands the role of community in open source, technology, and learning and we want to embrace that in this project as well.  There are many great tools and projects available in the web ecosystem and we want to highlight those projects here, that have been important in the development of Greenwood (in no particular order).</p>
<ul>
<li><a href="https://webpack.js.org/"><strong>webpack</strong></a> (and friends) - Module bundler and powers the local development environment for Greenwood.</li>
<li><a href="https://github.com/GoogleChrome/puppeteer"><strong>puppeteer</strong></a> - Headless Chrome browser used for serializtion.</li>
<li><a href="https://github.com/hutchgrant/wc-markdown-loader/"><strong>WC Markdown Loader</strong></a> - A webpack plugin that transforms a markdown file into a Web Component! ðŸ¤¯</li>
<li><a href="https://github.com/fernandopasik/lit-redux-router"><strong>LitReduxRouter</strong></a> - Declarative client side routing library.</li>
</ul>
<p>And of course we can't forget our <a href="https://github.com/ProjectEvergreen/greenwood/graphs/contributors">Contributors</a>!</p>
<p>So please come join our community and help us grow and advance the web together!  We can't wait to see what you come up with.  ðŸ’¡</p>
```

Wondering if there's anything about our implementation that should change re: pre-rendering vs SSG?

- A pretty interesting discussion from the React side of things [here](https://twitter.com/rauchg/status/1235408113223602182) 
[...]

PR #396, opened 2020-07-21
Rfc/issue 354 strict / SSG mode

thescientist13:
## Related Issue
in support of #354, introduces a "strict" mode that strips out all client side JavaScript for projects that don't need any "runtime" JavaScript

1. Introduced modes configuration option 
    - **ssg** - ships NO javascript (has some caveats)
    - **spa** - current behavior 
1. **spa** mode marks the `<script>` tag as `async`

Note this neat trick with [inline `<script>` tag](https://github.com/ProjectEvergreen/greenwood/pull/401)

> Posted some initial performance results [here](https://github.com/ProjectEvergreen/greenwood/pull/396#issuecomment-662150746).  In both cases, performance is improved, even if Greenwood website's performance score is low.  The hope is that enabling **progressive** mode could solve this, but probably an issue just for Lighthouse score should be ,ade.

## Questions
1. [x] Should we change the name to something other than mode, which [we might need for this](https://github.com/ProjectEvergreen/greenwood/issues/199)?
    - yes, will call it **optimization**

## TODO
1. [x] initial SPA benchmark / performance report
1. [x] rename **ssg** to **strict** mode (think of another name for **spa** mode?)
1. [x] Make ~~**strict**~~ **spa** mode default (the long term hope is that **progressive** mode would be the new default)
1. [x] Testing
    - [x]  **strict** as default for smoke test and test cases
    - [x] error case
    - [x] add unit tests and update `Config` query
    - [x] add test case for **spa** mode
1. [x] add documentation
1. [x] Refactoring
    - [x] Use [`async` or `defer`](https://discourse.mozilla.org/t/async-v-s-defer/53819/4)?  I think in our case `defer` would work well for us?  (avoid jankiness while hydrating mid load)
      - [x] See if **html-webpack-plugin** can do `async` setting?  (Not sure if `async`/ `defer` will work when serializing though...)
      - [x] page bundles are included as `<script>` tags in index.html - should be `async` too, or removed in **strict** mode?
    - [x] Don't render `*.bundle.js` script tags in **strict** mode
    - [x] Don't render `<lit-redux-router>` in **strict** mode (shaved off 5KB from _index.html_ for Greenwood website!)
1. [x] Rename from **mode** to **optimization**
1. [x] Optimize the 404 page (nice to have, follow up issue) - done
1. [x] Final benchmark / performance report for both modes
    - Strict 
    - SPA
1. [x] Create an issue for tracking getting performance close to 100 for Greenwood / all users - made #411
   -  ~~Would like to look more into why our index bundle is so large or is it as optimized as it can be, for SPA mode.  (maybe split out the webpack runtime?  Might include reviewing the merits of our default babel config too.~~
   -  ~~related issues:~~ #357 , #321, #305, #235
1. [x] Explore an issue for migrating Greenwood website to **strict** mode - #410


thescientist13:
Here are some initial benchmarks from local testing in Greenwood and my personal website.  Will complete with some stats of current performance in productions

## www.greenhouse.io
### Before
TODO

### After
greenhouse io-ssg-local-network-js
greenhouse io-ssg-local-lighthouse
greenhouse io-ssg-local-network


## Greenwood
TODO

thescientist13:
A couple things that are also interesting to observe is that production is already pretty bad, but none of the things having to do with the size of our JavaScript will fix things, seemingly?
> These suggestions can help your page load faster. They don't directly affect the Performance score.

The issues with LCP, TTI, TBT are seemingly a result of other causes then?

Naturally, strict / ssg mode passes with flying colors as can be seen in the above comment.

For comparison though, my website is doing somewhat better than greenwood production website though
greenhouseio-network
greenhouseio-lighthouse

Interestingly, the Getting Started repo, which is bare bones, fares surprisingly better out of all of them?  (Which I guess is expected, but why such a sharp drop off?)
getting-started-network
getting-started-lighthouse

-----

So perhaps the issues here are related to 
- something specific to our app code that isn't being done in the Getting Started? 
- would be good to see if commenting out some of the bundled code would help improve the score?  (like if bundle size was closer to 100KB?)
- Maybe we should use more `HTMLElement` where possible?

hutchgrant:
* comment out external font from `style/theme.css`
* comment out  analytics, polyfills, markdown plugins from `greenwood.config.js`

In spa mode I noticed an immediate reduction in **Total Blocking Time** when testing lighthouse in mobile mode (where the problem is).  Which resulted in an increase score of performance 67.  Not great but an improvement none the less.

in ssg mode those changes resulted in a performance score of 100 when testing lighting house in mobile mode.

hutchgrant:
nearly a perfect lighthouse score on mobile and desktop. One issue I noticed, our "get started" button is now green again.

PR #398, opened 2020-07-22
task/issue 276 update dependencies

### LINKED ISSUE #276: update browserslist version
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Browserslist is being a squeaky wheel and requesting we update it.
```
Browserslist: caniuse-lite is outdated. Please run next command `yarn upgrade`
```

## Details
It's actually breaking tests (addressed as part of #115) because of [this](https://github.com/browserslist/browserslist/issues/361) and it writing to `stderr`, but we should at least fix it.   In addition, I think it's important for development to see it, but probably not users.  We should also add the environment export to the `bin` command in the _package.json_ for `@greenwood/cli` so it doesn't bother users at least.

hutchgrant:
## Related Issue
Resolves #276

## Summary of Changes

* update package.json
* update cli/package.json
* update lockfile


hutchgrant:
It's not that simple unfortunately.  I tried updating just babel, just postcss-preset-env and neither was the single cause of the bug. yarn upgrade everything and bug was gone.

PR #402, opened 2020-07-28
enhancement/issue 400 switch to internal fonts

### LINKED ISSUE #400: site uses external fonts, should use internal fonts instead
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary

The greenwood site uses external fonts. We should download and use the fonts internally and prevent any additional external source from slowing down the initial page load. This may improve performance with this small optimization.

## Details

Fonts within `theme.css` and `components/shelf/shelf.css` could instead be fetched internally.  Google fonts can be downloaded and hosted.

hutchgrant:
## Related Issue
Resolve #400

## Summary of Changes

* switch theme.css import to use local font
* switch shelf/shelf.css import to use local font


hutchgrant:
yea google didn't provide woff2 fonts in the "download font family" link. I'll have to do it another way.

hutchgrant:
With woff2 internal font I saw and 8% performance gain in lighthouse test score locally.

www/assets/fonts/SourceSansPro.css
```diff
@@ -0,0 +1,63 @@
+/* cyrillic-ext */
+@font-face {
+  font-family: 'Source Sans Pro';
+  font-style: normal;
+  font-weight: 400;
+  font-display: swap;
+  src: url('/assets/fonts/SourceSansPro/SourceSansPro-Regular.ttf') format('truetype');
```
thescientist13, line 7: still seeing a _.ttf_ here?

hutchgrant:
I'll do one better we can support IE6 - IE8 https://google-webfonts-helper.herokuapp.com/fonts/source-sans-pro?subsets=latin

hutchgrant:
By adding these additional charsets and styles, you'll notice the fonts appear bolder. I think this is a fix rather than a bug. Our heading elements have a font weight which wasn't reflected correctly by original font. The downside is the fonts are larger.

hutchgrant:
Okay woff2, woff, and ttf  for android,safari and ios.

hutchgrant:
Definetly

hutchgrant:
Yes we can do that here as well.

PR #415, opened 2020-08-14
change the project evergreen logo in the header

aholtzman:
## Related Issue
Add New Logo to header - no related issue

## Summary of Changes
1. [x] add new project evergreen logo to header
1. [x] adjust styling for better logo placement


www/assets/evergreen.svg
```diff
@@ -0,0 +1,45 @@
+<svg version="1.0" xmlns="http://www.w3.org/2000/svg"
```
thescientist13, line 1: Do you think we could indent the tags here?

Issue #355, opened 2020-05-31
bundle / build less development


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of explorations into #305, continued into #354, and inspired by projects like [**Vite**](https://github.com/vitejs/vite), [**es-dev-server**](https://open-wc.org/developing/es-dev-server.html), and [**Snowpack**](https://www.snowpack.dev/), it would be of interest to investigate a way to move away from bundling by default for development.  webpack could still be used for production builds maybe, or maybe even something that is more ESM friendly, like [**Rollup**](https://rollupjs.org/guide/en/).

> Note I tried **Snowpack**, but it has a hard dependency on bootstrapping from a _package.json_, as opposed to being just more of an opinionated dev server like **Vite**. 

## Details
This would definitely require a lot of new work, but in a ESM first world, there are so many advantages in that the entire site wouldn't need to be prebuilt / bundled, and transformations could just happen on the fly + on demand.  Plus all syntax could just be standardized to use `import` for everything.

Not opposed to coming up something of our own, maybe as its own package?  Also, simplifying this and using something like Rollup may allow us to not have to rely on having to webpack-ify everything, and also free up a ton a `devDependencies` in the `@greenwood/cli` package.

Would allow us to explore usage of [import-maps](https://github.com/WICG/import-maps) as a possible solution (bare imports and routing in general), [ex](https://github.com/thescientist13/git-explorer/blob/master/src/client/index.html).
```html
<script>
{
  "imports": {
    "lit-element": "./vendor/lit-element/lit-element.js",
    "lit-html": "./vendor/lit-html/lit-html.js",
    "lit-html/lit-html.js": "./vendor/lit-html/lit-html.js",
    "lit-html/lib/shady-render.js": "./vendor/lit-html/lib/shady-render.js",
    "lit-html/directives/unsafe-html.js": "./vendor/lit-html/directives/unsafe-html.js"
  }
}
</script>
```

Some more specific webpacky things that we use webpack for that would have to be redone:
- dev server w/ live reload and file watching
- NormalModuleReplacementPlugin for mapping imports to user directories
- templating _index.html_
- bundling (just use Rollup?)

thescientist13:
Could even push it farther with [web packaging](https://twitter.com/justinfagnani/status/1196115567142260737) for production

[...]

Issue #185, opened 2019-08-26
[RFC] Plugins: Transforms


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
Taken from a comment in https://github.com/ProjectEvergreen/greenwood/issues/17#issuecomment-523706157 in regards to a general discussion around the API and capturing the specific action item re: the `lifecycle` plugin type here.

## Details
This would essentially be enabling potentially different source file types in addition to the default markdown support, like [JavaScript](#170)  or maybe YAML?

At minimum, should probably make sure to support at least _.html_ files out of the box.

thescientist13:
Looks like we have a new paradigm and opportunity to solve this as part of the work being done #417 with Koa.  I think just trying to find the common denominator in terms of things in common when need to resolve / transform files.  Things like
* filename - e.g. a _.ts_ file would need to become a _.js_ file for the browser
* source / contents - typescript would need to become javascript
* others?

Rather than pass the whole `app` / `ctx` object around, specify a (testable) API between the Greenwood CLI and consumers that can hopefully account for 90% of use cases.  I think some shared code will likely come up between them (like referencing `context` or DOM / string manipulation) so feel free to come up with those.  Plugin developers could be encouraged to use those as well for their own plugin offering!

I would say out of the box Greenwood should support all web languages and some friends * out of the box, everything else would be community plugins / packages
- HTML
- CSS
- JavaScript
- PNG, JPG, etc image formats
- SVG
- Markdown *
- _node_modules_ *

Ideally plugins should be decoupled as much as possible.  So given the above, I would expect at least 6 different transforms.

thescientist13:
Roadmap
1. [ ] Default Tranform Plugins
1. [ ] External existing plugins
1. [ ] Other plugin examples - TypeScript, others?

thescientist13:
Had a couple more thoughts on technical direction here that we could achieve as part of trying to integrate our plugins into this system

### TransformInterace Constructor
I think we should change the signature of the `TransformInterface` constructor to accept a request and compilation object, and then a third object with all the transform options, e.g.
```js
// tranform.interface.js
module.exports = class TransformInterface {

  constructor(request, compilation, { extensions = [], contentType = '' }) {
    const { context, config } = compilation;
    this.extensions = extensions; // ['.foo', '.bar'] 
[...]

PR #419, opened 2020-10-09
Rfc/issue 355 organize serve lifecycle

hutchgrant:
## Related Issue
#355 

## Summary of Changes
* Organize serve.js lifecycle so its more easily navigable and managable.


thescientist13 (commented):
Nice, thanks for getting the ball rolling on this, as getting this file in order is definitely a good task for us to start tackling.  Let me put this on the agenda for our upcoming meeting so we can discuss the design together.

For now I would say I think we should look to take it one step further, and make each of these filters (scoped by file / resource type their intended for) into their own files into a directory and brought in via `require`.

Ideally this should also enable users to provide their own, but I think out of the box Greenwood should handle the "web" by default, plus some close friends
- html
- css
- js
- md
- node_modules
- svg
- images (png, jpg, gif, etc)

We can make Greenwood plugin packages for other favorites.   (e.g. TypeScript, SCSS, etc).  I guess this would be effectively be our [Transforms API](https://github.com/ProjectEvergreen/greenwood/issues/185)?

thescientist13 (approved):
Left some thoughts from our discussion in #185 for next steps part two for this.

Issue #278, opened 2020-01-17
GraphQL (Apollo) data as its own package


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Should `@greenwood/cli/data`, e,g. _package/cli/data_ be it's own distinct package? 

## Details
There's also the case that it can be its own package (to maintain its own test cases, dependencies, etc) but have it be a direct dependency of `@greenwood/cli`.  Essentially they would be functionally coupled, but benefit by having code logically organized.

Also, for now, we can just alias it as `@greenwood/data` and that way, if it becomes a package or not, it won't break user land code since the `import` path will stay the same.

thescientist13:
I think coming out of #355 this will happen organically.

thescientist13:
Re-purposing to help solve #418 
1. Make data a plugin and something lighter than apollo?
1. update all usages to use `@data/client` instead of `fetch` in _www/_
1. See if [data hydration still an issue](https://github.com/ProjectEvergreen/greenwood/issues/349)
1. [reduce duplicate graphql calls](https://github.com/ProjectEvergreen/greenwood/issues/272)?

PR #449, opened 2020-12-18
Bug/issue 386 graph not matching filesystem order

### LINKED ISSUE #386: nested pages are not returned in order as they are on the filesystem
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Getting intermittent test failure similar to the issue observed in #271 , in that the order of pages returned did not seem to returned from the graph in the same order as they are on the filesystem, in particular when there are multiple files in nested directories.  I notice our test case does not act and trying to do a map in the template, e.g.

While it is always true that users can do their filtering and sorting client side, Greenwood should at least to a consistent, default, predictable return order.

## Details
For example, when working on [the PR](https://github.com/ProjectEvergreen/greenwood/pull/383) for #371 I created a directory structure for the graph test case, [similar to my website](https://github.com/thegreenhouseio/www.thegreenhouse.io/tree/master/src/pages/blog) in that it had nested directories like below
```shell
src/
  index.md
  pages/
    blog/
      first-post.md
      second-post.md
    index.md
```

And in a template, listing the posts on the page
```js
class BlogTemplate extends LitElement {
  ...

  async connectedCallback() {
    super.connectedCallback();
      query: ChildrenQuery,
      variables: {
        parent: 'blog'
      }
    });

    this.posts = response.data.children.filter(post => {
      return post.filePath.indexOf('/blog/index') < 0;
    });
  }

  /* eslint-disable indent */
  render() {
    const { posts } = this;

    return html`

      <div class='container'>
          <ul>
            ${posts.map((post) => {
              return html`
                <li>
                  <a href="${post.link}" title="Click to read my ${post.title} blog post">${post.title}</a>
                </li>
              `;
            })}
          </ul>
      </div>
    `;
  }
  /* eslint-enable */
}

customElements.define('page-template', BlogTemplate);
```

However, intermittently, i would get this error
```shell
551 passing (5m)
  4 failing

  1) Build Greenwood With:
       Data from GraphQL
         Home (Page Template) w/ Navigation Query
           should have a expected NavigationQuery output in the <header> tag:

      AssertionError: expected '/blog/second-post/' to equal '/blog/first-post/'
      + expected - actual

      -/blog/second-post/
      +/blog/first-post/

      at Context.<anonymous> (packages/cli/test/cases/build.data.graph/build.data.graph.spec.js:114:57)

  2) Build Greenwood With:
       Data from GraphQL
         Blog Page (Template) w/ Navigation and Children Query
           should have expected navigation links in the <header> tag tag when using NavigationQuery:

      AssertionError: expected '/blog/second-post/' to equal '/blog/first-post/'
      + expected - actual

      -/blog/second-post/
      +/blog/first-post/

      at Context.<anonymous> (packages/cli/test/cases/build.data.graph/build.data.graph.spec.js:156:57)

  3) Build Greenwood With:
       Data from GraphQL
         Blog Page (Template) w/ Navigation and Children Query
           should have expected blog posts links in the <body> tag when using ChildrenQuery:

      AssertionError: expected '/blog/second-post/' to equal '/blog/first-post/'
      + expected - actual

      -/blog/second-post/
      +/blog/first-post/

      at Context.<anonymous> (packages/cli/test/cases/build.data.graph/build.data.graph.spec.js:175:57)

  4) Build Greenwood With:
       Data from GraphQL
         Blog Post Pages (Post Template) w/ custom Graph Query
           should have expected navigation links in the <header> tag tag when using NavigationQuery:

      AssertionError: expected '/blog/second-post/' to equal '/blog/first-post/'
      + expected - actual

      -/blog/second-post/
      +/blog/first-post/

      at Context.<anonymous> (packages/cli/test/cases/build.data.graph/build.data.graph.spec.js:219:56)
```

----

Interestingly, there is a nested test case, but it doesn't have multiple files in the nested directory, like described above.  I would start by updating that test case to something like
```shell
src/
  index.md
  pages/
    blog/
       2019/
        first-post.md
        second-post.md
      2020/
        first-post.md
        second-post.md
        third-post.md
    index.md
```

And return in the following order
1. _/blog/2019/first-post.md/
1. _/blog/2019/second-post.md/
1. _/blog/2020/first-post.md/
1. _/blog/2020/second-post.md/
1. _/blog/2020/third-post.md/
1. _/blog/index.md/

thescientist13:
## Related Issue
resolves #386 (and incorporates #455 / #331 somewhat) , as I have noticed it is happening in my personal project still, and fairly consistently- https://github.com/thegreenhouseio/www.thegreenhouse.io/issues/174#issuecomment-744641842

## Summary of Changes
1. Made `walkDirectory` logic synchronous, which makes this issue go away at least (async programming is hard 😞 )
1. Bulked up the Nested Directory test case with a bunch of fixtures to ensure this gets solid test coverage
1. Cleaned up page object pushed into the graph just to prune properties that aren't so relevant anymore
1. Made sure `route` is normalized and "safe" to use and always end with `/`, solves #331 

> _As an observation, I think as part of the data work, we should avoid remapping properties in graph.json and GraphQL.  For example in graph.json it's label, in GraphQL it's `id`.  We should just mirror them unless there's a reason they can't, IMO.  Something I can do as part of #278 and its accompanying documentation updates._

## Nice To Have
1. [ ] Try and get it working asynchronously, though I don't want that to be show stopper, as I am happy to make a refactoring issue to track finding an asynchronous solution (and at least have a working solution until then)
1. [ ] refactor `walkDirectory` to not rely on `context`, so it can move out of the scope of `module.exports`

edit: tracking both of the above in Trello for post 1.0 efforts / optimizations


packages/cli/src/transforms/transform.html.js
```diff
@@ -33,7 +33,7 @@ class HTMLTransform extends TransformInterface {
         let body = await getAppTemplate(barePath, this.workspace);
         body = await getAppTemplateScripts(body, this.workspace);
         body = getUserScripts(body, this.workspace);
-        body = getMetaContent(url, this.config, body);
+        body = getMetaContent(url.replace('index.html', ''), this.config, body);
```
thescientist13, line 36: this was sending URLs w/ _index.html_ in the `url` so this strips that to keep nice clean URLs, e.g. 

www.example.com/about/index.html 

is now 

www.example.com/about/

----

although this makes more sense to do in `getMetaContent`?

thescientist13: edit x2: actually, looks like I don't need it there, the tests pass without it, so I must have fixed the issue that caused this.  😎

PR #451, opened 2020-12-20
add github action for release branches

thescientist13:
## Related Issue
N / A

## Summary of Changes
I think this should solve the issue with `release/xxx` branches not building in GitHub Actions.  🤞 
https://github.com/ProjectEvergreen/greenwood/pull/436
Screen Shot 2020-12-20 at 1 03 34 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,7 +25,7 @@ jobs:
         node-version: ${{ matrix.node-version }}
     - name: Installing project dependencies
       run: |
-        yarn install  --frozen-lockfile && yarn lerna bootstrap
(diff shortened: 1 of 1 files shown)


thescientist13:
Not sure if this has to be merged into master?

thescientist13:
It seems that it should work even if not merging into the master branch, and so as long as this workflow is present in the branch it might act on, then it should work.  I suppose worst case scenario is to open this to master as well.

thescientist13:
Cool, looks like it worked!
https://github.com/ProjectEvergreen/greenwood/actions?query=workflow%3A%22Release+Branch+Integration%22
Screen Shot 2020-12-20 at 1 16 10 PM

I should set it under branch protections as well.

Issue #432, opened 2020-11-08
ensure correct working order of website and expected styles (e.g. shelf and scrolling, docs, markdown plugins, etc)


thescientist13:
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
As part of #355 and #418 , there were some warts left into the website, like:
1. [ ] click areas for the shelf loading and chevron expansion
1. [ ] make sure heading scrolling works
1. [ ] make sure custom markdown plugins from _greenwood.config.js_ are working
1. [ ] Noticed `#### Examples` headings are showing up in side bars - https://github.com/ProjectEvergreen/greenwood/pull/452#issuecomment-750407404
1. [ ] generally make sure everything matches www.greenwoodjs.io as it did before

## Details
Basically make sure the website is still working / looking as expected, and fix what isn't working.  Should save this for the end.

thescientist13:
Noticed the Plugins table styles are off a little bit.
<img width="1179" alt="Screen Shot 2020-12-30 at 11 19 43 AM" src="https://user-images.githubusercontent.com/895923/103366579-033d6280-4a91-11eb-8971-9ad0ff3fb8c1.png">

I think in my effort to keep the styles still working in an un-nested manner until PostCSS could land, I may have changed some of the site structure.  Maybe good if someone could just double check that because once #426 lands, everything should be back to normal from a tooling perspective.

thescientist13:
>  Noticed #### Examples headings are showing up in side bars - #452 (comment)

I think this may be caused by our commented out code still being picked up by the auto headings markdown plugin?

PR #452, opened 2020-12-22
Docs/issue 430 update technical documentation to reflect current project status

thescientist13:
## Related Issue
resolves #430 

> _**Note**: this captures the current state of things.  Going forward PRs will need to update docs as functionality and features are restored._

## Summary of Changes
1. Lerna doesn't bump _www/packages,json_ on custom pre release versions, had to bump myself fyi
1. Updates _Docs_ and _Getting Started_ sections to reflect current state of the project (including marking things as **Coming Soon** and commenting out the docs where the feature is missing now)
1. Left _Plugins_ section alone since there is. #450 open now, where we should start documenting all new features going forward as PRs come in for them
1. Removed need to escape `$` and ``` in code fences (it just works now 🥳 )
1. Deleted "dead" code / files
1. Removed code, docs, and test for **webpack** specific configuration: `devServer.host` and `publicPath`
1. Left a comment on thoughts about layout and organization in https://github.com/ProjectEvergreen/greenwood/issues/325#issuecomment-749832718
1. Added some technical design docs to CONTRIBUTING.md
1. fixed some typos, probably added some new typos, lol

## TODO
1. [x] Confirm environment variables are captured correctly in guides, after unescaping
1. [x] need to document deprecation of `config.publicPath` / `config.devServer.host` that were **webpack** specific features
1. [x] Report bug around how `<link>` tags are detected.  Current assumption is wrong, as there is no closing `</link>` tag involved - created #457 
1. [x] Review correct side bar headings / nav structure - https://github.com/ProjectEvergreen/greenwood/pull/452#issuecomment-750407404 (tracked as part of https://github.com/ProjectEvergreen/greenwood/issues/432)
1. [x] Proof read, etc

Issue #428, opened 2020-11-07
final release and questions checklist


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of #355 and #418 , here are some low hanging things to track.

## Details
1. [x] Keep **colors** package? - might be a good issue to back in if we can do it internally
1. [x] Make sure communicate Node v12 minimum dependency
1. [x] Is a _theme.css_ still needed now from the CLI specifically? - No
1. [x] Make sure all TODOs / lifecycles are addressed and code is cleaned up
1. [x] Rollup refactoring TODOs
1. [x] use only named exports for plugins - will make a discussion
1. [x] delete _templates/_ dir - N/A
1. [x] Make sure all tests are re-enabled
    - markdown + code examples
    - revisit smoke tests (hello?) / consolidate / put in default? / 404
    - `describe.only` / `it.only` / `xit` 
    - make sure all tests are deleted or turned on
    - enable `checkCoverage` / `include` config
    - unit tests for transforms
    - eject
    - replace `beforeEach` w/ `beforeAll`
1. [x] Alpha testing in getting started and www.thegreenhouse.io
1. [x] import maps as a plugin?  or default for now, should check caniuse.com
1. [x] final performance and [build time benchmarking](https://twitter.com/slightlylate/status/1335299502652080130)
1. [x] Migration Path (for Release Notes)
     - rename a _<xxx>-template.js_ => _<xxx>.html_
     - before / after for getting started
     - `<entry>` -> `<content-outlet>`
     - use forward slashes for all paths
     - _theme.css_

Issue #112, opened 2019-07-04
improve the responsiveness and user experience for mobile sidenav / shelf


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Currently the left hand sidenav (on pages like About and Getting Started) although stacks nicely on mobile, with longer lists, this pushes all the content down and in turn can lead to extra and more cumbersome scrolling for users.  ex.
<img width="534" alt="Screen Shot 2019-07-04 at 11 38 40 AM" src="https://user-images.githubusercontent.com/895923/60677891-c3bc9880-9e50-11e9-8db0-7343cb4e4531.png">

## Details
Some thought
1. What to do at medium / tablet breakpoints?  How long should menu text be?  Should it collapse into a sidebar on tablet?  What do about the links / text?
1. Maybe on mobile, the shelf should overlay / dropdown from the top and only show as a tab / slideout.  This would prevent content being displaced / bumped down and up and each time it opens.

thescientist13:
We should  also try and improve the experience of the link target for the expand / collapse arrow
[inconsistent-shelf-expand-collapse-simple.mov.zip](https://github.com/ProjectEvergreen/greenwood/files/4561923/inconsistent-shelf-expand-collapse-simple.mov.zip)

thescientist13:
Should also try and handle underline alignment for links
https://github.com/ProjectEvergreen/greenwood/pull/335#pullrequestreview-403068722

thescientist13:
will be carried out in a discussion first in https://github.com/ProjectEvergreen/greenwood/discussions/503

PR #459, opened 2020-12-30
Fix/issue 432 correct shelf and markdown plugins

hutchgrant:
## Related Issue
#432

## Summary of Changes

* Fix shelf not expanding properly
* Fix page headers not being linked (markdown plugins not being executed in unified.js)
* Fixed `<ol>`
* Fixed missing link icons on headers (order of markdown plugins + merged postcss from release/0.10.0 branch)
* Fixed inline code block highlighting
* brought back our markdown config tests from master
* Fixed linkHeadings not being respected because not using our previous graphql query
* Fixed expanded subitems not linking to the menus page + anchor, rather than page it was initially clicked on

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -29,8 +29,8 @@ module.exports = {
   markdown: {
     plugins: [
       '@mapbox/rehype-prism',
-      'rehype-autolink-headings',
       'rehype-slug',
+      'rehype-autolink-headings',
       'remark-github'
     ]
   }
--- a/packages/cli/src/lifecycles/graph.js
+++ b/packages/cli/src/lifecycles/graph.js
@@ -77,6 +77,12 @@ module.exports = generateGraph = async (compilation) => {
               // parse markdown for table of contents and output to json
               customData.tableOfContents = toc(fileContents).json;
               customData.tableOfContents.shift();
+
+              // parse table of contents for only the pages user wants linked
+              if (customData.tableOfContents.length > 0 && customData.linkheadings > 0) {
+                customData.tableOfContents = customData.tableOfContents
+                  .filter((item) => item.lvl === customData.linkheadings);
+              }
             }
             /* ---------End Menu Query-------------------- */
 
--- a/packages/cli/src/plugins/resource/plugin-standard-html.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-html.js
@@ -202,14 +202,15 @@ class StandardHtmlResource extends ResourceInterface {
           });
 
           // TODO extract front matter contents from remark-frontmatter instead of frontmatter lib
+          const settings = config.markdown.settings || {};
           const fm = frontmatter(markdownContents);
           processedMarkdown = await unified()
-            .use(remarkParse) // parse markdown into AST
+            .use(remarkParse, settings) // parse markdown into AST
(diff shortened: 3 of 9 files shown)


thescientist13 (commented):
Awesome!

Do we know if any of these other known issues are now solved too?   (Maybe will need #453 to gets merged and rebased here?)
1. https://github.com/ProjectEvergreen/greenwood/pull/452#issuecomment-750407404
1. https://github.com/ProjectEvergreen/greenwood/pull/452#pullrequestreview-559284867

Basically just curious if the website now has feature parity with itself or if any other work is needed here to achieve that?

thescientist13:
Oh, additional bonus points.  With this fix, we should be able to re-enable these test cases now?
Screen Shot 2020-12-30 at 10 28 56 AM

thescientist13:
@aholtzman 
In our shift from nested -> un-nested -> back again as temporary work arounds, it seems like I may have put the code through a round of telephone, and little styles (or the markup) need to be put back into place to better reflect what we originally had (which should all work in theory).  Wasn't sure if that's something you could help out with here?  

Examples are in the parent ticket for reference.

The only thing I did want to discuss was the click targets for the shelf.  I think the prior working action was clicking the shelf label controlled the routing _and_ the expand / collapse of sub items.  Not sure how good this is for UX, so I localized the expand / collapse to just the arrow.  Open to suggestions here on how best to approach this interaction.

hutchgrant:
* fixed <ol>
* fixed missing link icons on headers (order of markdown plugins + merged postcss from release/0.10.0 branch)
* fixed inline code block highlighting

hutchgrant:
Should we only allow link icons to be displayed on h2, h3 ?

thescientist13:
> Should we only allow link icons to be displayed on h2, h3 ?

I think it was `<h3>` before, as we used `<h2>` to start every doc page.

thescientist13 (approved):
Looks good, and not super urgent (nothing we can't resolve as part of #428 ) but when using `yarn serve` pages with the scroll still render a little bit pre scrolled.  (header is cut off)
Screen Shot 2020-12-31 at 12 07 47 PM


Not sure if it's the _index.html_ being shown in the browser URL, e.g. _docs/index.html_ that may be tripping up the scroll or if it's related to just some CSS?  

Other minor thing: If you expand a shelf item and click one of its children, it wont load that section of the docs.  For example go to docs and expand markdown, and go to click one of the sub children.  

Not a big deal on either of these, as the click action of the shelf is something we can tack onto #112 or #428 .

thescientist13:
@hutchgrant 
Before I merge, just want to confirm we think we know what's going on with the header offset issue i'm seeing here? (even after merging prod server url)
https://github.com/ProjectEvergreen/greenwood/pull/459#pullrequestreview-560520742

I can push to another issue but ideally just want to make sure we have an answer for it either way.

thescientist13 (commented):
Found that there was a bit of an opportunity to refine the scroll logic.  In production, it looks like `<a>` tags were being passed through the `filter`
Screen Shot 2021-01-20 at 9 41 59 PM

Made a [small tweak](https://github.com/ProjectEvergreen/greenwood/pull/459/commits/96d3ba4c1575bfa2d4d2ae52f631f0bc547539b0) and now everything is groovy 💯

Issue #463, opened 2020-12-31
use pretty urls in mock prodServer


hutchgrant:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
In production server, currently we have URLs  being rewritten to end with index.html e.g. `/docs/index.html`. Since we're using pretty URLs by default in development and production, our mock prodution server should also use pretty URLs.

## Details

Modify server. js not to redirect links ending with `/` to `index.html` and instead read both .html and / ending links then return the body with the request files.

PR #464, opened 2020-12-31
Fix/issue 463 pretty urls in prodServer

hutchgrant:
## Related Issue
#463

## Summary of Changes
* remove url redirect from prodServer in serve lifecycle.
* refactor prodServer slightly

### DIFF
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -77,7 +77,7 @@ function getProdServer(compilation) {
     }
 
     if (url.endsWith('.js')) {
-      const contents = await fsp.readFile(path.join(outputDir, ctx.request.url), 'utf-8');
+      const contents = await fsp.readFile(path.join(outputDir, url), 'utf-8');
 
       ctx.set('Content-Type', 'text/javascript');
       ctx.body = contents;
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -75,7 +75,7 @@ function getProdServer(compilation) {
     }
 
     if (ctx.request.url.endsWith('.html')) {
-      const contents = await fsp.readFile(path.join(outputDir, ctx.request.url, 'utf-8'));
+      const contents = await fsp.readFile(path.join(outputDir, ctx.request.url), 'utf-8'));
 
       ctx.set('Content-Type', 'text/html');
       ctx.body = contents;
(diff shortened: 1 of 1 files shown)


packages/cli/src/lifecycles/serve.js
```diff
@@ -69,7 +69,9 @@ function getProdServer(compilation) {
     const { outputDir } = compilation.context;
 
     if (ctx.request.url.endsWith('/')) {
-      ctx.redirect(`http://localhost:8080${ctx.request.url}index.html`);
+      const contents = await fsp.readFile(path.join(outputDir, ctx.request.url, 'index.html'), 'utf-8');
```
thescientist13, line 72: Does combining the `/` and `.html` tests not work well together? 
```js
if (ctx.request.url.endsWith('/')) {
  const contents = await fsp.readFile(path.join(outputDir, ctx.request.url, 'index.html'), 'utf-8');
  ctx.set('Content-Type', 'text/html');
  ctx.body = contents;
}
```

Or I suppose it could also make sense to keep `.html` as a stand alone handler as well?

thescientist13 (approved):

packages/cli/src/lifecycles/serve.js
hutchgrant: right I think need to keep it in case?

thescientist13: > right I think need to keep it in case?

in case of what though?  I wasn't sure if you had a case in mind, so the question was mostly for you.  ;)

What cases do we have of needing to process _.html_ differently?  I could at least see treating both with the same logic perhaps?
```js
if (ctx.request.url.endsWith('/') || ctx.request.url.endsWith('.html')) {
  const barePath = url.endsWith('/') ? '' : ctx.request.url;
  const contents = await fsp.readFile(path.join(outputDir, barePath, 'index.html'), 'utf-8');

  ctx.set('Content-Type', 'text/html');
  ctx.body = contents;
}
```

Mostly just spitballing at this point, so no big deal.

hutchgrant: lets go with that then. I don't have an immediate use case are we resolving all custom html files to their own routes and prettifying them ?  e.g. docs.html would just be /docs/ ? I don't know I havent tested that yet.

thescientist13: yeah, give it a test.  At the end of the day they both _return_ the contents of an HTML file, it's just that we can do it simply without including a `redirect` in either case.

hutchgrant:
did a quick refactor as well.

Issue #435, opened 2020-11-08
resolve (deeply) nested relative template paths to expected workspace path


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
In a user's workspace, given a structure like this
```
src/
  components/
    header.js
    footer.js
  pages/
    one/
      two/
        three/
          index.md
    index.md
  layouts/
    page.html
```

Users could reference files in their templates one of two ways
1. using an "absolute" relative path`/`
1. using a "true" relative path `../`

Right now, only the first option is supported / documented *
```html
<head>
  <script type="module" src="/components/header.js"></script>
  <script type="module" src="/components/footer.js"></script>
</head>
```

_* Technically the first option is supported but only if you are at 1 level of nesting as `../` can't break out of the web root, but once you go further, you start resolving the path relative to the **browser's URL**, leading to URLs like this will fail when resolved to the actual filesystem_
```sh
/one/two/components/header.js
```

## Details
Ideally it would be great if Greenwood could support the `../` as deeply as needed, which would then allow editors and IDEs be able to do the directory resolution and autocomplete for users as well as not breaking path lookups to the workspace.
```html
<head>
  <script  type="module" src="../components/header.js"></script>
  <script  type="module" src="../components/footer.js"></script>
</head>
```

What is nice about this, is that editors can resolve this for users and do autocomplete lookup, which is really nice.  This also aligns with how paths are done in ESM, e.g.
```js
import something from '../utils.js';
```

Would also be good to resolve / address the following
- [handling paths in templates, e.g. `./, ../`](https://github.com/ProjectEvergreen/greenwood/issues/273)
- [or normalize everything to `/`](https://github.com/ProjectEvergreen/greenwood/issues/331)

thescientist13:
I think as part of #466 , we can now support relative paths at least via the Koa server.

However, for Rollup, we would need to resolve this because this being referenced from the _docs/_ directory?
```
  Error: ENOENT: no such file or directory, open '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/www/docs/styles/theme.css'


  Error: ENOENT: no such file or directory, open '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/www/docs/styles/page.css'
```

Should be relatively easy to resolve I think though.

thescientist13:
[...]

PR #466, opened 2021-01-19
Rfc/issue 185 resources refactor

### LINKED ISSUE #185: [RFC] Plugins: Transforms

thescientist13:
## Related Issue
resolves #185 - a new take on an existing design as started in #450 

## Summary of Changes
1. Introduced the concept of a Resource plugin, and the `ResourceInterface` base class (f.k.a Transforms) with various lifecycles.  
1. Built in "standard" resource plugins for Greenwood for like JS, CSS, HTML / md, etc.  (note: given the tight coupling and shared logic, merged _.md_ and _.html_ back into one)
1. Refactored Koa middleware logic a bit (and more options possible using Koa Cascading middleware)
1. There is a TypeScript like example for a _.foo_ resource you can see on the home page
1. Restored / refactored Google Analytics and Polyfills plugins
1. Added tests and documentation

> _I know there is a lot of ground to cover here in this PR, so will go through a full demo / walk through during our next weekly meeting._

## Notes / Thoughts
- For GraphQL, we should now have the ability to handle `@greenwood/data` and return JS for _.gql_
  - will likely just need to expose a server's plugin API now
- For TS - _.foo_ should provide a basic POC

##  TODOs 
- [x] check to see if this new resolve logic now solves / supports #435 (for its own PR) - left a comment, will need to give Rollup a little help here
- [x] move HTML serialize teardown stuff into optimize hook in standard HTML plugin
- [x] Need at least one example of `intercept` - `importMap`?
- [x] remove example _.foo_ code from the website - you can still the same functionality in the test cases for plugins
- [x] fix unfinished comment in docs - _Note: you can always_
- [x] More issue / Trello tracking
  - do we need to add more font / icon types?
  - refactor `devServer` to use standard plugins now?
  - refactoring out parts of **StandardHtmlResource** plugin like `importMap`, `liveServer` into their own?
  - should Greenwood standard plugins be pushed to plugins array?


packages/cli/src/config/rollup.config.js
```diff
@@ -37,9 +37,28 @@ function greenwoodHtmlPlugin(compilation) {
...
-        return '';
+    async load(id) {
+      const extension = path.extname(id);
+      
+      if (extension === '.html') {
+        return Promise.resolve('');
+      }
+
+      // handle custom user file extensions
+      const customResources = compilation.config.plugins.filter((plugin) => {
```
hutchgrant, line 48: Nice!

hutchgrant, line 48: Nice! Was curious how we were going to implement this in rollup.  This is exactly whats needed.

packages/cli/src/lib/resource-interface.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  // handle an already resolved / served resource
+  // eslint-disable-next-line no-unused-vars
+  shouldIntercept(url, headers) {
+    return false;
+  }
+
+  // eslint-disable-next-line no-unused-vars
+  async intercept(contents, headers) {
```
hutchgrant, line 42: Interesting idea +1

packages/cli/src/lib/resource-interface.js
```diff
@@ -0,0 +1,62 @@
...
+    return Promise.resolve(contents);
+  }
+
+  // handle a (final) resource type post build, pre optimize, 
+  // ex: remove es shim <script>, convert .ts -> .js and update path references 
+  // this is only an _index.html_ file, BYOA (Bring Your Own AST)
+  // eslint-disable-next-line no-unused-vars
+  shouldOptimize(contents, url) {
```
hutchgrant, line 50: Perfect for post-serialize plugin functionality

packages/cli/src/lifecycles/serialize.js
```diff
@@ -1,44 +1,53 @@
...
             const outputPath = `${outputDir}${route}index.html`;
-            
-            // TODO allow setup / teardown (e.g. module shims, then remove module-shims)
-            let htmlModified = html;
-  
-            // TODO should really be happening via plugins or other standardize setup / teardown mechanism
-            htmlModified = htmlModified.replace(/<script src="\/node_modules\/@webcomponents\/webcomponentsjs\/webcomponents-bundle.js"><\/script>/, '');
-            htmlModified = htmlModified.replace(/<script type="importmap-shim">.*?<\/script>/s, '');
-            htmlModified = htmlModified.replace(/<script defer="" src="\/node_modules\/es-module-shims\/dist\/es-module-shims.js"><\/script>/, '');
-            htmlModified = htmlModified.replace(/<script type="module-shim"/g, '<script type="module"');
```
hutchgrant, line 29: :100:  agree with this refactor.

packages/cli/src/lifecycles/serve.js
```diff
@@ -1,61 +1,109 @@
...
+    // Greenwood default standard resource plugins
+    pluginResourceStandardCss.provider(compilationCopy),
+    pluginResourceStandardFont.provider(compilationCopy),
+    pluginResourceStandardHtml.provider(compilationCopy),
+    pluginResourceStandardImage.provider(compilationCopy),
+    pluginResourceStandardJavaScript.provider(compilationCopy),
+    pluginResourceStandardJson.provider(compilationCopy),
+
+    // custom user resource plugins
+    ...compilation.config.plugins.filter((plugin) => {
```
hutchgrant, line 28: My only concern here is we're no longer accepting pre-transform(resource) plugins and we're defaulting them all to pos transforms.  I don't have an immediate pre-default use case but for now I think this is okay.  Also another issue might be that nobody can override the default transforms(resources).

hutchgrant (approved):
I had got used to calling these `transforms` plugins `¯\_(ツ)_/¯`

That modification to rollup config is perfect, I was thinking people were going to have to include their own configuration with their plugins or something. The additional intercept and optimize capability is a good idea as well.

My only gripe is with no pre-transform(resource) or default override capability.  This could be addressed at a later time.

thescientist13:
> My only gripe is with no pre-transform(resource) or default override capability. This could be addressed at a later time.

Yeah, and I think once there are some clear use cases for that, I think it will make it easier to understand what a good design for that could look like.  It _could_ be just as easy as running `shouldServe` before and after maybe?
```js
  // then handle serving urls
  app.use(async (ctx, next) => {
    // run user plugins

    // run greenwood plugins

    // run user plugins again?
  });
```

Or perhaps cascading middleware is better for this, but certainly solving in its own PR will allow us to drill down into the design and come up with something good once we have that clear use case.

----

> edit: also I'm not sure if I'm in agreement with the combined markdown and html resource. When I wrote #450 and attempted to decouple them, I recall it being slightly difficult because they both needed to make use of app/page templates.

Maybe I misunderstand your comment, but given my experience is the same as yours (decoupling meant more logic spread out in multiple places, specifically `shouldServe` type logic and thus more bugs), I also found keeping them as one actually makes our code less susceptible to errors as a result and easier to reason about (or at the very least you only have to be looking in one place).  So not sure how that makes a case for splitting them up?

PR #471, opened 2021-02-03
Server Plugin API

### LINKED ISSUE #470: extend a server api so that developers can call custom server plugins
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary

Calling a server during the develop.js command would be handy for people who want to use say a database, graphql, or liverload/hmr during development.

## Details

Add a ServerInterface for use with server plugins. The plugin `"type: server"` can be declared in the `greenwood.config.js` and then will be called during the develop.js command and potentially in production as well.

Move livereload into a seperate server plugin, for now as an example.

hutchgrant:
## Related Issue
Resolves #470

## Summary of Changes
* Adds a new Server Plugin API similar to resource plugins, similar plugin definition 
* Utilizes `ServerInterface`
* Plugins declared with `type: "server"`
* develop.js command plugin will call `start() ` function from plugin when running greenwood in development mode
* livereload is used as an example of how to implement

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -1,6 +1,7 @@
 const path = require('path');
 const pluginGoogleAnalytics = require('./packages/plugin-google-analytics/src/index');
 const pluginPolyfills = require('./packages/plugin-polyfills/src/index');
+const liveReloadServer = require('./packages/cli/src/plugins/server/plugin-livereload');
 
 const META_DESCRIPTION = 'A modern and performant static site generator supporting Web Component based development';
 const FAVICON_HREF = '/assets/favicon.ico';
@@ -24,7 +25,8 @@ module.exports = {
     pluginGoogleAnalytics({
       analyticsId: 'UA-147204327-1'
     }),
-    pluginPolyfills()
+    pluginPolyfills(),
+    liveReloadServer()
   ],
   markdown: {
     plugins: [
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -1,7 +1,6 @@
 const path = require('path');
 const pluginGoogleAnalytics = require('./packages/plugin-google-analytics/src/index');
 const pluginPolyfills = require('./packages/plugin-polyfills/src/index');
-const liveReloadServer = require('./packages/cli/src/plugins/server/plugin-livereload');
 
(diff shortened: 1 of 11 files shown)


thescientist13 (commented):
Nice, this is definitely a great start!

One thought here is that if we can extract the [livereload `<script>` tag snippet](https://github.com/ProjectEvergreen/greenwood/blob/v0.10.0-alpha.4/packages/cli/src/plugins/resource/plugin-standard-html.js#L91) out of our standard HTML plugin, we could completely encapsulate all the related functionality.  🎉 
  
So the `export` could be something like this:
```js
class LiveReloadServer extends ServerInterface {
  ...
}

class LiveReloadResource extends ResourceInterface {
  ...
}

module.exports = (options = {}) => {
  return [{
    type: 'server',
    name: 'plugin-live-reload-server',
    provider: (compilation) => new LiveReloadServer(compilation, options)
  }, { 
    type: 'resource',
    name: 'plugin-live-reload-resource',
    provider: (compilation) => new LiveReloadResource(compilation, options)  }];
};
```

And _greenwood.config.js_ would look something like this.
```js
const pluginLiveReload = require('./packages/cli/src/plugins/server/plugin-livereload');

module.exports = {
  plugins: [
    ...pluginLiveReload({ /* options */ }) // notice the spread! ...
  ]
};
```

I'm sure it could get more elegant with some extra thought, but this definitely gets the job done for now and is pretty effective, IMO.

What I really like about being able to combine it all into one is that later on, I would like to introduce a [**HotModuleReplacement (HMR**) plugin](https://webpack.js.org/concepts/hot-module-replacement/), and thus allow user's the option to choose which developer experience they would like, would be 👌 


----

On a related note about extracting LiveReload from the CLI down the road, is that we will likely want to deprecate `devServer` from _greenwood.config.js_ since likely each server plugin will have its own unique configuration needs.  I think this is good though, since it cleanly moves things out of core and into plugins, which is quite nice.  😃

thescientist13 (approved):
Added some docs and refactored to make **plugin-livereload** a Greenwood default server plugin (for now at least).

I will rebase this branch against the `release/0.10.0` branch, but will let my changes stand until we can review during this week's meeting, at which point I think then this will be good to merge, barring any major objections.

PR #465, opened 2021-01-06
issue 429 javascript module imports

hutchgrant:
## Related Issue
#429

## Summary

This is a test PR demonstrating the issue related to javascript module imports of unsupported libraries

## Problems

* problems related to javascript library imports for unsupported libraries e.g. redux, lodash, (for this example)

lodash
```
Uncaught (in promise) SyntaxError: The requested module 'blob:http://localhost:1984/ec43201d-db76-4dd3-943a-02002e7c8fb9' does not provide an export named 'default'
async function (async)
topLevelLoad @ es-module-shims.js:252
processScripts @ es-module-shims.js:453
(anonymous) @ es-module-shims.js:440
(anonymous) @ es-module-shims.js:472

```

redux  - related to a transform path?

```
es-module-shims.js:409 Uncaught (in promise) Error: Unknown Content-Type ""
    at es-module-shims.js:409

```

Issue #426, opened 2020-11-07
restore Babel / PostCSS / Browserslist for no bundle development


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of #355 and #418 , need to track restoring support or Babel, PostCSS, and Browserslist.  Though perhaps these should be plugins?

## Details
Will need to support the above three
1. [ ] Production _and_ development mode?
1. [ ] Restore nested CSS in _www/_

PR #472, opened 2021-02-04
Enhancement/issue 434 expand script style link tag production bundling

### LINKED ISSUE #434: handle bare imports in HTML <script> tags when running build command
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Coming out of #417, need to handle this case
```html
<script type="module">
  import '@evergreen-wc/eve-container';
</script>
```

or maybe even
```html
<link href="/node_modules/prism/prism.css" rel="stylesheet"></link> 
```

## Details
Otherwise getting this error in the console
Screen Shot 2020-11-08 at 12 34 30 PM

thescientist13:
## Related Issue
resolves #434 

I gotchu fam

Screen Shot 2021-02-03 at 8 45 32 PM
Screen Shot 2021-02-03 at 8 45 55 PM

## Summary of Changes
1. Refactored _rollup.config.js_ to align on a single html parsing library
1. Now the below type of inline `<script>` tag usage will get properly bundled by Rollup and inlined back into the HTML
    ```js
    <script type="module">
      import '@bare/specifier';
    <script>
    ```
1. Added test cases and fixed a rollup `<script>` tag matching bug - (also seen in https://github.com/ProjectEvergreen/greenwood/pull/475/files#diff)
1. Updated docs with notes on a recommendation to use file based style and script tag usage

## TODO
1. [x] I think JavaScript handling breaks on multiple usages of `<script src="...">` in the same document... (looks like this is due to a bug I saw happening inhttps://github.com/ProjectEvergreen/greenwood/pull/475/files#diff-57634df7e2b1ba3ea0031d8adc1e73230ca3c93bbecfdea7b44d00c7e100ec8b as well)
1. [x] Tests (think I will add this the _import_ case added in #465 once it gets merged )
1. [ ] ~~Would like handle similar case for inline CSS, e.g.~~ - can't do this easily, just yet.  how to avoid Puppeteer added inline styles?
    ```css
    <style>
      @import '../some.css'
      .my {
         .nested-styles {
          }
      }
    </style>
    ```
1. [x] Documentation
1. [x] Clean up intermediate emitted (scratch) files
1. [x] pick `Buffer` or `crypto`

_Nice To Have_
1.  [x] Would like to _also_ support / verify this, but will likely just push this off to another issue for the 1.0 release
    ```html
      <script type="module" src="@bare/specifier"></script>
      <link rel="stylesheeet" href="@bare/specifier"/>
     ```
1. [x] avoid commented out code - this seems to happen already?

## Summary / Thoughts / Takeaways
So for where this stands as of now, the general (to be) documented recommendation will be to use a file based entry point for all your scripts and styles, for anything anything doing any "heavy lifting" as this will afford the most opportunity for full and consistent bundling and optimization (at this time).  

The mean reason is that since inline `<script>` and `<style>` don't initiate any fetch / network requests, the ability for the Greenwood server to intercept these and run tranforms and optimizatoin's is limited and so may lead to different outcomes between `build` and `develop` commands.  This is something we could try and support, but I think can come at a later time.  For now, we can at least do node_modules directly, which is pretty great.  

Here are some examples of what is / isn't supported.
```html
<!-- this will work -->
<script type="module" src="/your/scripts/main.js"></script>
<script type="module" src="/node_modules/some-package/index.js"></script>
<script type="module">
  import 'some-lib'; // bare import from node_modules

  // inline JavaScript works just fine!
  window.onload = () => {
     const route = window.location.pathname.split('/')[1];
        
     document.getElementsByTagName('app-shelf')[0].setAttribute('page', route);
  }
</script>

<style>
  p {
    color: red;
  }
</style>

<link rel="stylesheet" href="/your/styles/main.css"></link>
<link rel="stylesheet" href="/node_modules/some-package/styles.css"></link>


<!-- this will NOT work -->
<style>
  /* @import only works in a stylesheet? */
  @import url('./some.css')
  
  /* cant transpiling, like when using PostCSS for nested styles */
  p {
    & span {
      color: red;
    }
 }
</style>
```

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -33,7 +33,6 @@
     "cssnano": "^4.1.10",
     "es-module-shims": "^0.5.2",
     "front-matter": "^4.0.2",
-    "htmlparser2": "^4.1.0",
     "koa": "^2.13.0",
     "livereload": "^0.9.1",
     "markdown-toc": "^1.2.0",
--- a/packages/cli/src/config/rollup.config.js
+++ b/packages/cli/src/config/rollup.config.js
@@ -298,9 +298,7 @@ function greenwoodHtmlPlugin(compilation) {
 
     // crawl through all entry HTML files and map bundled JavaScript and CSS filenames 
     // back to original <script> / <link> tags and update to their bundled filename in the HTML
-    generateBundle(outputOptions, bundles) {
-      const mappedBundles = new Map();
-      
+    generateBundle(outputOptions, bundles) {      
       for (const bundleId of Object.keys(bundles)) {
         const bundle = bundles[bundleId];
 
@@ -327,16 +325,6 @@ function greenwoodHtmlPlugin(compilation) {
 
                 if (facadeModuleId && facadeModuleId.indexOf(pathToMatch) > 0) {
                   newHtml = newHtml.replace(src, `/${innerBundleId}`);
-                } else {
-                  // TODO better testing
-                  // TODO no magic strings
-                  if (innerBundleId.indexOf('.greenwood/') < 0 && !mappedBundles.get(innerBundleId)) {
-                    // console.debug('NEW BUNDLE TO INJECT!');
-                    newHtml = newHtml.replace(/<script type="module" src="(.*)"><\/script>/, `
-                      <script type="module" src="/${innerBundleId}"></script>
-                    `);
-                    mappedBundles.set(innerBundleId, true);
-                  }
                 }
               }
             }
--- a/packages/cli/src/config/rollup.config.js
+++ b/packages/cli/src/config/rollup.config.js
@@ -81,21 +81,16 @@ function greenwoodWorkspaceResolver (compilation) {
     name: 'greenwood-workspace-resolver',
     resolveId(source) {      
       // TODO better way to handle relative paths?  happens in generateBundle too
-      if ((source.indexOf('./') === 0 || source.indexOf('/') === 0) && path.extname(source) !== '.html' && fs.existsSync(path.join(userWorkspace, source))) {
-        const resolvedPath = source.replace(source, path.join(userWorkspace, source));
-        
-        return resolvedPath;
+      if ((source.indexOf('./') === 0 || source.indexOf('/') === 0) && path.extname(source) !== '.html' && fs.existsSync(path.join(userWorkspace, source))) {        
+        return source.replace(source, path.join(userWorkspace, source));
       }
 
       // handle inline script / style bundling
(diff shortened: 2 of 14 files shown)


thescientist13:
Had some more time to play around with this and the CSS side of things in particular, and I think there may be some caveats to keep in mind with supporting these two use cases in particular.
```html
<script type="module>
  import '@some/code';
</script>

<style>
  div {
     a {
       color: blue;  /* notice nested selectors */
     }
  }
</style>
```

because none of this will get handled by Koa, since it won't trigger a network request, I think it might be counter intuitive to run Babel / PostCSS on this code in Rollup since this would _not_ be something users would be experiencing during development (for instance, nested selectors would not work), and so I think this introduces conflicting outcomes between  development and production, which I think would not be a good workflow to enable.

That said, I think we can still apply basic optimizations like minification (**terser** / **cssnano**) for free on these that we could put into the `optimize` functions our of standard CSS / JS resources like I started to do in #465.

The more I've been thinking about this, the more I don't think we should be calling PostCSS (or Babel) directly in _rollup.config.js_ either, as I don't want to couple these tools, since there is more options than these tools for optimizing code and so makes a much better case for plugins to take up this task (e.g. `optimize`).  I actually plan on doing some refactoring as part of #426 by making both PostCSS and Babel standalone plugins, and refactor our _rollup.config.js_ accordingly.  We can move **cssnano** to the `optimize` function of our standard CSS plugin, and thus we can actually push all the PostCSS plugins out of the CLI too!

----

That said, I could definitely see us supporting these use cases fairly easily still...
```html
<script type="module" src="@evergreen-wc/eve-container">
<script type="module" src="@package/dist/index.js">

<link rel="stylesheet" href="@bootstrap/bootstrap">
<link rel="stylesheet" href="@bootstrap/bootstrap/dist/bootstrap.css">
```

But I would like to merge #465 first for it's refactoring, and then wrap up #426 , since it would introduce even more refactoring.

thescientist13:
So in the case here, where we can potentially have different bundling / optimizations in `build` and `develop` based on that develop is mostly driven by network requests, like `import`, which uses `fetch` under the hood, I think the best place to leave them at is on equal footing.

We can certainly make it work such that **plugin-resource-html** can scan for inline `<style>` or `<style>` tags and reach out the user and other internal plugins to process the inline code content.  Once we can do that, I think we can make sure to promote / ensure that Rollup will do the same.  Added as a note for myself to track, though not sure if it's something we should aim for 1.0.0, or post 1.0.0.

Issue #304, opened 2020-03-18
<meta> Component package


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Coming out of #303 it makes sense to expose a standalone "`<meta>`" component with a general purpose API that could be used a la [**react-helmet**](https://www.npmjs.com/package/react-helmet).

## Details
Something like this, ideally opting for a declarative API as a custom element (see #299 for reference)
```html
<meta .title="${title}" .meta="${meta}"/>
```

- `title` would be a string for the page's `<title>`
- `meta` would be an Array of `<meta>` tags

> Interesting [tweet](https://twitter.com/garybernhardt/status/1240024841895305216) for consideration

In addition, we can now tap into the graph as part of `ResourceInterface` which means we can get _per page_ data, really really great for SEO.  🤩 

## Questions
How should we package this, as a standalone package to publish to NPM?  If so, will it then need to be a dependency of `@greenwood/cli`?  

Or maybe there's a plugin we could use to pull this in for users through a transitive dependency?  Does that mean it's not default?

Could also have this all done via SSR as Resource with an interceptor?

PR #474, opened 2021-02-13
restore default meta tags and update website

### LINKED ISSUE #446: default <meta> tags are missing from default app template in 0.10.0-alpha.0
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Compared to [previous _app-template.js_](https://github.com/ProjectEvergreen/greenwood/blob/v0.9.0/packages/cli/src/templates/index.html), looks like this default meta is missing
```html
<meta charset='utf-8'>
<meta name='viewport' content='width=device-width, initial-scale=1'/>
<meta name='mobile-web-app-capable' content='yes'/>
<meta name='apple-mobile-web-app-capable' content='yes'/>
<meta name='apple-mobile-web-app-status-bar-style' content='black'/>
```

## Details
Should be restored to new [_app.html_ template](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/cli/src/templates/app.html#L7) and some tests should probably be added [to the default test case](https://github.com/ProjectEvergreen/greenwood/tree/release/0.10.0/packages/cli/test/cases/build.default).

thescientist13:
## Related Issue
resolves #446 

## Summary of Changes
1. Restore missing `<meta>` tags to default app template
1. Update default specs
1. Update website _app.html_

## Thoughts / Questions
I guess one thing this does present, is how would it look like for an application to edit more custom tags on a per page basis (like for a blog post's unique title, description, author, `og:*`, etc)?  

I know in React / Gatsby land they use [**React Helmet**](https://github.com/nfl/react-helmet) so from any component you can dynamically update the `<head>` after pulling data out of GraphQL.  I suppose this would justify #304 and something we can introduce as a standalone library?  

Otherwise, it's just as easy to recommend / suggest using basic `document` methods, but either way we should probably have a combination of one or more (or all) of these - example, doc, package.  I can update #304 with some comments and put it into our next project board.


thescientist13:
So one good idea coming out of the discussion on this for us would be to make a Favicons plugin, that can take a path to a _favicon.png_ or something like that, and spit out a bunch of favicon image formats, as well as adding all the needed `<meta>` tags for that.  This seems like a pretty useful feature to have for 1.0.0 and would pair really nicely with some boilerplate / stater kit repos like a blog or portfolio.

Issue #457, opened 2020-12-24
<link> tag detection is incorrectly assuming closing </link> tag (v0.10.0-alpha.1)


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
Our `<link>` tag detection rules seem a bit off.  For some reason I started out with thinking a `<link>` tag also has an accompanying closing `</link>` tag, but I don't see that in [any docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link).
```js
headLinks.forEach(link => {
  appTemplateContents = appTemplateContents.replace(/<\/link>/, `
    </link>\n
    <link ${link.rawAttrs}></link>\n
 `);
});
```

<img width="586" alt="Screen Shot 2020-12-24 at 1 14 45 PM" src="https://user-images.githubusercontent.com/895923/103102490-51161e80-45ea-11eb-9461-2bf632ca34e4.png">

## Details
Given the spec is something like
```html
<link href="/path/to/styles.css" rel="stylesheet">
```

We should update our code to correctly detect.  It seems puppeteer is fine with since the output seems to be what is expected
```sh
view-source:https://deploy-preview-436--elastic-blackwell-3aef44.netlify.app/
```
```html
<link rel="stylesheet" href="/styles/theme.bd54f1d4.css">
<link rel="stylesheet" href="/styles/home.752326c9.css">
```

but if user's aren't using `</link>` in their templates, then I think things will break for them.

Should make sure to update tests and docs as well.

thescientist13:
It would probably be nice to account for "short" variants in general, just be a little forgiving, e.g.
```html
<script type="module" src="..."></script> 
<script type="module" src="..."/>

<link rel="stylesheet" href="..."></link
<link rel="stylesheet" href="..."/> 
```

PR #475, opened 2021-02-13
Bug/issue 431 template <head> tags merge order

### LINKED ISSUE #431: assemble page and app <head> tags in the correct order
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Currently when a page is assembled into an app template in HTML transformation, script / style / link tags from the page are not appended _after_ all the app ones, instead they are appended after just the _first one_

currently its like this
```html
<head>
   <script app-1>
   <script page-1>
   <script page-2>
   <script app-2>
</head>
```

it should be like this
```html
<head>
   <script app-1>
   <script app-2>
   <script page-1>
   <script page-2>
</head>
```

## Details
Basically this code is too naive with the regex, maybe it would be better to use HtmlParser like in rollup?
```js
headScripts.forEach(script => {
  if (script.rawAttrs !== '') {
    appTemplateContents = appTemplateContents.replace(/<\/script>/, `
      </script>\n
      <script ${script.rawAttrs}></script>\n
    `);
  }

  if (script.rawAttrs === '') {
    appTemplateContents = appTemplateContents.replace(/<\/script>/, `
      </script>\n
      <script>
        ${script.text}
      </script>\n
    `);
  }
});

headLinks.forEach(link => {
  appTemplateContents = appTemplateContents.replace(/<\/link>/, `
    </link>\n
    <link ${link.rawAttrs}></link>\n
  `);
});
```

In general, this naive regexing throughout may want to be refactored.  But that could be another ticket.  Not sure which is faster (regex vs parser?

thescientist13:
## Related Issue
resolves #431 (and some other bugs I found along the way)

## Summary of Changes
1. Correctly merge page and app template heads tags
    - `<script type="module" src"..."></script>`
    - `<script> /* code */ </script>`
    - `<link rel="stylesheet" href="..."></link>` (~~note: `</link>` is incorrect but will fix in #457~~ )
1. Found / added _missing_ support for merging `<style> /* code */ <style>` tags
1. Updated `build.default.workspace-template-page` test case to cover all generate purpose page template testing
    - _found a bug where CSS path was mapping to `<script>` src during rollup bundling_
    - consolidated `build.default.workspace-template-page-styles` specs into this one
1.  Created a new case `build.default.workspace-template-page-and-app`
1.  Decided to pull in #457 since there are full tests here

## Thoughts / Observations
Potentially I had thought of adding short tag support for `<script>` but VSCode definitely does not like that, whether I do either or
```html
<script type="module" src="foo.js"/>
<script type="module" src="foo.js">
```
Screen Shot 2021-02-13 at 8 03 53 PM
Screen Shot 2021-02-13 at 8 04 09 PM

So I opted not to support it, but `<link>` works with both, so added specs for it and left it in


thescientist13:
So I think I could probably easily handle the same cases for `<script>` without too much effort.  I noticed in the MDN docs they only show the `<script ...></script>` variant, so with a simple regex we could probably just rewrite it on the fly, e.g.
```js
const someHtml = '...';

if (someHtml.match(/<script (.*)\/>/) {
 someHtml.replace('/>', '></script>') 
}
```

If I can pull this off in an hour or so, I'll get this in as part of this PR, otherwise I will create another issue for it.

thescientist13 (commented):
So in an interesting turn of events, I tried changing it to `<script src="..."/>` and the site actually started breaking, so it looks like the browser definitely _does_ seem to care. (no header)
Screen Shot 2021-02-24 at 9 02 04 PM

Screen Shot 2021-02-24 at 9 01 53 PM

Even the view source syntax highlighting breaks
Screen Shot 2021-02-24 at 8 56 46 PM

So I guess in this case, closing tag "forgiveness" would actually be a bad idea.  There are some good points made for this as well in this SO post.
https://stackoverflow.com/questions/69913/why-dont-self-closing-script-elements-work

So, going to leave this as then and promote `</script>` tag usage.

PR #481, opened 2021-02-22
Enhancement/issue 426 plugin babel

### LINKED ISSUE #426: restore Babel / PostCSS / Browserslist for no bundle development

thescientist13:
## Related Issue
resolves #426 (finally!)

## Summary of Changes
1. Added a new plugin for supporting Babel (using private fields / members as an example)
1. Added initial support for `exportMap` handling in **plugin-node-modules**
1. Established `peerDependency` for **plugin-postcss** and **plugin-babel**
1. Restored support for browserslist for Babel and PostCSS
1. Dropped IE11 from _.browserslistrc_
1. Restored `eject` functionality

Demo code using `#private` members running via branch preview
Screen Shot 2021-03-14 at 6 16 14 PM

## TODO
1. [x] Tests
1. [x] improve support for / remove commonjs dependency from Babel
1. [x] Add support for preset-env / browserslist for both PostCSS and ~~Babel~~ 
1. [x] Remove test code / _www/package.json_ / _.eslintignore_ override

## Questions / Thoughts
Do we still need an `eject` feature at this point as now all but our (equivalent) **webpack** configuration is gone from the project and is being moved to plugins, and I would honestly prefer not to couple so tightly to any particular technology (e.g. Rollup).  I think the Rollup plugin API is actually a leaky abstraction but I recognize can be a useful escape hatch.  However, if I can find a way to reuse something like `ResourceInterface.intercept` during build time instead, I might consider deprecating the Rollup plugin API, or at least dropping support for `eject`.


greenwood.config.js
```diff
@@ -23,6 +25,10 @@ module.exports = {
     { name: 'google-site-verification', content: '4rYd8k5aFD0jDnN0CCFgUXNe4eakLP4NnA18mNnK5P0' }
   ],
   plugins: [
+    ...pluginBabel({
+      mergeConfigs: true
+    }),
+    ...pluginCommonJs(),
```
thescientist13, line 31: Will try and remove the need for this this through better `exportMap` handling

Issue #418, opened 2020-09-22
Roadmap To 1.0 And Beyond


thescientist13:
This issue is intended to outline a the short and long term planning of the project for review and consensus gathering.  It is broken down into the following milestones as part of the discovery done as part of the [NoNo POC](https://github.com/thescientist13/nono-poc) and the work to port that to Greenwood here in #417.

This will also serve as a way to update current GitHub Issues, Projects, Milestones

## Questions
- [`<meta>` component / package](https://github.com/ProjectEvergreen/greenwood/issues/304) still needed? - yes!  have some **react-helmet** like features
- [RFC Declarative App templates](https://github.com/ProjectEvergreen/greenwood/issues/299) still needed? - no, not really I don't think

## Feature Parity (relatively)  by 12/31/2020
This is work needed to accomplish the bulk of the work necessary to get #355 merged without much disruption to existing users.  Merged into an ongoing `release/1.0.0` release branch and tracked in the Architecture Digest Project / MVP Milestone.

Will want to validate these using an `rc` or `next` branch release before merging into `master`
* Greenwood Getting Started - https://github.com/ProjectEvergreen/greenwood-getting-started/pull/22
* www.thegreenhouse.io - https://github.com/thegreenhouseio/www.thegreenhouse.io/pull/182
* projectevergreen.github.io - https://github.com/ProjectEvergreen/projectevergreen.github.io/pull/36

Main thing here should be ensure that performance improves significantly and a number of existing issues in the backlog can be resolved.  Each of the below numbered items would be a PR and tracked in the release branch PR.

1. ~~PR and associated TODOs~~ from #417 
1. ~~Refactor _serve.js_ into plugin based architecture~~ bring done as part of #185 
1. ~~SPA support~~ - delegated to #354 
    - (how to) handle strict / ssg `optimization` in _greenwood.config.js_
    - Decouple spa specific code from the bundle
    - [Smaller / lighter router](https://router.matsu.fi/getting-started)?
    - [rollup dynamic input vars](https://github.com/rollup/plugins/tree/master/packages/dynamic-import-vars)
    - puppeteer optional?  e.g. SPA no SSG, SSG, vs SPA vs Hybrid (strict?)
1. ~~[GraphQL as a package](https://github.com/ProjectEvergreen/greenwood/issues/278)~~, or [something simpler](https://jaredpalmer.com/gatsby-vs-nextjs)
    - update all usages to use `@data/client` instead of `fetch`
[...]

Issue #347, opened 2020-05-07
reduce duplicate fetches for cache.json by storing in client.js memory


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As a minor optimization for [_client.js_](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/data/client.js#L23), it would be good to cache in memory previously `fetch`ed calls to cache.json.

## Details
In this way, the browser won't have to make redundant HTTP requests to something it already has access to.

hutchgrant:
You cannot cache them in client.js because it's not a singleton. A new instance is created every single route change, for each component that imports it.

hutchgrant:
We can however reuse the `<script data-state="apollo"></script>` to hold our `window.__APOLLO_STATE__` data.  On route change currently, that data remains static from when the page was initially loaded(even that isn't being utlized at all).
Modify client.js like so:

```js
client.query = (params) => {

  if (APOLLO_STATE && Object.keys(APOLLO_STATE).length === 0) {
```
and it will use the apollo_state on initial page load. The problem is when the route changes, with this condition it will try to reach graphql server.

thescientist13:
Basically if we provide client side routing of any kind (SPA / hybrid model), then we should expect to have to reload the whole page (app shell) if we want things to work as would expect in SSG mode (which would be doing a full page reload)
- `<meta>`
- `<title>`
- data

Part of paying the cost to enable going beyond just progressive enhancement.  But makes things simpler for at least achieve SSG mode.  :)

thescientist13:
Moving to deferred for this project to be added to our next project, but it looks we can take advantage of the work being deon for adding an MPA setting in #477 to implement what we need for this feature.  

More info here - https://github.com/ProjectEvergreen/greenwood/pull/487#issuecomment-798941402

thescientist13:
not sure where i stand on this one, given that I am considering deprecating the GraphQL plugin - https://github.com/ProjectEvergreen/greenwood/issues/1390

PR #487, opened 2021-03-08
Enhancement/issue 278 graphql plugin

### LINKED ISSUE #278: GraphQL (Apollo) data as its own package

thescientist13:
## Related Issue
resolves #278 

## Summary of Changes
1. Created **graphql** package, and moved _data/_ folder from **cli** package
1. Upgraded `<app-header>` and `<app-shelf>` components to use `client`
1. Upgraded to Apollo 3 *
1. Found a bug with **rollup** `load` handling for serving custom resources
1. Restored test cases / unit test, moved under **plugin-graphql** and improved to cover all query cases
 
graphiql

## TODOs
1. [x] Enable Unit Tests / Specs
1. [x] Documentation
1. [x] Review status of #347 and #349
1. [x] `@apollo/client` support in _client.js_ (see thoughts below) - added as a [Roadmap to 1.0 Tracking item](https://github.com/ProjectEvergreen/greenwood/issues/418)
1. [x] clean up example code / TODOs / commented out code
1. [x] Track open items

### Apollo Client
There were a number of issues / hacks / workarounds I bumped into while trying to use `@apollo/client` in _client.js_ so I have put that work into [a branch](https://github.com/ProjectEvergreen/greenwood/compare/enhancement/issue-278-graphql-plugin...enhancement/issue-278-graphql-plugin-apollo-client) for now, but will still try and kick the tires for a bit longer
 - Due to [one of `@apollo/client`'s dependencies](https://github.com/apollographql/apollo-client/issues/5961#issuecomment-775828986) still being CommonJS [**(fast-json-stable-stringify)**](https://github.com/epoberezkin/fast-json-stable-stringify/pull/20), including our **plugin-import-commonjs** was required
- Although ESM, _graphql/language/parser.mjs_ doesn't provide a default `export`, which I had to handle using `intercept`
- There is an odd quirk with **ts-invariant** dependency [needing to be resolved within `@apollo/client` package's _node_modules_](https://github.com/apollographql/apollo-client/issues/7406#issuecomment-741621089)
- Had to stop using [`getQueryHash`](https://github.com/ProjectEvergreen/greenwood/blob/v0.9.0/packages/cli/src/data/common.js#L35) for the time being since Apollo isn't available on the client now

no-default-export-console
no-default-export-source
ts-invariant-shim

----

Ultimately landed on this bug, same thing happened in `apollo-client`. 😞 

observable-constructor-new

1. need to ensure a clear distinction between when file extensions are / aren't needed when using `import` as the spec says they should always be there but for now, will make it so _node_modules_ can be bare, and can make an issue to add support for bare module specifier as a plugin
1. should create implement shared _package.json_ dependency walker for `importMaps` like in [**plugin-babel** / `regenerator-runtime`](https://github.com/ProjectEvergreen/greenwood/pull/481/files#diff-091732f9dd50157f76708e4269d733499ab1d22c84e56f3366a605c196366b9fR48)
1. Currently need to have distinction between common and server client's until we can get to node v14 where we can use `import` on NodeJS code
1. Should really implement [`server.stop` handling on `process` end](https://stackoverflow.com/questions/10522532/stop-node-js-program-from-command-line)
1. try and avoid need for `createCache` round-robin 🐦


thescientist13:
Took a little time to look into #347 and #349 to see if there were any quick wins but there as the new baseline is specifically targeted towards SSG, there would at least need to be an MPA (or SPA) option in place in order to be able to keep existing data in memory, since as of right now, every page is a hard load.  Will move those two issues to deferred and will plan to schedule them for the next Project.

thescientist13:
Was hoping closing and reopening would trigger a Netlify build a la #489 but didn't seem to happen.  Maybe because no change in git hash?

PR #495, opened 2021-03-27
[demo / bug] inline optimization is broken on subsequent page load

thescientist13:
## Related Issue
related to #447, this is an example of the `inline` **optimization** setting.  Click [here for the simpler diff](https://github.com/ProjectEvergreen/greenwood/compare/enhancement/issue-354-restore-optimization-for-no-bundle...enhancement/issue-354-restore-optimization-for-no-bundle-inline).  _**This is a known broken issue since it works in the specs, but not with the Greenwood**_

Score seems about the same?

_Network requests (not entirely everything is inlined, but most)_
optimization-inline-network

_Lighthouse_
optimization-inline-lighthouse

_On Navigation_ 😬 
optimization-inline-route-change

## Summary of Changes
1. Sets `optimization` setting to **inline** in _greenwood.config.js_
1. Set `<base href="/">` to set consistent reference point for all JS / CSS file requests

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -11,6 +11,7 @@ const FAVICON_HREF = '/assets/favicon.ico';
 module.exports = {
   workspace: path.join(__dirname, 'www'),
   mode: 'mpa',
+  optimization: 'inline',
   title: 'Greenwood',
   meta: [
     { name: 'description', content: META_DESCRIPTION },
--- a/www/templates/app.html
+++ b/www/templates/app.html
@@ -2,6 +2,7 @@
 <html lang="en" prefix="og:http://ogp.me/ns#">
 
   <head>
+    <base href="/">
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
     <meta name="mobile-web-app-capable" content="yes"/>


thescientist13:
looking at this some more now after syncing with master, getting new errors related to double component definition errors?
Screen Shot 2021-04-28 at 9 04 08 PM

thescientist13:
Ok, so following the first error at least, it looks like something is happening for pages just using the shelf and attributable to [this section of the shelf code](https://github.com/ProjectEvergreen/greenwood/blob/v0.13.0/www/components/shelf/shelf.js#L41)

Screen Shot 2021-07-02 at 1 10 36 PM

```js
  expandRoute(path) {
    let routeShelfListIndex = this.shelfList.findIndex(item => {
      let expRoute = new RegExp(`^${path}$`);
      return expRoute.test(item.route);
    });
```

As can be seen, there's a random HTML starting tag right in the middle of that regex in the minified and inlined code? 😳 
```html
expandRoute(e){let t=this.shelfList.findIndex(t=>new RegExp(`^${e}<!DOCTYPE html><html lang="en" prefix="og:http://ogp.me/ns#">
```

thescientist13:
The plot thickens... checked a couple more things

- When checking the _.greenwood/_ directory, the HTML files are fine, so it seems to indicate the issue happens after prerendering (Puppeteer), which really only leaves the bundle step (Rollup)
- If I check the source of the bundled _shelf.js_ in the _public/_ directory (as bundled by Rollup) the Regex is fine!

Screen Shot 2021-07-07 at 1 35 43 PM

So, basically it must be isolated to what is happening in our _rollup.config.js_, but specifically in the inlining logic.  Let me setup a test environment know that I've been able to narrow things down.

I probably botched a regex or `replace` call somewhere...

----

On an unrelated note, I'm not sure we actually need the `^` and the `$` since we're not using multiline, but I'll revisit that a bit later.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#advanced_searching_with_flags
> _The m flag is used to specify that a multiline input string should be treated as multiple lines. If the `m` flag is used, `^` and `$` match at the start or end of any line within the input string instead of the start or end of the entire string._

thescientist13:
Dang, so I think this is the issue
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter
>  $` | Inserts the portion of the string that precedes the matched substring.


Essentially, the Regex inside our source code is getting interpreted by `replace` as a parameter it should do something with. :/

thescientist13:
Ok. moved [to an issue](https://github.com/ProjectEvergreen/greenwood/issues/656) now that I can reproduce in a spec locally.

Issue #21, opened 2019-03-27
[RFC] External Data (Graph) Sources


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Need to figure out a way to pull in external data at build time and pull it in a build time.   First POC should likely be with Wordpress.  (Greenwood blog?)

## Details
I think it should be pretty straight forward, expose a "Graph" Plugin type that gets read on build to extract the metadata and then just serves the route as any other would?

From there, other plugins could be built around that in follow up PRs, like for WordPress or Sanity support.

thescientist13:
I think if we were to get a working example, we could officially consider our project JAMStack friendly, and should probably add that to the README / website home page.

thescientist13:
For incoming content / data, not sure if we would want to account for some sort of DOM validation / purification?  ex.
https://github.com/cure53/DOMPurify

Or at least make sure user's can apply it themselves through our plugins.

PR #509, opened 2021-04-04
Enhancement/graphql custom resolvers and typedefs

### LINKED ISSUE #508: support ad-hoc / custom GraphQL resolvers and definitions (schemas)
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
In some discussions with @hutchgrant and exploring use cases around #21 , there is a case to be made that you could use GraphQL for not specifically tied to the graph specifically.  For external content sure, that would need to go into the graph, but anything not content, like the config, might want to establish its own top level schema.  But maybe there could be some overlap?

For example, for image galleries (as was the use case being discussed) your data source there would just be a set of images, either from a directory or cloud host.

## Details
So basically this would require a convention / API for **plugin-graphql** to be able to scan for schemas in a repo and merge those into the overall schema.  Similar to NextJS or Netlify, something like
```js
// src/data/schema.js in a user's workspace
const gql = require('graphql-tag');

const getGallery = async (root, query, context) => {
  return [{
    name: 'Gallery 1',
    title: 'Gallery Title',
    images: [{
      path: '/assets/gallery1/image1.png'
    }, {
      path: '/assets/gallery1/image2.png'
    }]
  }];
};

const galleryTypeDefs = gql`
  type Image {
    path: String
  }

  type Gallery {
    name: String,
    title: String,
    images: [Image]
  }

  extend type Query {
    gallery: [Gallery]
  }
`;

const galleryResolvers = {
  Query: {
    gallery: getGallery
  }
};

module.exports = {
  customTypeDefs: galleryTypeDefs,
  customResolvers: galleryResolvers
};
```

You could then query it with your own _.gql_
```js
query {
  gallery {
    name,
    title,
    images {
      path
    }
  }
}
```

thescientist13:
## Related Issue
resolves #508 

## Summary of Changes
1. Added the ability for users to provide a schema (defs / resolvers) in a directory called _{workspace}/data/schema/_

Screen Shot 2021-04-04 at 3 54 18 PM
Screen Shot 2021-04-04 at 4 23 33 PM

> _Found a bug in #573  that will prevent the formal demo here from working with `build` until it is merged since I am using a second inline `<script>` but the this PR does work.  You should be able to validate by merging that branch into here._

## TODO
1. [x] Flesh out example in _www/_ to actually pull images from a directory in the resolver (based on a variable)
    - multiple galleries (query should take a variable) sourced from frontmatter
    - demonstrate a gallery somewhere
1. [x] Allow configuration for _data/schema_ locations - captured in #570
1. [x] How can _plugins_ define or provide this to **plugin-graphql** (similar naming convention as default?) - #570
1. [ ] ~~Maybe for #21 , it could be more about how to get things into the _graph_ specifically?~~  - updated
1. [x] support an assets graph - #571
1. [x] Test cases
1. [x] Documentation
1. [x] Revert example code


packages/plugin-graphql/src/schema/schema.js
```diff
@@ -1,50 +1,78 @@
...
+const fs = require('fs');
 const gql = require('graphql-tag');
+const path = require('path');
 
-module.exports = (graph) => {
-  let uniqueCustomDataDefKeys = {};
-  let customDataDefs = '';
+module.exports = (compilation) => {
+  const { graph } = compilation;
+  const uniqueCustomDataDefKeys = {};
```
thescientist13, line 10: Naming / organization could maybe some thought here, just so it's clearer what everything and where it is coming from.

packages/plugin-graphql/src/schema/schema.js
```diff
@@ -1,50 +1,78 @@
...
+  const customSchemasPath = `${compilation.context.userWorkspace}/data/schema`;
```
hutchgrant, line 11: So this is great for unique queries for each individual user.  If you want to extend it to have plugins include schema, we would need to do an additional directory walk of any plugins containing a graphql flag of somekind.

hutchgrant (commented):
This is good for individual users wanting to BYOQ(bring your own query) because now  they can write custom resolvers and utilize the graph in many different ways.  The gallery, as an example, is fine. But would like to confine the gallery typedefs + resolver + query + wrapper component to its own package eventually.

PR #515, opened 2021-04-10
differentiate client side routes from (external) URLs

### LINKED ISSUE #500: links not clickable on the website using `mpa` optimization in Firefox / Safari
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
In Firefox or Safari, click any links on the site, nothing happens.

## Details
Looks like an error in the router, when anything is clicked
router-broken

thescientist13:
## Related Issue
resolves #500 

## Summary of Changes
1. Better distinction between routes (_/about/_, _/docs/_) and URLs, (http://www.polymer-project.org) such that URLs are not treated the same as app routes
1. Turned `mpa` setting back on
1. Tested a bunch of links, deep links, and external links and everything is working in Safari*, FF, Edge, Chrome

### Safari / Event Object Detection
So far this still "doesn't work" for Safari, in the sense that _href_ will always be `null` and so it will always do a hard reload, even for a route.  However, I don't think that's Safari's fault, per se.

Given the current code
```js
const href = e.path && e.path[0]
  ? e.path[0].href // chrome
  : e.originalTarget && e.originalTarget.href
    ? e.originalTarget.href // firefox
    : null;
```

None of that is web standard and so won't work in Safari 😨 ; [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event) and [`MouseEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent).  The only consistent one I can find is `target` and `currentTarget` but I don't the `<a>` tag for either of those.  🤔 
Screen Shot 2021-04-10 at 11 25 08 AM

Event tried do something like this, but the event never even fires for the `<a>` link click.  😞 
```js
document.addEventListener('DOMContentLoaded', () => {
  Array.from(document.getElementsByTagName('a')).forEach((link) => {
    console.log('bootstrap link', link);

    link.addEventListener('click', function(e) {
      // but nothing happens though!?
      console.debug('@@@@@@@ e', e);
    });

  });
});
```

Will make an issue to take a much deeper look into better cross browser support for this.  Maybe a good case to get some E2E Testing setup for this project?  I can make a discussion for that if anyone is interested.

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -10,7 +10,7 @@ const FAVICON_HREF = '/assets/favicon.ico';
 
 module.exports = {
   workspace: path.join(__dirname, 'www'),
-  // mode: 'mpa',
+  mode: 'mpa',
   title: 'Greenwood',
   meta: [
     { name: 'description', content: META_DESCRIPTION },
--- a/packages/cli/src/lib/router.js
+++ b/packages/cli/src/lib/router.js
@@ -7,11 +7,9 @@ document.addEventListener('click', function(e) {
     : e.originalTarget && e.originalTarget.href
       ? e.originalTarget.href // firefox
       : '';
+  // we only want to handle links like /about/ and /docs/ to trigger client side routing
   const isUrl = href && href.match(urlRegex);
-  // we only want routes like /about/, /docs/ to trigger client side routing
-  // and also want to exclude # links
   const canClientSideRoute = href && href !== '' && !isUrl;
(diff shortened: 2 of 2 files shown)


thescientist13:
Made a follow up issue and discussion to keep the conversation going.
https://github.com/ProjectEvergreen/greenwood/issues/559
https://github.com/ProjectEvergreen/greenwood/discussions/560

Issue #559, opened 2021-04-16
make router standards compliant with consistent cross browser support


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of working #500 / #515, realized that the router is a bit of hack at the moment, without proper Safari support at the moment.  😞

Need to make this a bit more proper and use standard's compliant event handling.
 

## Details
Given the current code
```js
const href = e.path && e.path[0]
  ? e.path[0].href // chrome
  : e.originalTarget && e.originalTarget.href
    ? e.originalTarget.href // firefox
    : null;
```

None of that is web standard and so won't work in Safari 😨 ; [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event) and [`MouseEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent).  The only consistent one I can find is `target` and `currentTarget` but I don't the `<a>` tag for either of those.  🤔 
![Screen Shot 2021-04-10 at 11 25 08 AM](https://user-images.githubusercontent.com/895923/114275183-89056e00-99ef-11eb-8a19-49fd3fb8f61a.png)

Event tried do something like this, but the event never even fires for the `<a>` link click.  😞 
```js
document.addEventListener('DOMContentLoaded', () => {
  Array.from(document.getElementsByTagName('a')).forEach((link) => {
    console.log('bootstrap link', link);

    link.addEventListener('click', function(e) {
      // but nothing happens though!?
      console.debug('@@@@@@@ e', e);
    });

  });
});
```

----

Will this allow us to unmark the [feature as experimental](https://www.greenwoodjs.io/docs/configuration/#static-router)?

thescientist13:
Great, well after merging #515, this isn't working on Netlify, just locally.  😡

thescientist13:
Could also consider investigating in an "MPA" style router, as a plugin - https://swup.js.org/

thescientist13:
Looks we found a solution in #1069 🤞 👀

PR #563, opened 2021-04-16
better handling of external links for MPA router

### LINKED ISSUE #562: MPA router treating all links as external links
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
MPA router, even when enabled now and "fixed" as part of #500 , is now not doing _any_ client side routing.  All links just do a full page reload.  :/

> _Just to preface that hopefully #559 will address this cleanly once and for all by _hopefully_ being able to target the `<a>` tag directly, combined with some investment in #560._

## Details
Looks like with the new [URL detection logic](https://github.com/ProjectEvergreen/greenwood/commit/bd49529b135b22de243456dff43dae3e8d8f8737#diff-e4b4bc56ca553ca4d6160f0710d5d2a6d75a1af1a3ee2235bff9cc8ea7931afaR10), while it was working locally on `localhost`, I would see something like _/docs/_,  In production, you'll see it actually returns as _https://www.greenwoodjs.io/docs_ which will of course pass the URL check.  Locally, if I edit my _etc/hosts_ and visit `local.greenwoodjs.io:8080` when running `yarn serve` I can reproduce the same issue as in production.

Basically, the best way I can figure out how to do this at the time, to distinguish between a "route" and an off site URL, is to see if the current link includes the current domain in it.  If it does, assume it is a route, otherwise consider it an external link.  🤷‍♂️ 

---

As you can tell this is not ideal, ideally for any given links, e.g.
```html
<a href="/docs/" ....>Greenwood Docs</a>

<a href="http://lit-element.polymer=project.org" ...>Lit Docs</a>
```

I would want to examine the _exact_ value of the href, not whatever the browser fully qualifies.

thescientist13:
## Related Issue
resolves #562 

> _Ultimately if we can address #559 , then we won't have to best guess at all, as ideally we can get access to explicit `href` attribute of the `<a>` tag itself, not with the browser fully qualifying the URL for us._

## Summary of Changes
1. Better "guess" handling of links on the site vs links off the site.

### DIFF
--- a/packages/cli/src/lib/router.js
+++ b/packages/cli/src/lib/router.js
@@ -5,8 +5,9 @@ document.addEventListener('click', function(e) {
     : e.originalTarget && e.originalTarget.href
       ? e.originalTarget.href // firefox
       : '') || '';
-  // best case guess is that if the link oriniates on the current site
-  // treat it as a client side route, ex:  /about/, /docs/
+  // best case "guess" is that if the link originates on the current site when resolved by the browser
+  // treat it as a client side route, ex:  /about/, /docs/ and trigger the client side router
+  // https://github.com/ProjectEvergreen/greenwood/issues/562
   const isOnCurrentDomain = href.indexOf(window.location.hostname) >= 0 || href.indexOf('localhost') >= 0;
   const canClientSideRoute = href !== '' && isOnCurrentDomain;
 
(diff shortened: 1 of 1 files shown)


thescientist13:
Hmm, getting this unrelated build failure from Netlify
```sh
4:24:29 PM: Started restoring cached go cache
4:24:29 PM: Finished restoring cached go cache
4:24:29 PM: Installing Go version 1.12
4:24:29 PM: I don't have any idea what to do with '1.12'.
  (using download type 'binary')
4:24:29 PM: Failed to install Go version '1.12'
```

thescientist13:
Phew, router is back to working for the most part.  😌

PR #573, opened 2021-04-25
fix hashing and marker detection and make specs more explicit

### LINKED ISSUE #572: multiple inline `<script>` tags not getting correctly optimized - all tag tags show the same contents
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Looks like there is a bug in our _rollup.config.js_ logic that tries to optimize inline `<script>` tags.  For example this
```html
<script type="module">
const one = 'script tag module inline one';

document.getElementsByClassName('output-script-inline-one')[0].innerHTML = one;
</script>

<script type="module">
const two = 'script tag module inline two';

document.getElementsByClassName('output-script-inline-two')[0].innerHTML = two;
</script>
```

should come out like this
```html
<script type="module">
document.getElementsByClassName('output-script-inline-one')[0].innerHTML = 'script tag module inline one';
</script>

<script type="module">
document.getElementsByClassName('output-script-inline-two')[0].innerHTML = 'script tag module inline two';
</script>
```

instead it is coming out with the second inlined version being the same code as the first inline `<script>` tag 
```html
<script type="module">
document.getElementsByClassName('output-script-inline-one')[0].innerHTML = 'script tag module inline one';
</script>

<script type="module">
document.getElementsByClassName('output-script-inline-one')[0].innerHTML = 'script tag module inline one';
</script>
```

## Details
Not sure if the issue is related to poor hashing, so maybe a good case to finally tackle #553?  While we're at, should confirm how this works
1. With / without `type="module"`
2. `<style>` tags

thescientist13:
## Related Issue
resolves #572 

## Summary of Changes
Ooof, a couple things going on here that we not properly implemented (my bad)

1. Hashing was not implemented correctly and not really generating unique IDs, so copied over hashing function from graphql (for now)
1. Fixed marker detection logic

## TODOs
1. [x] Check for other inline cases of plain `<script>` and `<style>`, add specs as needed
1. [x] `eve-container` module [not being bundled now](https://github.com/ProjectEvergreen/greenwood/pull/573#issuecomment-826359284)
1. [x] Should make an issue to make sure `<style>/* code */` and `<script>/* code */</style>` are getting optimized / minified - https://github.com/ProjectEvergreen/greenwood/issues/581
1. [ ] I think this might be the root cause of #495, should see if it helps after this is merged

### DIFF
--- a/packages/cli/src/config/rollup.config.js
+++ b/packages/cli/src/config/rollup.config.js
@@ -194,7 +194,7 @@ function greenwoodHtmlPlugin(compilation) {
             // handle <script type="module">/* some inline JavaScript code */</script>
             if (parsedAttributes.type === 'module' && scriptTag.rawText !== '') {
               const id = hashString(scriptTag.rawText);
-              
+
               if (!mappedScripts.get(id)) {
                 const marker = `${id}-${tokenSuffix}`;
                 const filename = `${marker}.js`;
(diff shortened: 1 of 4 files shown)


thescientist13:
Hmm, now i'm seeing this error again :/
```
(index):1 Uncaught TypeError: Failed to resolve module specifier "@evergreen-wc/eve-container". Relative references must start with either "/", "./", or "../".
```

From looking at [the generated source](view-source:https://deploy-preview-573--elastic-blackwell-3aef44.netlify.app/docs/), looks like it is not being bundled correctly
```html
<script type="module">
  import '@evergreen-wc/eve-container';
</script>
```

PR #640, opened 2021-06-20
add getting started repo stackblitz to quick start section of getting started docs page

thescientist13:
## Related Issue
related to discussion https://github.com/ProjectEvergreen/greenwood/discussions/639

## Summary of Changes
1. Adds **greenwood-getting-started** as Stackblitz to ~~home~~ Getting Started page quick start section
1. Added error handling for puppeteer initialization so it doesn't just blow like it does in Stackblitz

Screen Shot 2021-06-30 at 12 39 34 PM
Screen Shot 2021-06-30 at 12 40 34 PM

## Questions
1. [x] Stackblitz doesn't `postinstall` scripts, so `yarn serve` with Puppeteer won't "work".  I think even so, having something like this from the home page is still worth having, as it's super interactive and demonstrable?  (see Known Issues in the linked discussion).  Should we at least `try / catch` the exception just so it looks a bit "cleaner" in the terminal?
1. [x] Is the home page a good place for this, or should this be moved to another location?  VS Code in an iframe seems like a big weight to add to the home page, so perhaps _Getting Started -> Quick Start_ is more appropriate?
1. [x] Should this only be enabled for non mobile?  Seems like it wouldn't be super helpful / accessible on mobile?
1. [x] Thoughts on dimensions / size / styling?

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -1,5 +1,4 @@
 const path = require('path');
-const pluginGoogleAnalytics = require('@greenwood/plugin-google-analytics');
 const pluginGraphQL = require('@greenwood/plugin-graphql');
 const pluginImportCss = require('@greenwood/plugin-import-css');
 const pluginImportJson = require('@greenwood/plugin-import-json');
@@ -27,9 +26,6 @@ module.exports = {
     { name: 'google-site-verification', content: '4rYd8k5aFD0jDnN0CCFgUXNe4eakLP4NnA18mNnK5P0' }
   ],
   plugins: [
-    pluginGoogleAnalytics({
-      analyticsId: 'UA-147204327-1'
-    }),
     ...pluginGraphQL(),
     pluginPolyfills(),
     pluginPostCss(),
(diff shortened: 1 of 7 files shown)


thescientist13:
From the meeting today, let's try this out on the Quick Start tab of the Getting Started page.  There we can put a little disclaimer that some features / operations may be limited, but otherwise it all works pretty well and is definitely a nice show piece!  🎨

thescientist13:
@aholtzman 
What are your thoughts on this for mobile?  I think our code blocks might not already be mobile friendly, but obviously with a VSCode window, it's definitely going to be pretty "compromised".
Screen Shot 2021-06-23 at 9 31 54 PM

Should we just hide it on mobile and just leave a link to it?

thescientist13:
Per our chat, we should hide this on mobile and just link to it instead.

thescientist13:
Improved the mobile experience by linking out to the Stackblitz and setting the breaking point at `768px` which when rotated, still seems to provide a decent experience.  Made it `display: none` just to make sure it only shows when absolutely needed.

Screen Shot 2021-06-30 at 12 25 16 PM
Screen Shot 2021-06-30 at 12 25 40 PM


This brings up a good point, we may want to add a feature that allows expanding / collapsing of the side bar.

aholtzman (approved):

PR #642, opened 2021-07-01
enable sourcemaps with Rollup

### LINKED ISSUE #319: output sourcemaps for production builds
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Came up as part of build time data access, #269.  Just checked, and it doesn't look like sourcemaps are being output?  (as part of a prod build anyway)
Screen Shot 2020-04-11 at 5 35 26 PM

## Details
Should make sure it is enabled in **webpack** and also make sure to install / use [source-map-explorer](https://www.npmjs.com/package/source-map-explorer)

Also should check it works in develop mode as well.

thescientist13:
## Related Issue
resolves #319 

## Summary of Changes
1. Enabled [`output.sourcemap` option](https://rollupjs.org/guide/en/#outputsourcemap) for Rollup

## TODOs
1. [x] Update / add specs accordingly
1. [x] Need to do some file cleanup as both HTML files (since they are the `input`s to Rollup) ~~and any "scratch" files emitted as part of inlining~~ should get cleanup too
Screen Shot 2021-07-01 at 12 48 33 PM
1. [x] Track full inlining - https://github.com/ProjectEvergreen/greenwood/issues/652


thescientist13:
Just realizing this now, it looks like `inline` optimization may not be acting "aggressively" enough.  For example, in one of our specs, when we `import` from _node_modules_, that code itself is not inlined.
```html
import"/lit-html.d69ea26f.js";import"/lit-element.ea26fec7.js";//# sourceMappingURL=1832243663-scratch.01361a49.js.map
```

To be truly inline so as not to initiate ANY network requests, we would want all of that _node_modules_ code inlined too.

Issue #410, opened 2020-08-06
explore strict mode compatibility for Greenwood website


thescientist13:
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Related to the work done in #396 as part of #354 with the introduction of the `optimization` configuration option , would like to explore the feasibility of making the Greenwood website work using **strict**.

## Details
The main consideration things that I can think of that would need to be addressed
1. The [banner rotation](https://github.com/ProjectEvergreen/greenwood/blob/v0.8.0/www/components/banner/banner.js#L32) rotation on the home page would need to be moved into inline `<script>` tags
1. The shelf sub nav would need to work without [LitElement `@click`](https://github.com/ProjectEvergreen/greenwood/blob/v0.8.0/www/components/shelf/shelf.js#L115) event handler
1. There are usages of custom elements being uses as CSS selectors which will not work

Maybe convert it to Lit+SSR with hydration for the progressive hydration for just the navigation?

---

> That said, the above all make great criteria for proving the idea of Progressive Enhancement in that we only pass through userland code _without_ the SPA bundling, but I'm also curious just to see what it would take, even if the PR doesn't get merged for strict mode in the Greenwood website.  I guess it depends on what comes first.  😄

thescientist13:
We're doing an overhaul of the website as part of #978 so can close this out.

PR #645, opened 2021-07-02
Enhancement/issue 546 optimization overrides

### LINKED ISSUE #546: enable support for all optimization settings and overrides
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Currently not all optimization settings are enabled yet and are marked as [experimental in the documentation](https://www.greenwoodjs.io/docs/configuration/#optimization).

In particular the `inline` and `static` options are not enabled, plus we need to support adding these at the tag level for even more granular overrides.

thescientist13:
## Related Issue
resolves #546 

## Summary of Changes
1. Add `static` and `inline` optimization overrides for `<script>` tags
1. Add `inline` optimization overrides for `<link>` tags
1. Add tests
1. Updated documentation, `none` override will have to [come later]()
1. Add examples to website to demonstrate #410 

## TODOs
1. [x] Document considerations for [ShadowDOM based styles](https://github.com/ProjectEvergreen/greenwood/commit/9e5df1af6171fbf9056a58a7395a56bfd9c8bda6#diff-00ab471ecd2a7bae33689ccdbb630c7aec570e6a95ef47f69606fa4251a89bb8R11) - [**done**](https://github.com/ProjectEvergreen/greenwood/pull/645#issuecomment-873125192)
1. [x] ~~clean up / remove override attributes from final HTML~~ none of the options we support keep the original tag, hence the attributes are not present, so nothing to do here really
1. [x] clean up inlined assets from public output
1. [x] Clean up `TODO`s
1. [x] Before / after benchmarks for website - Screen Shot 2021-07-07 at 12 39 46 PM
    - keep / remove overrides from _www/_? (see #410 for additional context)


thescientist13:
As far as this commit where [`!important` was added](https://github.com/ProjectEvergreen/greenwood/commit/9e5df1af6171fbf9056a58a7395a56bfd9c8bda6#diff-00ab471ecd2a7bae33689ccdbb630c7aec570e6a95ef47f69606fa4251a89bb8R11) for the footer, the footer looked like this for the production build.
Screen Shot 2021-07-02 at 12 39 31 PM

It should look like this.
Screen Shot 2021-07-02 at 12 38 39 PM
----

Using dev tools, what I could see is that global styles where inheriting into this element. which is what was causing the undesired padding.

Screen Shot 2021-07-02 at 11 55 52 AM

Which makes sense, since the footer.css uses / assumes ShadowDOM and usage of `:host`, in this case where the JavaScript is not shipped to the user due to using `data-gwd-opt="static"`, all CSS was now global and inlined by puppeteer.

Screen Shot 2021-07-02 at 11 49 07 AM

```html
app-footer .footer.app-footer {
  background-color: #192a27;     min-height: 30px;     padding-top: 10px;
}

app-footer .footer.app-footer h4.app-footer {
  width: 90%;       margin: 0 auto;       padding: 0;       text-align: center;
}

app-footer .footer.app-footer a.app-footer {
  color: white;       text-decoration: none;
}
```

This means that no encapsulation can be guaranteed, so in some cases, additional consideration / tweaks may be needed, like using `!important` or using some other technique for managing specificity / the cascade when not relying on JavaScript at runtime.

thescientist13:
Just a heads up here re: #652

thescientist13:
Did a before / after comparison.  Lighthouse scores are about the same, but we can definitely shave off a few network requests and bring down our payload size by `12KB` and 5 requests, so it's something at least.  That said, putting this branch on production Netlify could potentially bump this into a 100 though, I think that tends to happen so... 🤞 

## Production
Screen Shot 2021-07-07 at 12 39 46 PM

Screen Shot 2021-07-07 at 12 40 24 PM

## Deploy Preview ([permalink](https://60e5d823958c4400077d8d34--elastic-blackwell-3aef44.netlify.app/))
Screen Shot 2021-07-07 at 12 39 38 PM

Screen Shot 2021-07-07 at 12 41 30 PM

PR #611, opened 2021-05-20
upgrade website to `lit@2`

thescientist13:
## Related Issue
related to this discussion on SSR
https://github.com/ProjectEvergreen/greenwood/discussions/576 and includes fix for #675 

In order for the website, using Lit to benefit from SSR, we would need to be on the latest version so figured I would start work on the discussion by introducing **Lit@2** as a first step

## Summary of Changes
1. [Upgraded website to **Lit@2**](https://lit.dev/docs/releases/upgrade/) and got website working with 1st party components (see TODO section 👇 )
1. Added `ExportAllDeclaration` case when walking a module with acorn
1. Did a little refactoring around setting entries into `importMap` and added bare import "re-map" support
1. Ensured backwards compatibility locally with [my website project](https://github.com/thegreenhouseio/www.thegreenhouse.io)
1. Per [Lit's website](https://lit.dev/docs/tools/requirements/#polyfills), shipping _polyfill-support.js_ with our polyfill plugin for now, but will refactor that into its own plugin(s) as part of #709 where dedicated Lit plugins will be created

Also, **LitElement** bundle size dropped significantly, going from 27KB -> 15KB!  ⚡ 

> _Will be doing some further research after this on how to actually implement SSR and hydration with Lit, as another PR (or more!)_

## TODO
1. [x] Need to get **@evergreen-wc** upgraded to **Lit@2**, or vendor for the short term
    - opened issue to initiate discussion - https://github.com/hutchgrant/evergreen-web-components/issues/33
1. [x] Need to get **lit-redux-router** upgraded to **Lit@2** (for SPA mode spec)
    - opened issue to initiate discussion https://github.com/fernandopasik/lit-redux-router/issues/55
    - re-enable SPA `mode` spec
1. [x] Get specs passing
1. [x] Needs #672 merged and rebased into this PR
1. [x] Find cause of big logo flash / FOUC - after more debugging, may be related to observations seen in the below TODO - https://github.com/ProjectEvergreen/greenwood/issues/675#issuecomment-914629130
1. [x] find a solution to https://github.com/ProjectEvergreen/greenwood/issues/675
1. [x] Breakout **plugin-node-modules** changes into standalone PR that should get merged first - https://github.com/ProjectEvergreen/greenwood/pull/613
1. [x] Lit perf benchmark
1. [x] Update website documentation and samples and merge (only when `Lit@2` is out of RC)
[...]

PR #650, opened 2021-07-04
typescript plugin

### LINKED ISSUE #647: TypeScript plugin
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
As TypeScript is becoming pretty prevalent in the frontend ecosystem, it would be good for us to have an official plugin for this, and is often featured prominently with newer projects and libraries, like [lit](https://lit.dev/playground/#sample=examples/hello-world-typescript).

## Details
It doesn't have to be anything too fancy, just something that would support the basics like in the sample above and should be able to accept a merging of user level `tsconfig.json#compilerOptions`.

```ts
import {html, css, LitElement} from 'lit';
import {customElement, property} from 'lit/decorators.js';

@customElement('simple-greeting')
export class SimpleGreeting extends LitElement {
  static styles = css`p { color: blue }`;

  @property()
  name = 'Somebody';

  render() {
    return html`<p>Hello, ${this.name}!</p>`;
  }
}
```

thescientist13:
resolves #647 

## Summary of Changes
1. Created a TypeScript plugin to wrap `tsc`
1. Add a typescript based component to the website to validate it works

> _**Note**: Should get merged after #642 to resolve this Rollup warning_
> ```sh
> @rollup/plugin-typescript: Rollup 'sourcemap' option must be set to generate source maps.
> ```

Screen Shot 2021-07-04 at 12 40 01 PM

> _Note: for TypeScript and new `Lit@2` ( #611 ) to work together, like seen in https://github.com/AnalogStudiosRI/www.analogstudios.net/pull/6, we will need to have #649 resolved, but no rush, just a heads up._

## TODO
1. [x] Add tests including custom _tsconfig.json_ overrides, like for `expirementalDecorators` (should get tests passing due to low coverage alert)
1. [x] Ensure live reload watching works for custom extensions resolving - this might require a standalone PR, since it might also be nice to make these extensions configurable outside of plugins, let me track as an issue - https://github.com/ProjectEvergreen/greenwood/issues/653
1. [x] Documentation (add to plugins list)
1. [x] Add (or document) linting TS with ESLint?
1. [x] Tracking additional follow up work here - https://github.com/ProjectEvergreen/greenwood/issues/658
1. [x] Pass through config to Rollup plugin
1. [x] Remove demo code


thescientist13:
Bot sure why we get a different value for the `[hash]` value in the filename from Rollup for [Windows](https://github.com/ProjectEvergreen/greenwood/runs/3031056080?check_suite_focus=true) vs [Linux / Mac]() but this is all Rollup so for the time being, not sure what we can do. 🤔 

The source is the same from what I can tell too.

_Windows_
Screen Shot 2021-07-10 at 6 28 22 AM

_macOS_
Screen Shot 2021-07-10 at 6 32 53 AM

Adding it to #658  for tracking.

Issue #652, opened 2021-07-06
inline optimization should inline everything


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Noticed in https://github.com/ProjectEvergreen/greenwood/pull/642#issuecomment-874788275 that we have an opportunity to make inline optimization setting more aggressive.

## Details
For example, in one of our specs, when we `import` from _node_modules_, that code itself is not inlined.

_Before_
```js
import { html } from 'lit-html';
import { LitElement } from 'lit-element';
```

_After_
```html
import"/lit-html.d69ea26f.js";import"/lit-element.ea26fec7.js";//# sourceMappingURL=1832243663-scratch.01361a49.js.map
```

To be truly inline so as not to initiate _ANY_ network requests, we would want all of that _node_modules_ code inlined too.

----

maybe at this time we could review renaming scratch to something more meaningful?

thescientist13:
Not sure if this is actually still an issue, but would be good to confirm so we can cross it off the list.

thescientist13:
Confirmed as working in https://github.com/ProjectEvergreen/greenwood/issues/581

PR #657, opened 2021-07-07
handle (inadvertent) lookahead patterns when using `replace` with inline JavaScript

### LINKED ISSUE #656: inline optimization string replace calls that have a lookahead Regex inside a template literal (`$``) in the replacement string breaks users pages
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Had been [working through an issue I saw](https://github.com/ProjectEvergreen/greenwood/pull/495) while testing our optimizations options and found a conflict between MPA mode and inlining optimization.

Basically the inlining was completely breaking the router, seemingly on this Regex inside [the shelf component](https://github.com/ProjectEvergreen/greenwood/blob/v0.13.0/www/components/shelf/shelf.js#L41)
image

image

With this being the area of code in question.
```js
 expandRoute(path) {
    let routeShelfListIndex = this.shelfList.findIndex(item => {
      let expRoute = new RegExp(`^${path}$`);
      return expRoute.test(item.route);
    });
```

## Details
After some digging, I found what [the issue was](https://github.com/ProjectEvergreen/greenwood/pull/495#issuecomment-875810894).  ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_a_parameter) saves the day!)

>  $` | Inserts the portion of the string that precedes the matched substring.

Essentially, the Regex inside the source code is getting interpreted by `replace` as a parameter it should do something with.

So I guess would need to go through and check all usages of `replace` to make sure we're handling source code appropriately or do some kind of escaping.  Seems like this would need to be done in a lot of places... 😬 
https://stackoverflow.com/a/22612228/417806

thescientist13:
## Related Issue
resolves #656 

## Summary of Changes
1. Added repro with (failing) test case (the bug [is here](https://github.com/ProjectEvergreen/greenwood/blob/v0.13.0/packages/cli/src/config/rollup.config.js#L432) specifically)
1. Enabled 'inline` mode for the website, [looks promising](https://github.com/ProjectEvergreen/greenwood/pull/657#issuecomment-879201657)

> _Note: given the need to audit replace usage, this will likely not land in the upcoming [`v0.14.0`](https://github.com/ProjectEvergreen/greenwood/issues?q=label%3Av0.14.0+) release as there is a large queue of PRs already lined up_

## TODOs
1. [x] Fix failing test case and validate with [Greenwood website test](https://github.com/ProjectEvergreen/greenwood/pull/495/files)
1. [x] Audit and address other potential "leaks" in using `replace` in the project


thescientist13:
Wow, with `inline` enabled, we've gotten down to 9 total request and 100KB transferred!  🤩 
Screen Shot 2021-07-13 at 11 45 08 AM

Compared to prod
Screen Shot 2021-07-13 at 11 46 05 AM

Just wait till we can complete #652 . almost be at 0 network requests!  🏆

Issue #557, opened 2021-04-16
leverage NodeJS resolve logic for looking up location of paths for node modules


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of looking into #505 as part of submitting #511 and so for then, was just getting something done that was quick and dirty, using unpkg.com as a fallback.  I'm not sure if this will support _everything_ like things added to `importMap`... but I did reach out in NodeJS slack channel and got this tip on using NodeJS [`require.resolve`](https://nodejs.org/api/modules.html#modules_require_resolve_request_options) and so that would certainly be interesting because if we could piggy back somehow of off Node's default resolve behavior, that would allow us to maintain all that on our own.

This would also ensure that when Greenwood is run from the command line using `npx`, _node_modules_ would always resolve correctly.

## Details
So essentially, give it a package name and it will look up the path for you.
```js
// /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/node_modules/es-module-shims/dist/es-module-shims.js
console.debug( require.resolve('es-module-shims') )
```

With a better local testing environment (using `npm|yarn link`), we could set it up so that it in **plugins-node-modules** `resolve` just delegates all this to `require.resolve` instead:
```js
// example url => /node_modules/es-module-shims/dist/es-module-shims.js
async resolve(url) {
  return new Promise((resolve, reject) => {
    try {
      const packageName = url.split('/')[1]; // ex. es-module-shims
      const packageFullPath = require.resolve(packageName); // ex. /Users/.../node_modules/es-module-shims/dist/es-module-shims.js
      const currentDirectory = packageFullPath.split('/node_modules/')[0]; //ex. /Users/.../
      const nodeModulesUrl = path.join(currentDirectory, url.replace.replace('/node_modules/'); //ex. /Users/.../es-module-shims/dist/es-module-shims.js

      resolve(nodeModulesUrl);
    } catch (e) {
      console.error(e);
      reject(e);
    }
  });
}
```

We might want to generate our own full path since the `url` is what we want based on walking _package.json_ (in theory) as opposed to what NodeJS _thinks_ we want.

----

[...]

PR #668, opened 2021-07-15
extract import map generation logic into utility

### LINKED ISSUE #537: create reusable walker, "package ranger" utility so that import map building logic can be shared
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
As part of developing a [**plugin-babel**](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/plugin-babel), it was observed that the logic for generating dynamic `importMap` entries from a package.json file would be a useful piece of logic.

## Details
Currently all this logic exists in the internal [**plugin-node_modules**](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/plugins/resource/plugin-node-modules.js#L72).  

I think the plan here would be to
1. Extract it to its own file in _cli/src/lib/_
1. Refactor **plugin-node_modules** to use it
1. Refactor **plugin-babel** to use it and remove the need for users to have to [peer `install` regenerator runtime](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/plugin-babel#options)
1. Bonus: Make a utility that helps ease the burden of [appending to an existing `importMap`](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/plugin-graphql/src/index.js#L40) in **plugin-graphql**

### LINKED ISSUE #538: refactor duplicate hardcoded dependencies list in plugin-graphql to a single shared data structure
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Refactor duplicate dependency logic in [**plugin-graphql**](https://github.com/ProjectEvergreen/greenwood/tree/release/0.10.0/packages/plugin-graphql)

## Details
Currently dependencies for the [`importMap`](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/plugin-graphql/src/index.js#L45) and [rollup](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/plugin-graphql/src/index.js#L109) are being hardcoded in two places.  It would be good to refactor this so that list is maintained once in the file and the needed output for both cases can be derived from the data structure instead.

thescientist13:
## Related Issue
resolves #537  (as a result of looking into https://github.com/ProjectEvergreen/greenwood/issues/557#issuecomment-877837997)

## Summary of Changes
1. Extract logic from **plugin-node-modules** into its own utility
1. Also resolved #538 

## TODOs
1. [x] extend to graphql plugin
1. [x] extend to babel plugin - https://github.com/ProjectEvergreen/greenwood/issues/906
    - Is going to require more work including installing dependencies and updating specs (there were none for this case!).  Should make a new issue for the 1.0 backlog tracker to account for installing those deps to the package, copying into the tests folder, stubbing out _package.json_.  Wonder if I could use walker package ranger to target against specific deps and build up the import map to shim that way???  🧐 

## Discussion
1. make a package?
    - pass a path or an object?  object assumes process.cwd()?
    - what kind of API?
<details>
1. Should be merged _after_ #557 , to account for hardcoded usages of _node_modules_?
    ```js
    // use .mjs version if it exists, for packages like redux
    if (!Array.isArray(entry) && fs.existsSync(`${process.cwd()}/node_modules/${packageJson.name}/${entry.replace('.js', '.mjs')}`)) {
      entry = entry.replace('.js', '.mjs');
    }
    ```
</details>


thescientist13:
Some refactoring ideas coming out of #832 
- just pass initial path to `walkModule` and let it do the fs.readFile?
- can't `updateImportMap` append `/node_modules/` hint string?

PR #669, opened 2021-07-16
Rfc/issue 570 context and theme packs

### LINKED ISSUE #570: Shared Workspaces and Contexts (Theme Packs)
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
There have been a couple discussions and opportunities observed coming out of #508 / #509 and while working on a private project idea of my own
-  plugin-graphql added the support to merge custom schemas provided by users in a _data/schema/_ directory in the user's work space.  Could plugins get that support as well, so plugins can ship custom queries and schemas?  This could could allow for entire plugins of functionality like adding a photo gallery that could handle mapping paths to resolvers
-  a private project of mine, where sharing templates and pages would allow plugins to provide entire "theme" packs with templates and styles ready to go out of the box

## Details
Off the top of my head, I wonder if this is something that could just be done by using `ResourceInterface` and hijacking any files that you own?  Or would you have to open up "context" to support providing additional directories for Greenwood to look at.

For example, to "share" a custom page template for use in a markdown file from a plugin
```md
---
template: 'acme-themepack-splashpage'
---

## Welcome to my site!

Lorum Ipsum
```

The page template would be able to resolve to something like _node_modules/acme-themepack/src/templates/acme-themepack.html_

----

Not sure if `ResourceInterface` is enough?
```js
class MyThemePackResource extends ResourceInterface {
  constructor(compilation, options) {
    super(compilation, options);
    
    this.extensions = ['.html'];
    this.contentType = 'text/html';
  }

  // if request matches a file we "own", resolve it
  // ex: `${userWorkspace}/templates/themepackXYZ-splash-page.html`
  async shouldResolve(url) {
    return Promise.resolve(true);
  }

}
```

Or if Greenwood needs to open up a new API, so creating this issue to track that cases are developed for, and to report back what the developer ergonomics are like and see if something new needs to be made.

---

For bonus points, we could also look to incorporate #574 ?  💡

thescientist13:
## Related Issue
resolves #570 

## Summary of Changes
1. Introduced the concept of a `context` Plugin type. as an `Array`
1. Those template paths are then used as part of `getAppTemplate` and `getPageTemplate` in **plugin-standard-html**
1. Also supports `template` from front matter
1. Added test cases emulating a "theme pack" like package
1. Created developer docs for `Context` plugin as well as a guide on _Creating a Theme Pack_
1. Allowed for customizing default location for pages and templates from _greenwood.config.js_

Example Usage
```js
module.exports = (options = {}) => [{
  type: 'context',
  name: 'my-theme-pack:context',
  provider: (compilation) => {
    return {
      templates: [
        path.join(compilation.context.projectDirectory, 'node_modules/my-theme-pack/dist/layouts')
      ]
    };
  }
}];
```

Then in a template called _layouts/title.html_ you can "share" a template like so
```html
<html>

  <head>
    <script type="module" src="/node_modules/my-theme-pack/dist/components/greeting.js"></script>  
  </head>

  <body>

    <h2>This is a custom page template called title from the layouts directory.</h2>
    <x-greeting></x-greeting>
    <content-outlet></content-outlet>

  </body>

</html>
```

Use it in frontmatter in your own workspace
```md
---
template: 'title'
---

### Title Page

#### Some content from my own title.md page.
```

Even combine it with a custom app template too, using _layouts/app.html_
```html
<html>

  <head>
    <link rel="stylesheet" href="/node_modules/my-theme-pack/dist/styles/theme.css"></link>
  </head>

  <body>
    <h1>This is a custom app template from the custom layouts directory.</h1>
    <page-outlet></page-outlet>
  </body>

</html>
```

Greenwood will automatically handle all the node resolution for that JS and CSS!!!  🏆 🙌 

## TODO
1. [x] Add example of using components (footer / header) - using `ResourceInterface`?
    - don't even need a `ResourceInterface`, just point everything to `node_modules` and Greenwood will take care of the rest!  Brings up another question though... 👇 
1. [x] How to [resolve these files for development though, where they will not be in _node_modules_](https://github.com/ProjectEvergreen/greenwood/pull/669#issuecomment-887909284)?  🤔 
    - should validate by merging #670 and writing some tests
1. [x] Should we suggest using `require.resolve` for finding the right node levels directory, a la #557 
    - I think as long as users use `/node_modules/<package_name>` convention, and Greenwood uses `require.resolve` correctly, then users can depend on Greenwood doing all the heavy lifting, which makes sense for such a fundamental mechanism like resolve _node_modules_.
1. [x] Should make sure we [can override a page, like an _index.html_](https://github.com/thescientist13/greenwood-starter-presentation/blob/master/src/pages/index.html) so users can provide a complete "interface" as part of their theme pack
     - see discussion section below
     - make a follow up issue to close this one out 👇 
     - https://github.com/ProjectEvergreen/greenwood/issues/681
1. [x] Address discussion points
    - ~~add a _greenwood.config.js_ option for context for overriding templates and pages dir (nice to have, might make separate issue)~~ - implemented!
    - ~~additional input formats (array vs. string)~~ - not needed at this time
    - ~~use an interface?~~ - doesn't seem warranted
    - ~~helpful utils for plugin authors?~~ - defer to the discussion and until we have a little more experience building these - **https://github.com/ProjectEvergreen/greenwood/discussions/682**
    - ~~best practices / naming conventions for users~~ - I think by virtue of having to have a unique package name on npm anyway, there would be automatic namespacing?
1. [x] Update documentation
1. [x] Create a guide for creating a "theme pack"

## Discussion
1. I think we should we support being able to specifically configure pages and templates directory from _greenwood.config.js_ as well.  The plugin approach is likely going to be more applicable to plugin authors, but for users, a "direct" configuration option, should they need it, seems best, so they can easily rename _pages/_ or _templates/_ directory name if they want to (without having to write a plugin just to do that)
    - well, one thing I realize now with pages is that those actually need to be pushed onto the graph _and_ resolved. Greenwood scans all pages first to build, so basically we also would need to incorporate #21 as well.  Since getting templates in is already a good start, I think we can / should at least merge what we have and make a new issue tracking support for pages via `context` and link to #21 and maybe they can both be done together?
1. Further API considerations
    - should we support more than just pages and templates?
    - should everything just be an array
    - should we have an interface
1. Documentation
   - encourage namespacing for template and file names?


thescientist13:
## Development v Publishing

So a couple thoughts here around handling developing a theme pack vs. distributing a theme pack since the location and paths of files would be different when develop (e.g. _src/some/path/_) as opposed to where they will ultimately be published to (e.g. _/path/to/node_modules/< package-name >/dist/etc_).

I think we can establish a couple ground rules here
1. Greenwood is likely going to be used (or will want to be used) in a development capacity by the plugin author (naturally)
1. Paths in templates should ideally not have to change from one context to the other if so (e.g. always assume _node_modules/_ in `<script>` or `<link>` tags)

In other words, it's probably a bit easier to bend Greenwood to make it easier for local development in a one-off fashion for a plugin author, rather than forcing them to have jump through a bunch of hoops or extra build steps when wanting to publish their templates.  Ideally when everything looks good, then can just publish everything as is and it will "just work".  🤞 ™️ 

So here are my thoughts 👇 

### Templates Location
Least hackiest approach I could think of right now is simply to have the theme pack plugin accept a secret `__isDevelopment` only flag for the plugin author to use, such that special logic (using [`npm ls <dep-name>`](https://stackoverflow.com/a/50841601/417806) can be determined from a _greenwood.config.js_ to tell if the plugin is installed or not, and use that to determine the route location for files.

Example
```js
// _greenwood.config.js_
const packageName = require('./package.json').name;
const myThemePackPlugin = require('./theme-pack-context-plugin');
const { spawnSync } = require('child_process');
const ls = spawnSync('npm', ['ls', packageName]);

module.exports = {
  plugins: [
    ...myThemePackPlugin({
      __isDevelopment: ls.stdout.toString().indexOf('(empty)') >= 0
    }]
};
```

```js
// in the theme pack plugin
const path = require('path');

module.exports = (options = {}) => [{
  type: 'context',
  name: 'my-theme-pack:context',
  provider: (compilation) => {
    const templateLocation = options.__isDevelopment // eslint-disable-line no-underscore-dangle
      ? path.join(process.cwd(), 'src/layouts')
      : path.join(context.projectDirectory, 'node_modules/my-theme-pack/dist/layouts');

    return {
      templates: [
        templateLocation
      ]
    };
  }
}];
```


### Paths in Templates
For paths in templates, e.g.
```html
<!DOCTYPE html>

  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1"/>
    <link rel="stylesheet" href="/node_modules/my-theme-pack/dist/styles/theme.css"></link>
  </head>

  <body>
    <h1>This is a custom app template from the custom layouts directory.</h1>
    <page-outlet></page-outlet>
  </body>

</html>
```

This looks a little cleaner, we can just have plugin authors use a `Resource` plugin for this in their _greenwood.config.js_ and never have to do anything in the plugin.  We can even make this part of Greenwood's "standard lib".
```js
class MyThemePackDevelopmentResource extends ResourceInterface {
  constructor(compilation, options) {
    super(compilation, options);
    this.extensions = ['*'];
  }

  async shouldResolve(url) {
    return Promise.resolve(url.indexOf('/node_modules/my-theme-pack/') >= 0);
  }

  async resolve(url) {
    return url.replace('/node_modules/my-theme-pack/dist/', path.join(process.cwd(), '/src/layout/'));
  }
}
```

With all that put together, and can pretty much keep the actual code the same, and still develop like a regular Greenwood "end user".  (using `develop` command)
Screen Shot 2021-07-27 at 4 54 21 PM

(so in this case, that URL doesn't actually exist 😉  )

PR #672, opened 2021-07-25
add import and directory support to export map subpath exports

### LINKED ISSUE #671: import map not correctly handling export maps that have subpath exports that are directories or are `import`
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Was trying to integrate a new version of [**lit-redux-router**](https://github.com/fernandopasik/lit-redux-router/issues/55) with **Lit@2** into a Greenwood router and was getting this error
Screen Shot 2021-07-25 at 11 46 08 AM

Looking at the the current importMap that is generated when developing, you'll notice that for a package like **lit-html,** where [its subpath exports](https://nodejs.org/api/packages.html#packages_subpath_exports) can actually be [_directories_](https://github.com/lit/lit/blob/main/packages/lit-html/package.json#L36), which is something we're not handling.
```html
"lit-html/directives/": "/node_modules/lit-html/directives//lit-html/directives/.js",
```
Screen Shot 2021-07-24 at 8 04 39 PM

## Details
Additionally, like for something like **regexparam**, we're not supporting [their `import` export](https://github.com/lukeed/regexparam/blob/master/package.json#L25).
Screen Shot 2021-07-25 at 11 44 17 AM

thescientist13:
## Related Issue
resolves #671 

## Summary of Changes
1. If subpath export is a directory, add each file to the `importMap`
1. Added support for subpath exports that are of type `import`

> Blocking #611 and https://github.com/AnalogStudiosRI/www.analogstudios.net/pull/8

## Discussion
I'm thinking now, based on the nature of this fix, if we actually stillvneed `walkModule` to manually crawl all the `import`s and `export`s of every dependency to add to the `importMap`?.  I would be curious to try now with this bug fix to see if our `importMap` still works without it?  Maybe should happen as part of a minor release though, so this PR can just go out as a patch.

Additionally, not sure if a sub path export is a directory, does that mean we should recursively export everything?  Could be part of an issue that also addresses the above question re: `walkModule`, which coincidentally, I've playing around with in #668 

Update: made a quick test ( https://github.com/ProjectEvergreen/greenwood/pull/672#issuecomment-888359598 ) so seems like it is still needed.

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-node-modules.js
+++ b/packages/cli/src/plugins/resource/plugin-node-modules.js
@@ -142,8 +142,10 @@ const walkPackageJson = (packageJson = {}) => {
             packageExport = esmPath
               ? esmPath
               : fallbackPath;
-          } else if (exportMapEntry.default) {
-            packageExport = exportMapEntry.default;
+          } else if (exportMapEntry.import || exportMapEntry.default) {
+            packageExport = exportMapEntry.import
+              ? exportMapEntry.import
+              : exportMapEntry.default;
             
             // use the dependency itself as an entry in the importMap
             if (entry === '.') {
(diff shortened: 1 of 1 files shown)


thescientist13:
As a simple test, tried commenting out the `walkModule` function calls and at least for this version of Lit, got this error, so seems like it is still needed.  
Screen Shot 2021-07-28 at 10 32 04 AM

Can re-evaluate further in #674 if desired, since no big changes you should really happen in a patch fix, IMO.

PR #511, opened 2021-04-08
Bug/issue 505 improve default npx support

thescientist13:
## Related Issue
resolves #505 

## Summary of Changes
1. Add fallback detection to unpkg.com if internal client side dependencies for **es-module-shims** and **@webcomponents/webcomponentjs** aren't found in `process.cwd()`
1. Add `npx` docs back to the home page and quick start
1. Add a test specifically to omit those dependencies from scaffolding (is just a copy / paste of default case)
1. Whitelist **unpkg.com** with puppeteer

## TODO
1. [x] noticed a bug when trying to merge user page templates while still using default app template, so will want to make a bug and fix that first, then can enable the `xdescribe` spec here and then we can merge.
1. [x] rebase after #513 is merged
 
## Notes
For now, just getting this quick and dirty happy path working, using unpkg.com as a fallback.  But I did reach out in NodeJS slack channel and got this tip on using NodeJS [`require.resolve`](https://nodejs.org/api/modules.html#modules_require_resolve_request_options).

Essentially, give it a package name and it will look up the path for you.
```js
// /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/node_modules/es-module-shims/dist/es-module-shims.js
console.debug( require.resolve('es-module-shims') )
```

With a better local testing environment (using `npm|yarn link`), we could set it up so that it in **plugins-node-modules** `resolve` just delegates all this to `require.resolve` instead:
```js
// example url => /node_modules/es-module-shims/dist/es-module-shims.js
async resolve(url) {
  return new Promise((resolve, reject) => {
    try {
      const packageName = url.split('/')[1]; // ex. es-module-shims
      const packageFullPath = require.resolve(packageName); // ex. /Users/.../node_modules/es-module-shims/dist/es-module-shims.js
      const currentDirectory = packageFullPath.split('/node_modules/')[0]; //ex. /Users/.../
      const nodeModulesUrl = path.join(currentDirectory, url.replace.replace('/node_modules/'); //ex. /Users/.../es-module-shims/dist/es-module-shims.js

      resolve(nodeModulesUrl);
    } catch (e) {
      console.error(e);
      reject(e);
    }
  });
}
```

We want to generate our own full path since the `url` is what we want based on walking _package.json_ (in theory) as opposed to what NodeJS _thinks_ we want.

----

[...]

Issue #505, opened 2021-04-04
npx commands showing error for not found node modules paths


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [x] Bug
- [ ] Other (please clarify below)

## Summary
When running the `npx` commands for using Greenwood, there are some errors for files not found in _node_modules_,

ex.
```sh
% npx @greenwood/cli serve
Downloading Chromium r818858 - 132.4 Mb [====================] 100% 0.0s
-------------------------------------------------------
Welcome to Greenwood ♻️
-------------------------------------------------------
Running Greenwood with the serve command.
Initializing project config
Initializing project workspace contexts
Generating graph of workspace files...
Started local development server at localhost:1984
Serializing pages at http://127.0.0.1:1984
pages to generate
 src/pages/index.md
serializing page... /

  Error: ENOENT: no such file or directory, open '/Users/owenbuckley/Workspace/scratch/node_modules/es-module-shims/dist/es-module-shims.js'


  Error: ENOENT: no such file or directory, open '/Users/owenbuckley/Workspace/scratch/node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js'

Serializing complete for page /.
done serializing all pages
@rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to `true`, as the next major version will default this option to `true`.
copying graph.json...
Initializing project config
Initializing project workspace contexts
Generating graph of workspace files...
Started production test server at localhost:8080

  Error: ENOENT: no such file or directory, open '/Users/owenbuckley/Workspace/scratch/node_modules/es-module-shims/dist/es-module-shims.js'


  Error: ENOENT: no such file or directory, open '/Users/owenbuckley/Workspace/scratch/node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js'
```

## Details
This is likely due to the fact that certain dependencies are assumed to be in the user's current directory where a _node_modules_ directory would be.  When running with `npx` then this wouldn't exist, so how to resolve these???

![Screen Shot 2021-04-03 at 6 46 06 PM](https://user-images.githubusercontent.com/895923/113524200-3ead6880-957b-11eb-9ea8-7bd6a004d7b4.png)

thescientist13:
Nice, it works now with no error from us.
```
% mkdir -p src/pages
[...]

PR #674, opened 2021-07-26
Enhancement/issue 557 resolve node modules using node require resolve

### LINKED ISSUE #557: leverage NodeJS resolve logic for looking up location of paths for node modules

thescientist13:
## Related Issue
resolves #557 

## Summary of Changes
1. Updated **plugin-node-modules** to leverage `require.resolve` instead of assuming where _node_modules_ is located on disk (in an effort to better handle #505 / #511 type situations)
2. Per the addition of testing in support of #570, had to add some falback / exception handling in case `require.resolve` blows up like happened for our theme pack specs (should see if I can mock this better or make an issue / discussion)
```sh
Error: Cannot find module 'my-theme-pack'
Require stack:
- /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/plugins/resource/plugin-node-modules.js
- /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/config.js
- /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/compile.js
- /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/commands/build.js
- /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
    at Function.resolve (internal/modules/cjs/helpers.js:94:19)
    at NodeModulesResource.resolve (/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/plugins/resource/plugin-node-modules.js:29:85)
    at /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/serve.js:4:421
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/serve.js:4:216
    at async /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/serve.js:4:216
    at async /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/serve.js:4:216
    at async /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/serve.js:4:216
    at async /Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/serve.js:4:216 {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/plugins/resource/plugin-node-modules.js',
    '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/config.js',
    '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/lifecycles/compile.js',
    '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/commands/build.js',
    '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/cli/src/index.js'
  ]
}
```

> Initial and ongoing analysis and experiments being conducted in this repo - https://github.com/thescientist13/node-resolve-experiments

Was able to confirm things seem to be working as expected now that we are using `require.resolve` in the `npx` test case by changing the content to say `<h2>hello require.resolve</h2>`
```sh
% cd packages/cli/test/cases/build.default.quick-start-npx
owenbuckley@Owens-MBP-2 build.default.quick-start-npx % npx greenwood build
Welcome to Greenwood (v0.14.1) ♻️
Running Greenwood with the build command.
Initializing project config
Generating graph of workspace files...
Started local development server at localhost:1984
Prerendering pages at http://127.0.0.1:1984
pages to render
 src/pages/index.md
prerendering page... /
prerendering complete for page /.
done prerendering all pages
copying graph.json...
owenbuckley@Owens-MBP-2 build.default.quick-start-npx % cat public/index.html
<!DOCTYPE html><html lang="en" prefix="og:http://ogp.me/ns#">
                        <head>
                        <link rel="modulepreload" href="/header.4557af3c.js" as="script">
                      <!-- Shady DOM styles for app-header --><style>body {transition: opacity ease-in 0.2s; }
body[unresolved] {opacity: 0; display: block; overflow: hidden; position: relative; }
</style>





    <title>My App</title>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <meta name="mobile-web-app-capable" content="yes">
    <meta name="apple-mobile-web-app-capable" content="yes">
    <meta name="apple-mobile-web-app-status-bar-style" content="black">


            <script type="module" src="/header.4557af3c.js"></script>

          </head>

  <body>


    <div>
      <app-header>
      <header class="style-scope app-header">This is the header component.</header>
    </app-header>
      <h2>hello require.resolve</h2>
    </div>



 </body></html>




```

## TODO
1. [x] Fix failing specs for Windows
1. [x] Check for any other hardcoded lookup paths for _node_modules_ when reading from the filesystem in Greenwood (I think ideally it should only need to happen in the plugin though)
    - left some initial thoughts here - https://github.com/ProjectEvergreen/greenwood/pull/674#issuecomment-887838612
1. [x] Update `npx` test case to not require **unpkg** fallback and / or figure out a way to test locally
1. [x] Probably a little room left for some cleanup / refactoring
    -  not sure if we need to [recursively support scoped packages](https://github.com/ProjectEvergreen/greenwood/pull/674/files#diff-e6731c225dc2a78d6caf722587c6ac2eacd3908ce5e23f13a3b6058ab3d7d183R203)?  e.g. can they have names like `@scoped-packages/something/else` - reached out in NodeJS slack, will see, but I think it would definitely be an edge case
    - wondering if we should change our logic [here](https://github.com/ProjectEvergreen/greenwood/blob/v0.14.1/packages/cli/src/plugins/resource/plugin-node-modules.js#L25) at all?  Had some [good lessons learned here](https://github.com/thescientist13/node-resolve-experiments) that might be applicable in terms of prioritization and just make sure we have it right.  👌 - I think it is OK for now
    - re-re-evaluate need for [`walkModule`](https://github.com/ProjectEvergreen/greenwood/pull/672#issuecomment-888359598)? - maybe for another PR / discussion

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -159,6 +159,37 @@ To test from a VM, you can
 
 You can disable plugins in _webpack.config.prod.js_ to remove production optimizations for testing purposes.
 
+## npx Testing
+[`npx`](https://www.npmjs.com/package/npx) is a useful CLI utitlity bundled with NodeJS that allows users to run npm packages globally but without having to install them.
+
+```sh
+% npx http-server
+Starting up http-server, serving ./public
+Available on:
+  http://127.0.0.1:8080
+  http://192.168.1.153:8080
+Hit CTRL-C to stop the server
+```
+
+It's featured on the Greenwood website [home page](https://www.greenwoodjs.io/) and in the [Quick Start guide](https://www.greenwoodjs.io/getting-started/quick-start/#command-line) as an option for using Greenwood.  There is a [spec for it](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/build.default.quick-start-npx) to try and simulate running it.
+
+You can use it for local development with Greenwood by using [npm link](https://docs.npmjs.com/cli/v7/commands/npm-link) to make `greenwood` available to your local CLI.
+```sh
+# From the root of the Greenwood repo
+$ npm link
+
+# then say in a test case folder
+$ cd packages/cli/test/cases/build.default.quick-start-npx
+$ npx greenwood
+-------------------------------------------------------
+Welcome to Greenwood (v0.14.1) ♻️
+-------------------------------------------------------
+Running Greenwood with the  command.
+
+          Error: not able to detect command. try using the --help flag if
+          you're encountering issues running Greenwood.  Visit our docs for more
+          info at https://www.greenwoodjs.io/docs/.
+```
 
 ## Docker
 A Docker container is available within the project to use as a development environment if you like.  It is configured to use the same image that runs as part of the project's [Continuous Integration environment](https://github.com/ProjectEvergreen/greenwood/blob/master/.github/workflows/ci.yml#L9).
--- a/packages/cli/src/lib/browser.js
+++ b/packages/cli/src/lib/browser.js
@@ -42,8 +42,7 @@ class BrowserRunner {
 
       if (
         interceptedRequestUrl.indexOf('http://127.0.0.1') >= 0 ||
-        interceptedRequestUrl.indexOf('localhost') >= 0 ||
-        interceptedRequestUrl.indexOf('unpkg.com') >= 0
+        interceptedRequestUrl.indexOf('localhost') >= 0
       ) {
         interceptedRequest.continue();
       } else {
--- a/packages/cli/src/plugins/resource/plugin-node-modules.js
+++ b/packages/cli/src/plugins/resource/plugin-node-modules.js
@@ -205,7 +205,7 @@ class NodeModulesResource extends ResourceInterface {
     return Promise.resolve(url.indexOf('node_modules/') >= 0);
   }
 
-  sync resolve(url) {
+  async resolve(url) {
     const packagePathPieces = url.split('node_modules/')[1].split('/'); // double split to handle node_modules within nested paths
     let packageName = packagePathPieces.shift();
     let nodeModulesUrl;
--- a/packages/cli/src/plugins/resource/plugin-node-modules.js
+++ b/packages/cli/src/plugins/resource/plugin-node-modules.js
@@ -264,7 +264,7 @@ class NodeModulesResource extends ResourceInterface {
   async intercept(url, body) {
     return new Promise((resolve, reject) => {
       try {
-        const { userWorkspace, projectDirectory } = this.compilation.context;
+        const { userWorkspace } = this.compilation.context;
         let newContents = body;
         const hasHead = body.match(/\<head>(.*)<\/head>/s);
 
@@ -280,10 +280,6 @@ class NodeModulesResource extends ResourceInterface {
           : fs.existsSync(`${process.cwd()}/package.json`)
             ? require(path.join(process.cwd(), 'package.json'))
             : {};
-        const esShimsFilename = '/node_modules/es-module-shims/dist/es-module-shims.js';
-        const esShimsPath = fs.existsSync(path.join(projectDirectory, esShimsFilename))
-          ? esShimsFilename
-          : 'https://unpkg.com/es-module-shims@0.5.2/dist/es-module-shims.js';
         
         // walk the project's pacakge.json for all its direct dependencies
         // for each entry found in dependencies, find its entry point
@@ -291,9 +287,10 @@ class NodeModulesResource extends ResourceInterface {
         // and then walk its package.json for transitive dependencies and all those import / exports
         walkPackageJson(userPackageJson);
 
+        // apply import map and shim for users
         newContents = newContents.replace('<head>', `
           <head>
-            <script defer src="${esShimsPath}"></script>
+            <script defer src="/node_modules/es-module-shims/dist/es-module-shims.js"></script>
             <script type="importmap-shim">
               {
                 "imports": ${JSON.stringify(importMap, null, 1)}
--- a/packages/cli/src/plugins/resource/plugin-node-modules.js
+++ b/packages/cli/src/plugins/resource/plugin-node-modules.js
@@ -206,7 +206,7 @@ class NodeModulesResource extends ResourceInterface {
   }
 
   async resolve(url) {
-    const packagePathPieces = url.split('node_modules/')[1].split('/'); // double split to handle node_modules within nested paths
+    const packagePathPieces = this.getBareUrlPath(url).split('node_modules/')[1].split('/'); // double split to handle node_modules within nested paths
     let packageName = packagePathPieces.shift();
     let nodeModulesUrl;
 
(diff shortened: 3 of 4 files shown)


thescientist13:
## Development v Publishing

So a couple thoughts here around handling developing a theme pack vs. distributing a theme pack since the location and paths of files would be different when develop (e.g. _src/some/path/_) as opposed to where they will ultimately be published to (e.g. _/path/to/node_modules/< package-name >/dist/etc_).

I think we can establish a couple ground rules here
1. Greenwood is likely going to be used (or will want to be used) in a development capacity by the plugin author (naturally)
1. Paths in templates should ideally not have to change from one context to the other if so (e.g. always assume _node_modules/_ in `<script>` or `<link>` tags)

In other words, it's probably a bit easier to bend Greenwood to make it easier for local development in a one-off fashion for a plugin author, rather than forcing them to have jump through a bunch of hoops or extra build steps when wanting to publish their templates.  Ideally when everything looks good, then can just publish everything as is and it will "just work".  🤞 ™️ 

So here are my thoughts 👇 

### Templates Location
Least hackiest approach I could think of right now is simply to have the theme pack plugin accept a secret `__isDevelopment` only flag for the plugin author to use, such that special logic (using [`npm ls <dep-name>`](https://stackoverflow.com/a/50841601/417806) can be determined from a _greenwood.config.js_ to tell if the plugin is installed or not, and use that to determine the route location for files.

Example
```js
// _greenwood.config.js_
const packageName = require('./package.json').name;
const myThemePackPlugin = require('./theme-pack-context-plugin');
const { spawnSync } = require('child_process');
const ls = spawnSync('npm', ['ls', packageName]);

module.exports = {
  plugins: [
    ...myThemePackPlugin({
      __isDevelopment: ls.stdout.toString().indexOf('(empty)') >= 0
    }]
};
```

```js
// in the theme pack plugin
const path = require('path');

module.exports = (options = {}) => [{
  type: 'context',
  name: 'my-theme-pack:context',
  provider: (compilation) => {
    const templateLocation = options.__isDevelopment // eslint-disable-line no-underscore-dangle
      ? path.join(process.cwd(), 'src/layouts')
      : path.join(context.projectDirectory, 'node_modules/my-theme-pack/dist/layouts');

    return {
      templates: [
        templateLocation
      ]
    };
  }
}];
```


### Paths in Templates
For paths in templates, e.g.
```html
<!DOCTYPE html>

  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1"/>
    <link rel="stylesheet" href="/node_modules/my-theme-pack/dist/styles/theme.css"></link>
  </head>

  <body>
    <h1>This is a custom app template from the custom layouts directory.</h1>
    <page-outlet></page-outlet>
  </body>

</html>
```

This looks a little cleaner, we can just have plugin authors use a `Resource` plugin for this in their _greenwood.config.js_ and never have to do anything in the plugin.  We can even make this part of Greenwood's "standard lib".
```js
class MyThemePackDevelopmentResource extends ResourceInterface {
  constructor(compilation, options) {
    super(compilation, options);
    this.extensions = ['*'];
  }

  async shouldResolve(url) {
    return Promise.resolve(url.indexOf('/node_modules/my-theme-pack/') >= 0);
  }

  async resolve(url) {
    return url.replace('/node_modules/my-theme-pack/dist/', path.join(process.cwd(), '/src/layout/'));
  }
}
```

thescientist13 (commented):
Might hold this back for the next release instead of going out with theme packs and 0.15.0 just to avoid too many big changes happening at once?  Will see.

thescientist13:
Made another issue to evaluate pushing this idea forward more if it indeed does workout.  🙌 
https://github.com/ProjectEvergreen/greenwood/issues/684

thescientist13:
I wonder if this means we can remove _all_ the setup work on our specs now that requires copy / pasting files out of root node_modules into the test cases directory?  🤔

PR #687, opened 2021-08-06
exclude node modules from user workspace resolution

### LINKED ISSUE #686: node modules paths are being resolved to "matching" user workspace paths causing broken package loading
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
After upgrading a handful of projects to Greenwood v0.15.0, the SPA projects, which also happen to be on Lit@2 RC versions, started to experience this issue _just_ when running `greenwood develop`.  Running `greenwood serve` was fine.
Screen Shot 2021-08-06 at 12 03 30 PM

Seems like a regression then if they were all working fine before?  Should try and solve ASAP though along with keeping an eye on #675.

## Details
Examples of the issue
- https://github.com/AnalogStudiosRI/www.analogstudios.net/pull/16
- https://github.com/ContributaryCommunity/www.contributary.community/pull/94

Other projects, (those not using **Lit@2** anyway), seem to be fine with the upgrade, so there's that going for us at least 😅 
- https://github.com/thegreenhouseio/www.thegreenhouse.io/pull/235
- https://github.com/ProjectEvergreen/greenwood-getting-started/pull/43
- https://github.com/ProjectEvergreen/projectevergreen.github.io/pull/60

thescientist13:
## Related Issue
resolves https://github.com/ProjectEvergreen/greenwood/issues/686#issuecomment-894470074

## Summary of Changes
1. Ignore resolving _node_modules_ paths from user workspace resolution, leading to "false" positives

For example in the test case, if we _didn't_ have this fix, requesting both files would always return the contents of _src/lit-html.js_.

## TODO
1. [x] Add a test case to cover this scenario

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-user-workspace.js
+++ b/packages/cli/src/plugins/resource/plugin-user-workspace.js
@@ -20,7 +20,7 @@ class UserWorkspaceResource extends ResourceInterface {
     const isAbsoluteWorkspaceFile = fs.existsSync(path.join(userWorkspace, bareUrl));
     const workspaceUrl = isAbsoluteWorkspaceFile
       ? isAbsoluteWorkspaceFile || bareUrl === '/'
-      : this.resolveRelativeUrl(userWorkspace, bareUrl);
+      : url.indexOf('node_modules') < 0 && this.resolveRelativeUrl(userWorkspace, bareUrl);
 
     return Promise.resolve(workspaceUrl);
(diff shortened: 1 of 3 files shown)


thescientist13:
Created a discussion to further refine how this resolution can work going forward - https://github.com/ProjectEvergreen/greenwood/discussions/691

PR #697, opened 2021-08-13
pull card image under greenwoods hosting

thescientist13:
## Related Issue
N / A

## Summary of Changes
1. Point `og:image` to URL [hosted / managed by Greenwood's website already](https://www.greenwoodjs.io/assets/greenwood-logo-300w.png) to avoid dependency on external system (S3)

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -20,7 +20,7 @@ module.exports = {
     { property: 'og:title', content: 'Greenwood' },
     { property: 'og:type', content: 'website' },
     { property: 'og:url', content: 'https://www.greenwoodjs.io' },
-    { property: 'og:image', content: 'https://s3.amazonaws.com/hosted.greenwoodjs.io/greenwood-logo.png' },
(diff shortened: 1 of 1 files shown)


greenwood.config.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
     { property: 'og:title', content: 'Greenwood' },
     { property: 'og:type', content: 'website' },
     { property: 'og:url', content: 'https://www.greenwoodjs.io' },
-    { property: 'og:image', content: 'https://s3.amazonaws.com/hosted.greenwoodjs.io/greenwood-logo.png' },
+    { property: 'og:image', content: 'https://www.greenwoodjs.io/assets/greenwood-logo-300w.png' },
```
thescientist13, line 23: `300w` seems the right size for a card. I presume, but we have a few sizes to pick from if we need an alternate
https://github.com/ProjectEvergreen/greenwood/tree/master/www/assets

PR #712, opened 2021-09-01
HTML Include Plugin

### LINKED ISSUE #788: HTML Include Plugin
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
In the spirit of the since [abandoned HTML Imports spec](https://www.html5rocks.com/en/tutorials/webcomponents/imports/) that was originally part of the init Web Components "feature suite", and given the renewed [interest in bringing it back](https://github.com/whatwg/html/issues/2791), I thought it might be nice to bring this feature to Greenwood, in a more static context, as opposed to something that would happen purely at runtime.  _Effectively, this proposal / plugin would
would take the HTML include spec and realize it as more a build time templating system for HTML.  The goal here is to enable developers the ability to ship more static HTML while allowing the authoring context to be JavaScript **and** leveraging standard semantics and web expectations._. 💚 

> **Note**: I think if you want this feature in its most strictest sense of the word, I would recommend the [**<html-include>**](https://github.com/justinfagnani/html-include-element) custom element, which provides a runtime implementation of this as a Web Component.

Some of the work has already been spiked in #712 and so this issue will aim to formally bring it to live.  🙌 

## Details
I see two initial flavors of supporting this out of the gate that I would feel comfortable supporting and maintaining as a plugin and would provide HTML and JS based implementations.

> _It should be noted that this plugin would more or less be another way to achieve what the [`static` optimization setting]() already does._

### `<link>` Tag (HTML only)
The simplest flavor would follow the spec more closely and I think primarily address the use case of where you have static HTML that you want to reuse across your pages, like a global header or footer.

So given a snippet of HTML, e.g.
```html
<!-- src/includes/header.html -->
<style>
  header h1.my-include {
    text-align: center;
    color: red;
  }
</style>

<header class="my-include">
  <h1>Welcome to my website!<h1>
</header> 
```

In a page template, you could then do this
```html
<html>

  <body>
    <!-- the location can be anywhere in the workspace, I just chose to call it includes -->
    <!-- rel and href attributes would be required -->
    <link rel="html" href="/includes/header.html"></link>

    <h2>Hello 👋</h2>
  
  </body>

<html>
```

And Greenwood would statically generate this
```html
<html>

  <body>
    <style>
      header h1.my-include {
        text-align: center;
        color: red;
      }
    </style>

    <header class="my-include">
      <h1>Welcome to my website!<h1>
    </header> 

    <h2>Hello 👋</h2>
  
  </body>

<html>
```

### Custom Element (JavaScript)
For more advanced use cases where customization of the output may need to be done on a case by case basis, the other flavor would be a more JavaScript / web components flavor, allowing developers to return functions for generating markup and data and then Greenwood will build the HTML for them on the fly.  This effectively aims to fill the gap where just static HTML alone would not be sufficient enough.

So using the [Greenwood footer as an example](https://github.com/ProjectEvergreen/greenwood/blob/v0.19.0-alpha.2/www/components/footer/footer.js), that displays the project version based on reading the contents of a _package.json_ file, let's create a JS file that `export`s two functions; `getTemplate` and `getData`
```js
// src/components/footer/footer.js
const getTemplate = async (data) => {
  return `
    <app-footer>
      <style>
        footer {
          grid-area: footer;
        }
        ...
      </style>

      <footer class="footer">
        <h4>
          <a href="/">Greenwood v${data.version}</a> <span class="separator">&#9672</span> <a href="https://www.netlify.com/">This site is powered by Netlify</a>
        </h4>
      </footer>
    </app-footer>
  `;
};

const getData = async () => {
  const version = require('../../package.json').version;

  return { version };
};

module.exports = {
  getTemplate,
  getData
}; 
```

In a page template, you could then do this
```html
<html>

  <body>
    <h2>Hello 👋</h2>

    <app-footer src="../components/footer/footer.js"></app-footer>  
  </body>

<html>
```

And Greenwood would statically generate this
```html
<html>

  <body>
    <h2>Hello 👋</h2>

    <app-footer>
      <style>
        footer {
          grid-area: footer;
        }
      </style>

      <footer class="footer">
        <h4>
          <a href="/">Greenwood v0.19.0-alpha.2</a> <span class="separator">◈</span> <a href="https://www.netlify.com/">This site is powered by Netlify</a>
        </h4>
      </footer>
    </app-footer>  
  </body>

<html>
```

IMO, think the JS flavor will really come to shine more when Greenwood adds support for #708 , and then you could use this for displaying user / session data, or serverlessly at the edge!

thescientist13:
## Related Issue
resolves #788 

<details>
> _Note: [this issue will manifest in this PR](https://github.com/ProjectEvergreen/greenwood/pull/712#issuecomment-945019172) until #782 is merged_
</details>

## Summary of Changes
1. Created a new plugin to reflect the design derived from this spike (see below details)
1. Updated website documentation and implemented the custom element variant for the website footer

Interesting observation about the footer [here](https://github.com/ProjectEvergreen/greenwood/pull/712#issuecomment-961174082)

<details>
1. Spiking out the idea of what a plugin might look like for supporting the old deprecated standard of [HTML includes](https://github.com/whatwg/html/issues/2791) (not to be confused with [HTML modules / imports](https://github.com/WICG/webcomponents/issues/645))

Screen Shot 2021-09-01 at 5 48 57 PM
Screen Shot 2021-09-01 at 5 49 22 PM

## TODO
Going further, would like to see how far we can take it if we were to support something like this (notice the _.js_ extension)
```html
<link rel="html" href="/includes/html-include-test.html.js"></link>
```

```js
const getIncludeTemplate = async (data) => {
  const { name } = data;
  
  return `Hello ${name || 'World'}! 👋`;
}

export { getIncludeTemplate };
```

Then the plugin could just `import` from that file and statically stamp out the contents that way?  🤔 
</details>


thescientist13:
Hmm... not sure why this is seemingly causing the docs pages to fail if you got to a page, and then click its own link????  😞 
https://deploy-preview-712--elastic-blackwell-3aef44.netlify.app/docs/
Screen Shot 2021-10-16 at 7 07 37 PM

thescientist13:
Interesting observation but I noticed while working on this PR that on the main site, the (font?) size of the footer is slightly smaller on the home page vs any of the other pages / templates.  Not sure if just me?
greenwood-footer-font-size

cc: @aholtzman

PR #721, opened 2021-09-10
Node 14x minimum version

thescientist13:
## Related Issue
N / A

## Summary of Changes
Not sure why [master branch failed](https://github.com/ProjectEvergreen/greenwood/actions/runs/1222579667) but our [CI didn't fail for the same PR](https://github.com/ProjectEvergreen/greenwood/runs/3536283002?check_suite_focus=true), but figured that since [Node 12 is in maintenance mode](https://nodejs.org/en/about/releases/) and will be unsupported soon, can't hurt to align this now and see if it stabilizes things.
Screen Shot 2021-09-10 at 4 20 59 PM

## TODO
1. [x] Get things passing
1. [x] Make an issue to track this deprecation as a good first issue seen in the test logs - no longer in the codebase
    ```sh
    Build Greenwood With: 
    (node:5249) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
    ```
Screen Shot 2021-09-10 at 6 07 43 PM

### DIFF
--- a/.github/workflows/ci-win.yml
+++ b/.github/workflows/ci-win.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [14]
+        node: [14, 16]
 
     steps:
     - uses: actions/checkout@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
 
     strategy:
       matrix:
-        node: [14]
+        node: [14, 16]
 
     steps:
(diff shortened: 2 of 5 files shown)


thescientist13:
Hmm, seems this is broken here, but [ode 12 is still working](https://github.com/ProjectEvergreen/greenwood/actions/runs/1222708748) again,  so no rush on this then, going to move to draft.

Issue #734, opened 2021-09-21
clean up unnecessary scrollbars on the website


thescientist13:
## Type of Change
- [ ] New Feature Request
- [x] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Not sure if its something new to our code, but it seems like we've acrrued some extra scrollbars now??

#### Home Page
Horizontal scrollbar
https://www.greenwoodjs.io
![Screen Shot 2021-09-21 at 12 48 21 PM](https://user-images.githubusercontent.com/895923/134213951-542bfd9b-ca75-467c-89e9-e68923b3d7d0.png)

#### Docs / Guides / etc pages
Double vertical scrollbars
https://www.greenwoodjs.io/docs
https://www.greenwoodjs.io/guides
![Screen Shot 2021-09-21 at 12 50 04 PM](https://user-images.githubusercontent.com/895923/134213906-9eb1d769-e657-4143-b69c-2c61eec25c6a.png)
![Screen Shot 2021-09-21 at 12 53 46 PM](https://user-images.githubusercontent.com/895923/134214000-13b7b524-8cc3-4e9f-901d-44ba4c0602d1.png)

theprojectcode:
Is this issue still open? I would like to contribute.

thescientist13:
Hey @theprojectcode , thanks for checking in on this one!

While it looks like there is already a PR open for this, I think these other two issues would be really helpful to us, if you're interested?
- #718 
- #643

If they do (or you're looking for other types of contribution areas) feel free to let us know, we do appreciate the interest and help.  😊

theprojectcode:
@thescientist13 can I contribute to ticket #643? I'd be happy to do that one.

thescientist13:
Definitely @theprojectcode !  I've assigned it to you, and feel free to reach out in that particular if you have any questions.  (or you can join us in our Slack too)

theprojectcode:
Awesome! @thescientist13 I'll work on it!

PR #735, opened 2021-09-21
fix scrollbar issues, small styling adjustments

aholtzman:
## Related Issue

resolves issu #734
## Summary of Changes
1. [x] Fixed width on cards section on home screen for horizontal scroll
1. [x] Changed method of keeping sidebar sticky to eliminate second scrollbar
1. [x] Changed some styling for cleaner/easier to read docs

### DIFF
--- a/www/components/shelf/shelf.css
+++ b/www/components/shelf/shelf.css
@@ -50,18 +50,6 @@
     }
   }
 
-  & hr {
-    display: block;
-    unicode-bidi: isolate;
-    margin-block-start: 0.5em;
-    margin-block-end: 0.5em;
-    margin-inline-start: auto;
-    margin-inline-end: auto;
-    overflow: hidden;
-    border-style: inset;
-    border-width: 1px;
-  }
-
   & div > ul {
     padding-left: 0;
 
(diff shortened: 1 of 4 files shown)


thescientist13 (commented):
Thanks @aholtzman .  Took a quick spin and scrollbars are gone!  🙌 

I realized I had grown attached to the `<hr>` and shading in the menu items now that I am seeing them gone, but for most other dev websites, I don't see any type of effect there either, so this changes does make things look a little cleaner and goes go in an overall more consistent direction.  I think from my side, the horizontal rules and / or shading helped give some distinction and structure to the nav so everything wasn't swimming in. a sea of white, but it's not bad per se without it either.  Hah, I guess i'm not much help here.  😅   (but definitely thank you for this, definitely not blocking this or anything, to be clear)

Will be interested to see how others see it.

I did notice it looks like we lost our sticky footer though?
https://deploy-preview-735--elastic-blackwell-3aef44.netlify.app/guides/
Screen Shot 2021-09-22 at 11 49 34 AM

----

On the topic of shading and accents, it looks like there is a gutter / drop shadow effect around the content on any of the non home pages?  Was this part of the original design?  I like it but just wanted to make sure it was intentional and not some odd padding / margin issue that just happened to look good too.  🙃
Screen Shot 2021-09-22 at 11 46 23 AM

thescientist13 (approved):

PR #740, opened 2021-09-28
hail mary kitchen sink at the wall

### LINKED ISSUE #739: intermittent pre-rendered content duplication with Lit@2 and MPA mode
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
As part of #611 and piggy backing off of #675 (which I think was inadvertently tracking two issues), I am spinning off the [remaining issue I still see](https://github.com/ProjectEvergreen/greenwood/pull/611#issuecomment-926018893) where every handful of times I run `yarn build` locally, I get a bunch of duplicated content and static HTML in the output folders for one of the top level pages and the styles are all broken.

So sometimes it will be the home page, sometimes it will be the Getting Started page, etc.  Was able to reproduce [in Netlify](https://61535d27acf077000834741c--elastic-blackwell-3aef44.netlify.app/) and I've attached a [zip file of all the built assets](https://github.com/ProjectEvergreen/greenwood/files/7246568/deploy-61535d27acf077000834741c.zip) where you can see the duplicate HTML in _index.html_

Screen Shot 2021-09-23 at 1 44 51 PM

Screen Shot 2021-09-23 at 1 44 51 PM

> I usually test by just clicking each link and refreshing, and going back to the home page and clicking through again.  I haven't seen it happen in Netlify yet.

## Details
I've opened a PR with a bunch of random stuff that I thought might help but it still keeps happening every few runs, so not really sure what to do next?  It must be something in puppeteer since that is where all the mangled HTML keeps coming from, so I think that's the best surface area to keep testing against.

> _One thing to consider is that if after a few more hours of time on this no solution is reached, it might be better to focus our efforts on #708 and just provide Lit's first class SSR solution instead._

- Found a _somewhat_ related issue in the puppeteer repo though it is related to [duplicate content in _screenshots_](https://github.com/puppeteer/puppeteer/issues/1576) 
- test with just `ssg` mode, or otherwise disable features in the website code and / or Greenwood?
- lit@2 is [additive now](https://github.com/lit/lit/issues/2189#issuecomment-925890641), so not sure if this could account for the duplication?  Or how to even test / confirm it is an issue?
- Need to use a [`waitForTimeout` / `waitForSelector` / `waitForResponse`](https://github.com/puppeteer/puppeteer/issues/1875#issuecomment-403441680) call instead?
- If nothing after all that, create a sample repro repo
- How to test for this...

thescientist13:
## Related Issue
resolves #739 

## Summary of Changes
Trying some things to see what sticks, even though none of them have helped so far, (but will probably want to keep some of these anyway in #611 )
1. Upgrade **@webcomponentsjs**
1. Set Yarn resolution for single version of Lit
1. New puppetter args - `'--enable-experimental-web-platform-features'`

Still not working 😞

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,9 @@
     "lint:css": "stylelint \"./www/**/*.js\", \"./www/**/*.css\"",
     "lint": "ls-lint && yarn lint:js && yarn lint:ts && yarn lint:css"
   },
+  "resolutions": {
+    "lit": "^2.0.0"
+  },
   "devDependencies": {
     "@ls-lint/ls-lint": "^1.9.2",
     "@typescript-eslint/eslint-plugin": "^4.28.2",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -25,7 +25,7 @@
   "dependencies": {
     "@rollup/plugin-node-resolve": "^9.0.0",
     "@rollup/plugin-replace": "^2.3.4",
-    "@webcomponents/webcomponentsjs": "^2.4.4",
+    "@webcomponents/webcomponentsjs": "^2.6.0",
     "acorn": "^8.0.1",
     "acorn-walk": "^8.0.0",
     "commander": "^2.20.0",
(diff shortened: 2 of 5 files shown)


thescientist13:
phew, think I have exorcised the demon!  👻 

opened #743

Issue #48, opened 2019-04-18
HMR for local development


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As opposed to watch / livereload, which refreshes the full browser on each local development change, enabling [HRM (hot module replacement)](https://webpack.js.org/concepts/hot-module-replacement/) can help speed up local development by only reloading the part of the page, that needs to change, not the whole application.

## Details
For example, given a page with an `<input>` (for example):
1. Type something in the input element (like your name)
1. Change the CSS background of the page

expected: the page would change color, but the text in the input would remain

thescientist13:
[Interesting thread from React land](https://twitter.com/timneutkens/status/1251942933953228802)

Would be cool if this could work with GraphQL as well, but we can certainly develop it, enabled only by a flag, with certain caveats documented; e.g. might not work for all use cases, YMMV, etc.

thescientist13:
This cool project from **pikapkg**
https://github.com/pikapkg/esm-hmr

thescientist13:
There is some work in the Community Protocols space happening around this discussion that may be worth following up on whenever (if) we get around to this - https://github.com/webcomponents-cg/community-protocols/issues/6

thescientist13:
Could be a good article for reference
https://bjornlu.com/blog/hot-module-replacement-is-easy

PR #760, opened 2021-10-10
enhancement/issue 629 cache unchanged assets in development

### LINKED ISSUE #629: long term cache workspace and node module files in development
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
During development, files like node_modules rarely changing, at least while the dev server is running.  Source code files only only change when they, well, change.

This means we should be able to do some really easy and intelligent caching during the lifetime of the development session so repeated requests to files on disk, like node_modules, don't have to initiate a filesystem read.  (See https://github.com/ProjectEvergreen/greenwood/discussions/628 )

## Details
I think the technique for this is using an [E-Tag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag), which stores a hash of the content as a header and uses it for comparing at request time to see if the header still matches and if not, just return a 304 Not Modified Header.  For _node_modules_ we can assume this after first fetch.  For user code, only the actual JS, CSS, etc file that changed should actually be refetched.

Additionally, we only need to build [`importMap`](https://github.com/ProjectEvergreen/greenwood/blob/v0.13.0/packages/cli/src/plugins/resource/plugin-node-modules.js#L244) once as well, so we should cache that too and not do it on `intercept` of an HTML page.

thescientist13:
## Related Issue
resolves #629 

## Summary of Changes
1. Cache `importMap` from initial walking of all dependencies
1. Set and check [`ETag` and `If-None-Match` headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) respectively as part of dev server request / response middleware.  Also setting `Cache-Control` header too.
1. Created shareable `hashString` function
1. Upgraded to latest version of **es-module-shims** dependency
1. Added a mention of this to the _How It Works_ section

Benchmarks - https://github.com/ProjectEvergreen/greenwood/pull/760#issuecomment-1046120992

## TODOs 
1. [x] A / B testing against master branch
1. [x] Figure out why [Chrome doesn't seem to support this for ESM specifically](https://stackoverflow.com/questions/43659756/chrome-ignores-the-etag-header-and-just-uses-the-in-memory-cache-disk-cache)?  Is fine with CSS.  maybe need to set a [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
1. [x] Should only happen during development
1. [x] Have packages share `hashString` implementation
1. [x] Verify if **es-module-shims** is [causing double renders / script executions](https://github.com/guybedford/es-module-shims#no-load-event-retriggers)?
    - I think default behavior is working fine, so no need to mess with anything IMO
    
## Testing / Demo
So far this works great in Firefox, but testing (as in this feature) can be thought of as a stepping stone to [HMR](#48).  The value here is in seeing if development is / feels faster with these changes as opposed to not having it. 

⚠️ Make sure you don't have _Disable Cache_ set in your dev tools!  ⚠️ 

1. Open a new tab with the network tab open
1. Load the website with the `develop` command and observe the status of everything should be `200`
Screen Shot 2021-10-10 at 12 07 58 PM
1. Refresh and now the status should change to `304`
Screen Shot 2021-10-10 at 12 08 14 PM
1. Change a file and observe now that there is new content, the status has changed back to `200` and the page should have changed, but everything else stays `304`
Screen Shot 2021-10-10 at 12 08 33 PM
1. Keep refreshing, and the status should go back to `304` for everything
Screen Shot 2021-10-10 at 12 08 45 PM

## Questions
1. [x] Should it be configurable?  Or at least documented for awareness?
    - It's always easy enough to just set_Disable Cache_ from the dev tools to bypass, which we can document as well.
1. [x] I notice with live reload, just saving with no changes causes a reload, I wonder if it only looks for a timestamp?  Not sure if it's something we can do but it would be great if no change didn't cause a reload, like https://github.com/postcss/postcss-cli/issues/320 - added to my personal backlog
1. [ ] Where should this happen, before or after post-processing in middleware chain?  If first, everything after has to do `null` check on `ctx.body`.  Maybe it's something we could apply to `ResourceInterface`?
1. [ ] Should this _exclude_:
    - binary files
    - API calls - Yes
    - external HTTP requests - Yes
    - HTML?

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -31,7 +31,7 @@
     "acorn-walk": "^8.0.0",
     "commander": "^2.20.0",
     "cssnano": "^5.0.11",
-    "es-module-shims": "^0.5.2",
+    "es-module-shims": "^1.2.0",
     "front-matter": "^4.0.2",
     "koa": "^2.13.0",
     "livereload": "^0.9.1",
--- a/packages/cli/src/lib/hashing-utils.js
+++ b/packages/cli/src/lib/hashing-utils.js
@@ -9,6 +9,6 @@ function hashString(inputString) {
   return Math.abs(h).toString();
 }
 
-module.exports = {
+export {
   hashString
 };
--- a/packages/cli/src/lib/hashing-utils.js
+++ b/packages/cli/src/lib/hashing-utils.js
@@ -1,0 +1,14 @@
+// https://gist.github.com/hyamamoto/fd435505d29ebfa3d9716fd2be8d42f0#gistcomment-2775538
+function hashString(inputString) {
+  let h = 0;
+
+  for (let i = 0; i < inputString.length; i += 1) {
+    h = Math.imul(31, h) + inputString.charCodeAt(i) | 0; // eslint-disable-line no-bitwise
+  }
+
+  return Math.abs(h).toString();
+}
+
+module.exports = {
+  hashString
+};
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -143,6 +143,7 @@ async function getDevServer(compilation) {
           ctx.status = 304;
           ctx.body = null;
           ctx.set('Etag', etagHash);
+          ctx.set('Cache-Control', 'no-cache');
         } else if (!inm || inm !== etagHash) {
           ctx.set('Etag', etagHash);
         }
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -128,20 +128,23 @@ async function getDevServer(compilation) {
   // ETag Support - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag
   app.use(async (ctx) => {
     const body = ctx.response.body;
-    // console.debug('body', body);
+    const { url } = ctx;
 
-    if (Buffer.isBuffer(body)) {
-      // console.warn(`no body for => ${ctx.url}`);
-    } else {
-      const inm = ctx.headers['if-none-match'];
-      const etagHash = hashString(body);
+    // don't interfere with extrenal requests or API calls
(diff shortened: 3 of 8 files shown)


thescientist13:
## Benchmarking

So did some local testing with Chrome and Firefox on master vs this branch.  Results are basically almost a 50% percent improvement for refreshed and live reloads, which is pretty awesome!  With E-Tags, unchanged assets stay the same naturally, which you can by their E-Tag staying the same, which adds a really nice level of granularity to loading.  💪 

It should be noted Firefox is pretty fast overall, compared to Chrome.  Just saying. 🦊 ✨ 

### Master

#### Firefox
**Load: ~1.10 - 1.20s**
Screen Shot 2022-02-19 at 5 41 06 PM

**Refresh: 650-785ms**
Screen Shot 2022-02-19 at 5 41 15 PM

**Live Reload: 650-763ms**
Screen Shot 2022-02-19 at 5 41 27 PM

#### Chrome

**_First Load: ~1.8 - 2s_**
Screen Shot 2022-02-19 at 5 46 54 PM

**_Manual Refresh: ~800ms - 1s_**
Screen Shot 2022-02-19 at 5 47 06 PM

**_Live Reload: ~1.9s_**
Screen Shot 2022-02-19 at 5 47 18 PM

----

### Current Branch

#### Firefox
**Load: ~800ms**
Screen Shot 2022-02-19 at 5 50 56 PM

**Refresh: ~80ms**
Screen Shot 2022-02-19 at 5 51 13 PM


**Live Reload: ~375-450ms**
Screen Shot 2022-02-19 at 5 51 28 PM


#### Chrome
**Load: ~1.83s**
Screen Shot 2022-02-19 at 5 57 02 PM

**Refresh: 500-600ms**
Screen Shot 2022-02-19 at 5 57 13 PM

**Live Reload: 450-600ms**
Screen Shot 2022-02-19 at 5 58 00 PM

PR #778, opened 2021-10-23
Feature/777 greenwood init command

### LINKED ISSUE #777: It would be nice to have a greenwood init/template scaffold command
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
As per discussion #770 it would be nice to have a command to scaffold a minimal template as well as scaffold other user-gen templates.

## Details

It's pretty much what we did with create-evergreen-app, we copy a basic template package.json with a simple src/pages/index.md as well as a gitignore

hutchgrant:
## Related Issue
Resolves #777

## Summary of Changes

* adds new `@greenwood/init` package that will copy a default minimal template to the current working directory and optionally install dependencies via `--install` also optionally through yarn with `--yarn`


thescientist13 (commented):
Cool, so this is off to a good start!  Left a couple small comments inline but just wanted to clarify our package structures real quick.

It generally follows this convention
```sh
/
package.json
README.md
src/
  index.js
tests/
  ...
```

So any of these template files or whatever we want to publish to npm, would just go into _src/_, e.g
```sh
/
package.json
README.md
src/
  index.js
  templates/
     greenwood.config.js
     pages/
       index.md
     ...
tests/
     ...
```

Then the only config we need in _package.json_ is 
```json
{
   "files": [
     "src/"
   ]
}
```
So thinking that might help in indicating where the best place for some of these files should go and help reduce some of the duplication.

---

Other than that, leaving this comment just to track at a high level what I see as remaining to get this across the finish line
1. [x] _README.md_ for this new package.  Doesn't have to be too lengthy, since this is a pretty straight forward package with one feature at the moment.
1. [x] Should add something to our [Quick Start docs](https://www.greenwoodjs.io/getting-started/quick-start/) highlighting how to use it to get started.  Maybe even the project's README too?
1. [x] Add a couple small test cases:
    - One to make sure the initial scaffolding output is correct (kind of like the `eject` command specs)
     - One to run the `init` command and makes sure `greenwood build` runs without error
     - One to run the `init` command and makes sure `greenwood develop` runs without error

thescientist13 (commented):
> I would also say that I have to more do work to yarn install && yarn develop after scaffold, as opposed to in the situation of only wanting to scaffold, having to add a flag for --scaffold-only.

To be clear, I get where you are coming in wanting to help developers save a little time and it is commendable, but you just need to be really careful when you start baking in opinionated workflows on all users that that forces users down two operations, one of them being quite lengthy potentially (`install`).  And then for anybody else not using the "blessed" package manager / workflow, they have look up an override to just get "the feature".  

On top of that, I'm still not clear how or where you presented how you would know:
1. What package manager the user is using (or if they even want to run an install right now)
1. What command they want to run (or if they even want to run a commond right now)


> As well, as demonstrated, other popular software does the same thing(installs dependencies after scaffold).

I would be very cautious about leaning to strong on a single data point, since it's just as easy to show another popular project, [**SvelteKit**](https://kit.svelte.dev/) that does it how I propose
```sh
npm init svelte@next my-app
cd my-app
npm install
npm run dev -- --open
```

> And if the other software you reference is Gatsby Recipies, then I don't feel Recipies is 1:1 to one with what we're trying to do right now given that:
> 1. The user has to have their CLI installed, so is already probably able to determine the package manager that way)
> 1. Has to write a bunch of IaaC (basically a plugin config in MDX)
>
> That isn't an API I really want emulate / maintain to be quite honest.

----

Honestly, all the above points aside, it really comes to do this point for me:

> IF someone, for whatever reason, wants to only scaffold, they can simply add that flag. If you dont want to run dev server by default, that's fine I'll add a different flag option for that or remove it. It seems like a nice thing to have, only needing to run init to scaffold, install, and launch dev server at once.

The scope of this package as outlined in the discussion was already single focused; _initializing (scaffolding)_ a Greenwood project by generating the boilerplate files to get a runnable project that _every_ user is going to need.  Thus, it's not clear to me why these supplemental and last mile use cases are now overtaking this issue and our time and discussions.  

_Beyond how to build it is, is knowing what we're building._

I thought from the discussion the plan was to start by getting the heavy lifting out of the way, and create some templates and boilerplate files so creating a blog or similar site would be possible.  You know, the really tediuos stuff.

_Then_ we would layer on features through a single flag that could prompt a user through whatever options we want to provide, using something like [**enquirer**](https://www.npmjs.com/package/enquirer) to straight up ask the user:
1. What package manager are you using? (defaults to npm)
1. Would like to add `{SASS|TypeScript|whatever}` to your project? (defaults to none)
1. Install deps automatically (defaults to yes)

And so the real point here is that our time is limited, and the hope in starting the discussion was to agree on an understanding of a shared vision / plan for the feature and the path to get there, so it could be accomplished through quick iterations that also be paralleized once the initial PR landed (this one).  Just want to make sure we are able to make the best use of our time and try and find the quickest path from A -> B for everyone's sake.  (us and users alike).

thescientist13 (approved):
OK, this should be good to go if anyone wants to do any final reviews.  

I made the following updates
1. [Fixed the name of the test (no _s_) directory](https://github.com/ProjectEvergreen/greenwood/pull/778/commits/d5e89aaab174a3609d4deea89ed6868f8a1050ca), that is why it wasn't getting picked up by the coverage reporter, now coverage is OK
1. Updated build and develop specs to use the `--install` flag
1. Added a spec for `init` and only the scaffolding
1. Noticed [`--yarn` flag wasn't implemented, so enabled that and added spec](https://github.com/ProjectEvergreen/greenwood/pull/778/commits/40789c889da6c55f1f852cdd5ca236efd2322a4f)
1. Added [_Init Command_ option / section to the **Getting Started** docs](https://github.com/ProjectEvergreen/greenwood/pull/778/commits/fd7e8a4150c62a3c3747482ac0bffd28a73407dd)

----

Made a follow up issue to track keeping **@greenwood/cli** version in sync - https://github.com/ProjectEvergreen/greenwood/issues/781

.ls-lint.yml
```diff
@@ -13,4 +13,5 @@ ls:
 
 ignore: 
   - .git 
-  - node_modules
\ No newline at end of file
+  - node_modules
+  - packages/init/node_modules
```
thescientist13, line 17: Ah, yes, it seems there is no glob support at the moment in **ls-lint**, will need to [wait for v2](https://github.com/loeffel-io/ls-lint/issues/35#issuecomment-737986143) it seems

.mocharc.js
```diff
@@ -2,5 +2,5 @@ const path = require('path');
 
 module.exports = {
   spec: path.join(__dirname, 'packages/**/test/**/**/**/*.spec.js'),
-  timeout: 30000
+  timeout: 60000
```
thescientist13, line 5: increased this to support running `install` as part of init specs.

PR #794, opened 2021-11-08
exclude 404 route from MPA partial generation

### LINKED ISSUE #793: MPA mode generating malformed partial output file for 404 page
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
Seemingly related to the introduction of #200, it looks like Greenwood is generating an unexpected file path for a for a __routes404.html_ when outputting to the output directory. 

greenwood-routes-404

```sh
% ls -ls public
total 552
.
.
48 -rw-r--r--   1 owenbuckley  staff  23954 Nov  8 10:53 404.html
 8 -rw-r--r--   1 owenbuckley  staff   1243 Nov  8 10:53 493492190-cache.json
 8 -rw-r--r--   1 owenbuckley  staff   2721 Nov  8 10:53 574823255-cache.json
 8 -rw-r--r--   1 owenbuckley  staff   3278 Nov  8 10:53 606222270-cache.json
 8 -rw-r--r--   1 owenbuckley  staff    258 Nov  8 10:53 756510913-scratch.73749a99.js.map
 0 drwxr-xr-x   9 owenbuckley  staff    288 Nov  8 10:53 _routes
24 -rw-r--r--   1 owenbuckley  staff   8711 Nov  8 10:53 _routes404.html
 0 drwxr-xr-x   7 owenbuckley  staff    224 Nov  8 10:53 about
.
.
```

## Details
404 page shouldn't really be generating a partial since it is designed to be its own page, so this not even fixing the file path, but would want to just eliminate this file from being generated at all.

thescientist13:
## Related Issue
resolves #793 

## Summary of Changes
1. Fixes the thing 😄


packages/cli/src/plugins/resource/plugin-optimization-mpa.js
```diff
@@ -33,7 +33,7 @@ class OptimizationMPAResource extends ResourceInterface {
   }
 
   async shouldOptimize(url) {
-    return Promise.resolve(path.extname(url) === '.html' && this.compilation.config.mode === 'mpa');
+    return Promise.resolve(url !== '404.html' && path.extname(url) === '.html' && this.compilation.config.mode === 'mpa');
```
thescientist13, line 36: If we ever get around to adding a 500 page or something, we might want a better system for identifying this "special" pages

PR #842, opened 2021-12-31
feat/841 add --template argument and clone template to init package

### LINKED ISSUE #841: As a developer, it would be nice if I could quickly scaffold from a curated list of templates via cli
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
As discussed [here](https://github.com/ProjectEvergreen/greenwood/discussions/770#discussioncomment-1599917) it would be nice to have a list of curated template repos in which to scaffold from

## Details

1. We fetch a list of `greenwood-template-x` repos
2. We utilize the init scaffold package with a `--template` arg which will then render a list of templates from the project evergreen organization from which you can scaffold
3. on select, the repo will be cloned and scaffolded to your project

hutchgrant:
## Related Issue
Resolves #841 

## Summary of Changes
1. Adds `--template` option to the `init` command
2. lists all repos with `greenwood-template-x` in name present in ProjectEvergreen org
3. clones template and then scaffolds the app

## Run Example
```
git checkout feat/841-template-command && yarn install
cd .. && mkdir testblog && cd testblog
node ../greenwood/packages/init/src/index.js --template --yarn --install
```

```
Initialize Greenwood Template ♻️
Yarn Enabled
? Which template would you like to use? (Use arrow keys)
❯ blog 
```

Once merged of course you'll be able to do this with:
```
npx @greenwood/init --template
```

## todo

* [x] docs
* [x] tests

### DIFF
--- a/packages/init/src/index.js
+++ b/packages/init/src/index.js
@@ -41,7 +41,7 @@ const program = new commander.Command(scriptPkg.name)
   .usage(`${chalk.green('<application-directory>')} [options]`)
   .option('--yarn', 'Use yarn package manager instead of npm default')
   .option('--install', 'Install dependencies upon init')
-  .option('--template', 'Select from list of Greenwood curated templates')
+  .option('--template [type]', 'Select from list of Greenwood curated templates')
   .parse(process.argv)
   .opts();
 
@@ -197,16 +197,31 @@ const listAndSelectTemplate = async () => {
     }
   ];
 
-  return inquirer.prompt(questions).then((answers) => {
-    // set the selected template based on the selected template name
-    selectedTemplate = templates.find(template => {
-      return template.name === answers.template;
-    });
+  // if the user has provided one, use that, else prompt from the list
+  if (typeof program.template !== 'boolean') {
+    const userSelection = program.template;
+    const matchedTemplate = templates.find((template) => template.name === userSelection);
+    
+    if (matchedTemplate) {
+      console.debug(`using user provided template => ${userSelection}...`);
+      selectedTemplate = matchedTemplate;
+    } else {
+      const choices = templates.map(template => template.name).join('\n');
 
-    if (selectedTemplate) {
-      console.log('\Installing Selected Template:', selectedTemplate.name);
+      console.error(`unable to match user provided template "${userSelection}". please try again.  choices are ${choices}`);
     }
-  });
+  } else {
+    return inquirer.prompt(questions).then((answers) => {
+      // set the selected template based on the selected template name
+      selectedTemplate = templates.find(template => {
+        return template.name === answers.template;
+      });
+
+      if (selectedTemplate) {
+        console.log('\Installing Selected Template:', selectedTemplate.name);
+      }
+    });
+  }
 };
 
 const cloneTemplate = async () => {
@@ -235,6 +250,7 @@ const run = async () => {
   try {
     if (program.template) {
       await listAndSelectTemplate();
+
       if (!selectedTemplate) {
         return;
       }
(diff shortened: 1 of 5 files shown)


packages/init/src/index.js
```diff
@@ -130,8 +139,106 @@ const install = async () => {
...
+      const repos = await fetch(PROJECT_API_URL).then(resp => checkStatus(resp));
+
+      // assuming it did resolve but there are no templates listed
+      if (!repos || repos.length === 0) {
+        console.log('Couldn\'t locate any templates, check your connection and try again');
+        return [];
+      }
+
+      const templateRepos = repos.filter(repo => {
+        return repo.name.includes('greenwood-template');
```
thescientist13, line 175: would this be an opportunity to use the variable; `templateStandardName`?
```js
return repo.name.includes(templateStandardName);
```

packages/init/src/index.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 /* eslint no-console: 0 */
+/* eslint-disable camelcase */
```
thescientist13, line 3: Just thinking out load here, but is this ESLint exception this because of` PROJECT_API_URL`?  I guess the uppercase isn't really a big deal for this variable is it?  It seems we use a lot of strings as constant "hints" throughout our code (`templateStandardName` also seems like one too? 🤔 ) but never capitalize them all either.

I wonder if we should
1. either just go all in on caps for any `const` that is a primitive (boolean, string, number)
1. just turn this ESLint rule off?

packages/init/src/index.js
```diff
@@ -130,8 +139,106 @@ const install = async () => {
...
+const cloneTemplate = async () => {
+  const git = simpleGit();
+
+  // check if .template directory already exists, if so remove it
+  if (await fs.existsSync(clonedTemplateDir)) {
+    await fs.rmSync(clonedTemplateDir, { recursive: true, force: true });
+  }
+
+  // clone to .template directory 
+  console.log('clone template', selectedTemplate.name, 'to directory', clonedTemplateDir);
```
thescientist13, line 225: minor, but don't forget about template literals in these cases as well
```js
console.log(`clone template ${selectedTemplate.name} to directory ${clonedTemplateDir}`);
```

packages/init/src/index.js
```diff
@@ -130,8 +139,106 @@ const install = async () => {
...
+  try {
+    await git.clone(selectedTemplate.clone_url, clonedTemplateDir);
+    templateDir = clonedTemplateDir;
```
thescientist13, line 228: Should we also clean up `clonedTemplateDir` too after we scaffold the template?  Also, as we've stablished _.greenwood_ as our scratch folder for build output and temp files / operations, would it make sense to keep _.template_ inside of that instead of making extra top level hidden directory in user's project?

thescientist13 (commented):
Looking good!

I wonder if it makes sense for any non-default features / flags / options to live in their own file?  _index.js_ seems like it is growing and keeping all the default functions in there, while using `import` to grab the rest out of other files, may help keep things organized and maintainable / readable over time?

---

Aside from that, if docs and tests can land in the next couple of days, happy to include this in the #840 release!

packages/init/src/index.js
hutchgrant: Had the same thoughts.

packages/init/src/index.js
hutchgrant: both good points

packages/init/src/index.js
thescientist13: Meant to add that I'm ok with option no. 2

thescientist13 (approved):
Sweet, added some specs and docs, and updated the blog post PR, this is good to go!  Thanks for helping make this feature better and better!  💯

PR #857, opened 2022-01-12
Feature/issue 855 prerender default to false

### LINKED ISSUE #855: make `prerender` config option false by default
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Coming out of #576 and #812 (and as observed in #639) puppeteer as a heavy dependency.  It's not new news by any means, but even for a simple getting started case like `npx @greenwood/cli`, for just a simple markdown file you still have to download puppeteer and a whole copy of chrome.  As ["Mr. Wonderful"](https://abc.com/shows/shark-tank/cast/kevin-o-leary) would say; "stop the madness!"

Also, it uses npm `postinstall` hooks which is usually limited / blocked in cloud environments like Stackblitz.

## Details
So by default, Greenwood should
1. Set `prerender` to false
1. Don't install puppeteer by default (make it an [`optionalDependecny`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies)?)
1. If user sets `prerender: true` and tries to run the build, check for puppeteer and if needed, install on the fly

thescientist13:
## Related Issue
resolves #855 

## Summary of Changes
1. Set `prerender` to `false` by default
1. Updated specs accordingly (since puppeteer was always adding its own `<style></style>` tags
1. Updated _Getting Started_ guide with a step for optimizing Web Components with `prerender` and `optimization` configuration options.

> This change shaved about 3m off the time it takes to run our tests!  🏃‍♂️ 

## TODO
1. [x] Found and need to fix two issues unmasked due to default puppeteer `<style></style>` tag behavior
    - https://github.com/ProjectEvergreen/greenwood/issues/858
    - https://github.com/ProjectEvergreen/greenwood/issues/859
1. [x] Updated `prerender` config docs
1. [x] Updated Getting Started guide for prerender and static optimization "advanced" configuration - feature/issue-855-prerender-default-to-false
1. [ ] Should [**greenwood-template-blog**](https://github.com/ProjectEvergreen/greenwood-template-blog/) become more "static" (`prerender` and `static` options) or leave as is?
1. [ ] (nice to have) Implement optional **puppeteer** installation - #916


thescientist13:
Based on conversation and feedback with the team on the weekly call, these are my thoughts:
- Carry this through as is for alpha.0
- In a second PR, will see if I can get auto detect and install for puppeteer, otherwise will just detect and error to the user if pretender is set to true and no custom renderer.  Will also indicate this in the docs

Issue #710, opened 2021-08-31
Create a Lit based render plugin that provides Lit based SSR + hydration


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of ongoing discovery in #576 , after enabling https://github.com/ProjectEvergreen/greenwood/issues/709, we should make a **@greenwood/plugin-renderer-lit** package as an alternative to the default puppeteer rendering option used by Greenwood.

## Details
Effectively as an alternative to puppeteer and should facilitate hydration via Declrative ShadowDOM

Thoughts / Questions
  - Should Lit be the default rendered? Or only after we can provide native `HTMLElement` support via #548 /  https://github.com/lit/lit/discussions/2092
  - Should make sure to communicate the pros / cons for using puppeteer(browser) vs a JS only / server solution, and what kind APIs are  aren't available like `window.location`
- Once we get SSR hydration, should create a new option to enable it for SSG / MPA modes too?

thescientist13:
Wonder if it would make sense to have a JSDOM renderer too (as another option?).  Then you have a couple choices before you have to start tapping into framework level renderers if you're just trying to stick to `HTMLElement` (though I guess you still need some sort of "hydrator", e.g. #548 ?

Issue #709, opened 2021-08-31
Add support for additional (SSR) rendering options through a render plugin type


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of ongoing discovery in #576 , one the key action item needed is to be able to facilitate different rendering options so there can be an alternative to only being able to use puppeteer.

## Details
As part of adding this new render plugin, puppeteer should be moved into a plugin of its own under the `@greenwood` namespace and become a `peerDependency`.  

As bonus points, would be nice to see if this could only be installed as needed to avoid the issue of having to pull down an entire browser if don't even have any JS.  (e.g. the `npx` hello world / getting started scenarios)

When #855 is added, by virtue of a user providing their own custom renderer, we should set `prerender: true` and use their renderer instead.

PR #869, opened 2022-01-17
Feature/issue 709 custom render plugin

thescientist13:
## Related Issue
#709 / #710

## Summary of Changes
1. Introduced a renderer plugin API
1. Introduced a Lit renderer plugin
1. Added headers to Resource plugins for optimize to support both static (file) and server (route) contexts
1. Added documentation for custom render plugin API

Screen Shot 2022-01-17 at 1 08 58 PM

## TODO
1. [x] make greenwood implementation a plugin
1. [x] Investigate how to delineate between render (server) vs prerender (client) and understand how to custom render an SSG site (w/ hydration?)
1. [x] html plugin refactoring
1. [x] Tests 
1. [x] Docs
1. [x] Remove demo code

## Thoughts / Questions
1. [ ] Should we use [VM context](https://github.com/lit/lit/tree/main/packages/labs/ssr#rendering-in-a-separate-vm-context) for Lit plugin?
    - It _is_ experimental, so maybe no, just try and contain it all in the Worker? (worker should provide isolation, right?)
    - FWIW, I think because we are in a Worker though (for reloading ESM in development), we are able to avoid issues like this, so maybe these are equivalent?  (Or at least use Workers until VM context gets unflagged)
        - https://github.com/lit/lit/issues/2393
        - https://github.com/lit/lit/issues/2391
1. [x] Ongoing Investigations (merge into the release branch issue tracking list)
    - static rendering w/ custom renderer, e.g. _www/_ + Lit (just have experimental support right now) and define patterns / practices
    - hydration support for SSG & SSR (how do we get the data in an SSG context though... ?)
    - streaming support for HTML via generators would be awesome!


thescientist13:
Good catch on the call @hutchgrant

I was trying to be a bit too clever maybe by having an option for SSR routes to be rendered by a custom renderer (e.g. Lit) _AND_ also puppeteer, but that seems excessive since the whole point of a custom renderer is to avoid that overhead.  Puppeteer was basically giving us its own HTML instead of the nice HTML that Lit gave us with declarative shadow DOM and all the nice hydration markers.

Verified it works as expected now and also updated that spec to test the affirmative (explicit name matches) as opposed for testing the negative (absence of default placeholder text).  Thanks!  🙌 
Screen Shot 2022-01-27 at 9 06 59 AM

Issue #708, opened 2021-08-31
Introduce an SSR mode via Greenwood configuration


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
As part of ongoing discovery in #576 , one the key action items needed is to enable a true SSR mode that will allow running Greenwood as a process on a server such that at least for the `serve` command, will effectively just run the koa server to generate pages / templates on the fly but serving bundled rollup code.

## Details
Basically we need to add [`ssr` as a mode](https://www.greenwoodjs.io/docs/configuration/#mode) such when a user runs `npm run serve`, we build and bundle all the pages and assets like normal, but additionally keep the Koa server going to keep serving requests to pages and content but in a built / production context.

Thoughts / questions
1. Still need to explore a bit more how server side templating and data fetching would work
1. Now that we need Koa for more than a dev server, should explore using something like [**fastify**](https://www.fastify.io/)?

PR #875, opened 2022-02-02
add configuration for prod server port

thescientist13:
## Related Issue
#708 

_**Question**: Should we make develop default port something more familiar like `3000`?_ - leaving it at `1984` 😄 

## Summary of Changes
1. Added the ability to customize server port
1. Added docs and serve spec


hutchgrant (approved):
lgtm

PR #876, opened 2022-02-05
Chore/ci windows increase network timeout

thescientist13:
## Related Issue
Noticed we get a lot of intermitted timeouts on `yarn install` on Windows
- https://github.com/ProjectEvergreen/greenwood/runs/5074493269?check_suite_focus=true
- https://github.com/ProjectEvergreen/greenwood/runs/4903665306?check_suite_focus=true

## Summary of Changes
1. Increased network timeout for `yarn install` on Windows per suggestions so hoping this will help, [1](https://github.community/t/yarn-esockettimedout-on-windows/118776/3), [2](https://github.com/yarnpkg/yarn/issues/8242), [3](https://github.com/yarnpkg/yarn/issues/4890#issuecomment-358179301)

PR #886, opened 2022-02-11
refactor to use octokit to get repos instead of node-fetch

thescientist13:
## Related Issue
N / A

Have been seeing the `init` tests fail intermittently a number of times now and not sure why.  I thought it was a rate limit issue (60 / hr) but even if no tests have been run for a couple days, it will still fail right away.  😵‍💫 
- https://github.com/ProjectEvergreen/greenwood/runs/5149709882?check_suite_focus=true
- https://github.com/ProjectEvergreen/greenwood/runs/5149709927?check_suite_focus=true
- https://github.com/ProjectEvergreen/greenwood/runs/5074715896?check_suite_focus=true
- https://github.com/ProjectEvergreen/greenwood/runs/4811762484?check_suite_focus=true

## Summary of Changes
1. Swap out **node-fetch** with [**octokit** SDK](https://github.com/octokit/octokit.js)

> Not sure if this will actually fix it per se, but will run this PR a few times through GitHub actions and see if it helps

### DIFF
--- a/packages/init/package.json
+++ b/packages/init/package.json
@@ -26,7 +26,7 @@
     "chalk": "^4.1.2",
     "commander": "^8.2.0",
     "inquirer": "^8.2.0",
-    "node-fetch": "^2.6.7",
+    "octokit": "^1.7.1",
     "simple-git": "^2.48.0"
   }
 }
--- a/packages/init/src/index.js
+++ b/packages/init/src/index.js
@@ -16,15 +16,15 @@ import chalk from 'chalk';
 import simpleGit from 'simple-git';
 import commander from 'commander';
 import { copyFolder } from './copy-folder.js';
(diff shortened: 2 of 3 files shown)


thescientist13:
lol, [first run failed](https://github.com/ProjectEvergreen/greenwood/runs/5163755235?check_suite_focus=true) due to #876 😅 😭

hutchgrant (approved):
lgtm

thescientist13:
Interestingly enough, this issue hasn't come up in our GitHub Actions at all recently.  🤔 🤷‍♂️

thescientist13:
So this issue basically hasn't reproduced itself in close to a year now, so I think we can close this one and leave it as is.   As part of #957 , we can migrate to `fetch` though 👍

PR #891, opened 2022-02-17
ensure SPA does not prerender by default

### LINKED ISSUE #890: SPA prerenders by default when no _greenwood.config.js_ is present
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
If no _greenwood.config.js_ is present, SPA applications are now prerendering by default instead.  This appears to be a regression from introducing #874.

## Details
Basically because we always needed a _greenwood.config.js_ to enable `mode: 'spa'`, [the logic for setting `prenreder:false` for a SPA application was only done under that particular scenario](https://github.com/ProjectEvergreen/greenwood/blob/v0.23.0/packages/cli/src/lifecycles/config.js#L209).  But now, there is no mode to set, so no need for a config as the simplest SPA application only needs an _index.html_
```sh
src/
  index.html
```

And so in this case, there is logic to handle that.   Whoops.

thescientist13:
## Related Issue
resolves #890 

## Summary of Changes
1. Handle SPA `prerender` setting when no _greenwood.config.js_ is present

### DIFF
--- a/packages/cli/src/lifecycles/config.js
+++ b/packages/cli/src/lifecycles/config.js
@@ -218,6 +218,11 @@ const readAndMergeConfig = async() => {
             reject(`Error: greenwood.config.js staticRouter must be a boolean; true or false.  Passed value was typeof: ${typeof staticRouter}`);
           }
         }
+      } else {
+        // SPA should not prerender by default
+        if (fs.existsSync(path.join(customConfig.workspace, 'index.html'))) {
+          customConfig.prerender = false;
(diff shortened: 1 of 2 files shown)


thescientist13:
Probably maybe we should increase the test timeouts overall?  Seems Windows always has a little trouble and the when a server based test fails, it usually leave the dev server still running, which causes subsequent tests to usually fail due to the port still being in use.

PR #930, opened 2022-04-09
Feature/issue 878 SSR + custom resources

### LINKED ISSUE #878: support resource plugin based transformations for standard module formats (ex: `import` JSON, CSS) for SSR
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Coming out #831 and the evolution of Greenwood from purely static sites, is the need to be able to support our Resource plugins when running server side code, e.g. NodeJS, and not in the browser, as was the use case when the original API was created.

### Routes
Server side routes right now can only be written in JS.  It would be great if users could author their routes in TS.
```ts
// artists.ts
import { Artist } from '../../services/artist/artist.model.ts';

async function getBody(): String {
  const artists<Artist> = fetch('http://www.example.com/api/artists')
    .then(resp => resp.json());

  return `
     ...
  `
}

export {
  getBody
}
```

## Resource Plugins 
There is a lot of overlap with the above since I think the solution would be the same, but SSR and any form of pre-rendering will fall down is if anywhere in the SSR chain we start using [resource plugins](https://www.greenwoodjs.io/plugins/resource/), or any sort of file transformation we are doing for the browser.
```js
import { css, html, LitElement, unsafeCSS } from 'lit';
import footerCss from './footer.css?type=css';

class FooterComponent extends LitElement {
  constructor() {
    super();

    this.STARTING_YEAR = 2007;
    this.currentYear = new Date().getFullYear(); 
  }

  static get styles() {
    return css`
      ${unsafeCSS(footerCss)}
    `;
  }

  render() {
    const { currentYear, STARTING_YEAR } = this;

    return html`
      <footer class="container as-footer">
        <div class="row">
      
          <div class="col-xs-12">
            <p class="copyright-text">&copy; ${STARTING_YEAR} - ${currentYear} Analog Studios</p>
          </div>
      
        </div>
      </footer>
    `;
  }
}

customElements.define('app-footer', FooterComponent);
```

NodeJS on its own I don't think will have any idea what to do with something like
```js
import footerCss from './footer.css?type=css'
```

## Details
I think that NodeJS supports things like custom loaders (if that is even the right API name) since I believe that is how projects like [**ts-node**](https://www.npmjs.com/package/ts-node) exist I assume?  My hope is that we can just get "in front" of NodeJS as it resolves file and just look at the extension and just apply the right resource at it, pretty much like we do now when serving it to the browser.

_Ultimately, if a user has a plugin for a file extension handled by a resource, Greenwood should use that resource for them on the server side too._

----

Not sure we can start leaning on [CSS or JSON imports]() for some of this now?  🤔

thescientist13:
## Related Issue
resolves #878 

## Summary of Changes
1. [ ] MVP using `--experimental-loader` to support custom resources in SSR mode (only works for `develop` right now)

Screen Shot 2022-04-09 at 6 25 00 PM

## TODO
1. [ ] `yarn build` for _www/_ not finishing?  (related to Rollup?)
1. [ ] How to make it opt-in
1. [ ] double bootstrap console logging
1. [ ] Testing
1. [ ] Documentation
1. [ ] Remove demo code


thescientist13:
resolved in #992

PR #937, opened 2022-05-31
router refactor to better handle hash URL changes and history navigation events

### LINKED ISSUE #936: nested navigation fails to navigate to correct heading/page with `staticRouter`
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
On greenwood site, when I click "Getting Started" page from from the header, then expand "Creating Content" from menu on the left, then select any single nested navigation, I'm forwarded to root `/getting-started/` route and not the correct page.

## Details

Click nested navigation it briefly forwards to right page and then back to the index page.  I noticed this in safari.

thescientist13:
## Related Issue
resolves #936

I believe the issue here is that even [updating a hash will trigger a `pop state` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event#when_popstate_is_sent), even if that means we are navigating "forward", eg.
1. Getting Started
1. Creating Content
1. -> Home Page Template

Opted to refactor the shelf a little bit so it only needs to use `<a>` tags for all navigation, and concentrated all the routing logic into the `staticRouter` to handle these various states of hash vs non hash.  So even without `staticRouter`, Shelf `<a>` + browser default behavior should still work fine.

## Summary of Changes
1. Re-enable `staticRouter` in _greenwood.config.js_ for the website
1. Refactor Shelf to just use standard `<a>` tag behavior for navigating
1. Refactor Router to better accommodate various modes of navigation, with and without hash changes, backwards AND forwards
1. Refactor Router to handle browser's we can support with `staticRouter` and just avoid trying to do any routing at all, so at least it still works instead of failing

### Testing Scenarios
Tested the following scenarios in Chrome and Firefox (currently `staticRouter` is not supported for Safari).  In all scenarios was able to click Back all the way to the starting page.  _Please test forwards and backwards navigation_.  🙏 

#### _Home -> Plugins + side nav_
```
1. Home
2. Plugins
3.   -> Copy
4.   -> Resource
5.   -> Renderer
6.   -> Rollup
```

#### _Home -> Getting Started + side nav + nested items_
```
1. Home
2. Getting Started
3.   -> Project Setup
3a.   -> Key Concepts
4.   -> Creating Content
4a.     -> Home Page Template
4b.     -> Blog Posts Template
4c.     -> App Template
4d.     -> Creating Pages
5.   -> Optimizing
5a.     -> Prerendering
```

#### _Home -> Getting Started + side nav + nested items_ -> Blog
```
1. Home
2. Getting Started
3.   -> Project Setup
3a.   -> Key Concepts
4.   -> Creating Content
4a.     -> Home Page Template
4b.     -> Blog Posts Template
4c.     -> App Template
4d.     -> Creating Pages
5.   -> Optimizing
5a.     -> Prerendering
6.  Blog
7.   -> State of Greenwood 2022
```

#### _Docs -> About + side nav_
```
1. Home
2. Docs
2a.   -> Configuration
2b.   -> Front Matter
3. About
4.   -> Goals
```

## TODO
1. [x] Validate with more testing
    - still an issue with Safari (need to validate post second refactor)
    - _forwards_ navigation testing 🤦 
1. [x] Clean up console logging

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -11,7 +11,7 @@ export default {
   workspace: fileURLToPath(new URL('./www', import.meta.url)),
   prerender: true,
   optimization: 'inline',
-  // staticRouter: true,
+  staticRouter: true,
   interpolateFrontmatter: true,
   plugins: [
     ...greenwoodPluginGraphQL(),
--- a/packages/cli/src/lib/router.js
+++ b/packages/cli/src/lib/router.js
@@ -2,9 +2,10 @@
 
 // ** THIS DOES NOT WORK ON SAFARI **
 // It will just load pages as if staticRouter was not enabled
-window.__greenwood.lastRoutes.push(new URL(window.location));
+// window.__greenwood.lastRoutes.push(new URL(window.location));
 
 document.addEventListener('click', function(e) {
+  console.debug(window.location);
   const href = (e.path && e.path[0]
     ? e.path[0].href // chrome + edge
     : e.originalTarget && e.originalTarget.href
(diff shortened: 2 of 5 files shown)


hutchgrant:
Appears to still be an issue.

thescientist13:
@hutchgrant 
Yeah, you are right, I had a local change that was giving me a false positive.  

Since this bug is rendering the docs site unusable at the moment, I just [disabled `staticRouter` directly on master branch](https://github.com/ProjectEvergreen/greenwood/commits/master) for now and will try and circle back to this tonight to see if I can find a proper resolution.  That way until we do solve this, folks can actually browser the docs.

thescientist13:
@hutchgrant 
Phew, think I got it now.  Updated the description with more info and scenarios I tested for.  All seems to be stable once again.

thescientist13:
OK, I think I got it for real this time.  Noticed a couple glaring issues in what I had so far
- `staticRouter` config was not actually enabled 🤦 
- forwards navigation was not working correctly from what I could test
- the solution seemed overly complex and hard to follow

So I started over and refactored this yet again and tried to leverage `window.location` as much as possible to delegate as much of the non intercepted routing needed to the browser.  This router router to really only handles the partials loading, and then only in that case, will call `history.pushState`.

This has not only made the code entirely less complex, it feels like it works even better now, as backwards and forwards seems to work quite flawlessly, even in Safari, without a lot of weird hacks and custom state tracking. 😌 

Feeling much better about this one, but please do continue testing (ideally in Chrome and Safari), and I have added a new test case in the description.  Please also validate backwards <> forwards navigation too, please.  🙏

PR #942, opened 2022-06-05
Bug/issue 940 static router breaking hybrid workspaces

### LINKED ISSUE #940: `staticRouter` used with a hybrid workspaces is breaks SSR route navigation and serving
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Bug

## Summary
When introducing an SSR route into a hybrid project with static content
```js
// www/pages/artists.js
import fetch from 'node-fetch';

async function getBody() {
  const artists = await fetch('https://www.analogstudios.net/api/artists').then(resp => resp.json());
  const html = artists.map(artist => {
    return `
      <wc-card>
        <h2 slot="title">${artist.name}</h2>
        <img slot="image" src="${artist.imageUrl}" alt="${artist.name}"/>
      </wc-card>
    `;
  }).join('');
  
  return html;
}

async function getFrontmatter(compilation, route) {
  console.debug({ route });
  return {
    menu: 'navigation',
    index: 7
  };
}

export {
  getFrontmatter,
  getBody
};
```

and enabling `staticRouter`, SSR  specific route navigation will break
Screen Shot 2022-06-04 at 2 15 50 PM

Further more, assuming we hit the route directly instead of through the router, the whole page is down!  😱 
Screen Shot 2022-06-04 at 6 57 47 PM

## Details
Ah, issue seems pretty obvious [looking at the code](https://github.com/ProjectEvergreen/greenwood/blob/v0.25.1/packages/cli/src/lib/router.js#L23) but basically not every client side route is a _static_ route.
```js
const route = href.replace(window.location.origin, '');
const routerOutlet = Array.from(document.getElementsByTagName('greenwood-route')).filter(outlet => {
  return outlet.getAttribute('data-route') === route;
})[0];

// what if there is no match though?????  this is the provblem
if (routerOutlet.getAttribute('data-template') === window.__greenwood.currentTemplate) {
  window.__greenwood.lastRoutes.push(window.location.pathname);

  routerOutlet.loadRoute();
} else {
  window.location.href = href;
}
```

So simply, the router should _not_ try and do any special routing for to an SSR page.

----

For the second issue, it seems that the router optimization resource needs to [handle static URLs (end in _.html_) vs SSR URLs, (ends in _/_) before trying to write them out as partials](https://github.com/ProjectEvergreen/greenwood/blob/v0.25.1/packages/cli/src/plugins/resource/plugin-optimization-mpa.js#L76).  I think in general it shouldn't even try and do that for SSR routes?  🤔 

My guess is we need to write a custom test case for this combination of features.

thescientist13:
## Related Issue
resolves #940 

## Summary of Changes
1. Fix route trying load a SSR route "statically"
1. Avoid and add specs to cover this
1. Rename resource plugin to accurately reflect behavior

## TODO
1. [x] should cross verify test with #936 / #937
1. [x] ~~Maybe support SSR routes by writing out a `<greenwood-router>`, e.g. if `isSSR`, load with `window.location.href`?~~ - will track as a future feature request
1. [ ] Should have this thrown a `500` instead of a `404`?

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-static-router.js
+++ b/packages/cli/src/plugins/resource/plugin-static-router.js
@@ -97,7 +97,6 @@ class StaticRouterResource extends ResourceInterface {
           </body>
         `);
 
-        console.debug('SO FAR SO GOOD?????');
         resolve(body);
       } catch (e) {
         reject(e);
--- a/packages/cli/src/plugins/resource/plugin-static-router.js
+++ b/packages/cli/src/plugins/resource/plugin-static-router.js
@@ -10,7 +10,7 @@ import path from 'path';
 import { ResourceInterface } from '../../lib/resource-interface.js';
 import { fileURLToPath, URL } from 'url';
 
-class OptimizationMPAResource extends ResourceInterface {
+class StaticRouterResource extends ResourceInterface {
   constructor(compilation, options) {
     super(compilation, options);
     this.extensions = ['.html'];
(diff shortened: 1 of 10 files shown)


packages/cli/src/plugins/resource/plugin-static-router.js
```diff
@@ -95,6 +98,7 @@ class OptimizationMPAResource extends ResourceInterface {
           </body>
         `);
 
+        console.debug('SO FAR SO GOOD?????');
```
thescientist13, line 101: woops!  need to clean this up 😊

PR #947, opened 2022-06-26
improve easing and rotation of home page banner text

thescientist13:
## Related Issue
N / A

The rotation of the words often encounters transitioning / easing issues, so wanted to try and fix it.

https://user-images.githubusercontent.com/895923/175831974-3bc1b316-2b56-42d9-a325-a8b939d95e74.mov

## Summary of Changes
1. Adjust timeouts to try and smooth out the effect

https://user-images.githubusercontent.com/895923/175831961-8423c12f-0166-4b55-a138-33c79c3cae91.mov

### DIFF
--- a/www/components/banner/banner.css
+++ b/www/components/banner/banner.css
@@ -75,7 +75,7 @@
     }
 
     & span.on {
-      animation: 1s fadeIn linear;
+      animation: 1.5s fadeIn linear;
     }
   }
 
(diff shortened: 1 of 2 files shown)


thescientist13:
Maybe should see if it's possible to do this even more cleanly using something like [requestAnimationFrame](https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame)?

PR #968, opened 2022-08-02
specify styles so production build doesn't break header logo render

thescientist13:
## Related Issue
Blog post `` style breaking header logo in production build (due to inline puppeteer rendering / Shady DOM)
Screen Shot 2022-08-01 at 8 26 56 PM

## Summary of Changes
1. Applied greater CSS specificity to blog post image


hutchgrant (approved):
:tada:

PR #969, opened 2022-08-02
Enhancement/issue 925 auto spread arrays of plugins

### LINKED ISSUE #925: auto spread arrays of plugins
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Bug

## Summary
Some Greenwood plugins return an array of plugins, and so user's have to spread the plugin itself in _greenwood.config.js_.  Example:
```js
import { greenwoodPluginFoo } from 'greenwood-plugin-foo';

export default {
  plugins: [
    ...greenwoodPluginFoo()
  ]
}
```

## Details
Instead, Greenwood easily detect if this value was an array and spread it automatically, so user's wouldn't have to do this themselves, so the equivalent would now look like this
```js
import { greenwoodPluginFoo } from 'greenwood-plugin-foo';

export default {
  plugins: [
    greenwoodPluginFoo()
  ]
}
```

We would also want to make sure to update our README's to remove the `...` from any example docs.

thescientist13:
## Related Issue
resolves #925 

## Summary of Changes
1. Support arrays of plugins without having to use `...`
1. Update README and docs, _greenwood.config.js_ for _www/_
1. Misc related refactoring

### DIFF
--- a/www/pages/blog/release/v0-26-0.md
+++ b/www/pages/blog/release/v0-26-0.md
@@ -187,7 +187,7 @@ So although WCC is now the default for SSR, Puppeteer is still available as a pl
     
     export default {
       plugins: [
-        ...greenwoodPluginRendererPuppeteer()
+        greenwoodPluginRendererPuppeteer()
       ]
     }
     ```
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -35,7 +35,7 @@ export default {
       }
     },
     greenwoodPluginIncludeHTML(),
-    ...greenwoodPluginRendererPuppeteer()
+    greenwoodPluginRendererPuppeteer()
   ],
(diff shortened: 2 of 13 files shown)


www/pages/blog/release/v0-26-0.md
```diff
@@ -187,7 +187,7 @@ So although WCC is now the default for SSR, Puppeteer is still available as a pl
     
     export default {
       plugins: [
-        ...greenwoodPluginRendererPuppeteer()
+        greenwoodPluginRendererPuppeteer()
```
thescientist13, line 190: I guess it makes sense to retro-actively fix this blog post?  Technically this _was_ the API at the time, should it stand as the part of the "record"? 🤔

Issue #971, opened 2022-08-07
refactor bundling lifecycle and resource optimizations


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Our [_rollup.config.js_](https://github.com/ProjectEvergreen/greenwood/blob/v0.26.0/packages/cli/src/config/rollup.config.js) is doing a LOT of work that is potentially being duplicated not only in this file, but later on in our [HTML optimization](https://github.com/ProjectEvergreen/greenwood/blob/v0.26.0/packages/cli/src/plugins/resource/plugin-standard-html.js#L499) step, (where scripts and things are swapped out / updated).  Since the beginning I had felt / wanted to refactor this because it felt wildly inefficient and redundant in a few places, but performed "well enough", but now I _think_ this is a contributing factor to the benchmarking hangups potentially surfacing as part of #970 .

## Details
I think we could simplify some of this a _a lot_.  Things that seems like good opportunities to check for / refactor:
- Capture resources (e.g. `<script>` / `<style>` / `<link>` tag) _once_ and just share that list with everything that needs it
- Reduce repeated looping through HTML files, maybe only doing them in the HTML optimization phase
- Optimize resources in parallel
- See if we can completely elimination needing Rollup specific configs and just rely in our resource plugins entirely - #550 
- See if helps align on #878 (maybe if we are using custom resources, we don't need to provide plugins to Rollup?)

----

Even if it doesn't improve benchmarking that much, I think the refactoring opportunity alone is worth it and getting even a little more decoupled from any one bundler fits within the spirit of the project.

PR #977, opened 2022-09-04
Enhancement/issue 763 parcel css bundling

thescientist13:
## Related Issue

#763 

## Summary of Changes
1. Swapped out PostCSS with ParcelCSS in CLI package
1. Updated PostCSS plugin and deps accordingly 
1. Ensure Style bundling includes interception

As for making our own CSS bundler, we probably could, but maybe more of 2.0 thing?

> _Builds off #971.  See [here](https://github.com/ProjectEvergreen/greenwood/compare/enhancement/issue-971-refactor-bundling-and-optimizations...enhancement/issue-763-parcel-css-bundling?expand=1) for the stacked diff._

## TODO
1. [ ] Update specs due to ParcelCSS "fixing" styles?
    ```sh
    -body{color:red}h1{color:#00f}
    +body{color:red}h1{color:blue}
    ```
1. [ ] Update tech stack docs
1. [ ] More specs failing in GitHub Actions than locally?
1. [ ] Getting a 404 on nested CSS paths when deep link refreshing, e.g. /blog/releases/0.27.0 - seems that only the `bundle` method supports inlining `@import`, [but it requires a file to seed the compiler](https://github.com/parcel-bundler/lightningcss#from-node).  Which means we have to work on scratch files, but that loses the context of relative paths, which is something [PostCSS Import had a solution for](https://github.com/postcss/postcss-import#usage).
    - so there is a custom resolver, but it borks on non-relative URLs?  (like google font URLs)
  <details>
    Screen Shot 2022-09-03 at 7 46 18 PM
  </details>

### DIFF
--- a/greenwood.config.js
+++ b/greenwood.config.js
@@ -15,11 +15,11 @@ export default {
   staticRouter: true,
   interpolateFrontmatter: true,
   plugins: [
-    ...greenwoodPluginGraphQL(),
-    ...greenwoodPluginPolyfills(),
+    greenwoodPluginGraphQL(),
+    greenwoodPluginPolyfills(),
     greenwoodPluginPostCss(),
-    ...greenwoodPluginImportJson(),
-    ...greenwoodPluginImportCss(),
+    greenwoodPluginImportJson(),
+    greenwoodPluginImportCss(),
     {
       type: 'rollup',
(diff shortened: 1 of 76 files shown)


thescientist13:
closing in favor of #980

Issue #951, opened 2022-06-29
Rendering Strategies (`prerender` vs SSR vs CSR)


thescientist13:
## Summary
Coming out of #921 / #879, just wanted to track and make sure we are accurately accounting for the various cases of rendering and how / when to use them, and how [we document them](https://www.greenwoodjs.io/docs/server-rendering/#render-vs-prerender)

Adding support for `prerender` in development would also be supper awesome here.

## Details
At least break them down a bit clearer and when they happen, and what the JS output / delivery is
- SSG - run at build time - CDN
- SSR - run at request time - server, serverless
- CSR - run at run time, like a SPA
- `prerender` - CSR (or SSR) run JS once at build time, e.g. puppeteer (one time render)
  - however, also means "export" for SSR, so will need to disambiguate
  - `export const prerender = true;`
- `export` -> prerender SSR at build time to produce SSG (e.g. `getStaticProps`)
- `csr = false` instead of static optimization, but how in HTML?  (custom `<script setup>` block?)
- move `prerender` configuration to be an optimization  setting?

It should also cover how client side JavaScript plays a role, hydration, eager / lazy loading, etc.

Also, for instance using the GraphQL plugin requires needing to set `prerender` configuration to `true` as well as only working with the puppeteer rendering plugin?  Should validate if this can work with WCC SSR too.

Maybe worth looking into along with #545 

Also, need to handle `globalThis` as per https://github.com/ProjectEvergreen/wcc/issues/100, and possibly refactoring / documenting using `globalThis` instead of `window`, like in the router.

thescientist13:
Documenting as part of the new website - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/117

Issue #923, opened 2022-04-01
import attributes for standard module formats; CSS and JSON 


thescientist13:
## Summary
Related to https://github.com/ProjectEvergreen/greenwood/discussions/606 / https://github.com/ProjectEvergreen/greenwood/issues/878

Currently we have a plugin for being able to `import` CSS, but it is non standard.  Luckily, [there is a standard coming for this in browsers and NodeJS](https://github.com/tc39/proposal-import-assertions)!  Here is an example and [overview](https://web.dev/css-module-scripts/)
```js
import sheet from './styles.css' assert { type: 'css' };

document.adoptedStyleSheets = [sheet];
shadowRoot.adoptedStyleSheets = [sheet];
```

## Details
Basically, what I would love to do is make this out of the box with Greenwood, or at least support it in [**plugin-import-css**](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/plugin-import-css) so that the plugin supports the standard syntax, and eventually, when all browser support CSS modules, user's can just stop using the plugin.

Scenarios that need to be handled:
* NodeJS
  - #878
  - #957 
* Rollup - https://github.com/rollup/rollup/issues/3799
* Acorn - (see above)
* ES Module Shims - https://twitter.com/guybedford/status/1428451066081730561
* Move away from `type=xxx` as it breaks in NodeJS
    ```sh
[...]

Issue #787, opened 2021-11-03
restore Init package `--install` based test cases


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
As part of #777 to introduce the new Init package, one issue that came up was that when running `yarn install` as part of the test suite, warnings were getting thrown as errors.
```
0 passing (3s)
  1 failing

  1) Scaffold Greenwood With Yarn: 
       default minimal template
         "before all" hook in "default minimal template":
     Error: the string "warning @greenwood/cli > markdown-toc > gray-matter > coffee-script@1.12.7: CoffeeScript on NPM has moved to \"coffeescript\" (no hyphen)\n" was thrown, throw an Error :)
      at processTicksAndRejections (node:internal/process/task_queues:93:5)
```

This would cause the test to fail, though there was nothing really wrong at all.  

While using [`ignore` for the stdio does help pass the test](https://github.com/ProjectEvergreen/greenwood/pull/786/files#diff-276757543950e6c5c92261a791b38ace541c72d56a907556529f742066dd778dR102)
```sh
 const process = spawn(pkgCommand, args, { stdio: 'ignore' });
 ```

The result would be that users would not get any install logs at all when running `@greenwood/init` during the `install` step which is confusing and potentially disruptive, making the user think something has broken and potentially cancelling out the whole operation.
```
 % npx @greenwood/init --install
-------------------------------------------------------
Initialize Greenwood Template ♻️
-------------------------------------------------------
Initialzing project with files...
Creating manifest (package.json)...
Installing project dependencies...

# terminal just "hangs" here while install happens 

Initializing new project complete!
owenbuckley@Owens-MBP-2 my-app %
```

## Details
Would like to restore this test case while still keeping the `stdio` setting such that the `install` logs still show.  Not sure if this is something to fix in [**gallinago**](https://github.com/thescientist13/gallinago) or something we can work around in our test setup or not, so making this issue to track and try and resolve.

[...]

PR #992, opened 2022-11-01
Feature/issue 878 ssr custom resources

### LINKED ISSUE #878: support resource plugin based transformations for standard module formats (ex: `import` JSON, CSS) for SSR

thescientist13:
## Related Issue
Resolves #878 (alternative to #930)

## Summary of Changes
1. Initial implantation with support for using `import` with CSS and JSON in NodeJS
1. Setup GitHub Actions testing to accommodate for tests for experimental features
1. Upgrade **gallinago** to support passing args to the child process that handles running Greenwood in tests

## TODO
1. [x] Depends on https://github.com/ProjectEvergreen/greenwood/pull/993
1. [x] Figure out JSON support
1. [x] Add basic test cases
    - ~~CSS~~
    - ~~JSON~~
    - ~~enable in GitHub Actions~~
1. [x] Depends on https://github.com/thescientist13/gallinago/pull/27
1. [x] Get experimental specs passing on Windows (find equivalent of `$(pwd)` in Windows)~~
    -  init test cases will get handled here - https://github.com/ProjectEvergreen/greenwood/issues/787#issuecomment-1320950877
1. [x] Validate other import plugins - found some issues, so created https://github.com/ProjectEvergreen/greenwood/issues/1004
    - GraphQL (_.gql_)
    - TypeScript (_.ts_)
1. [x] Test all _greenwood.config.js_ plugins for SSR support (may be WCC related)
    - `staticRouter` - breaks on `window`
    - `greenwoodPluginPolyfills` - breaks on `window`
    - tracked in https://github.com/ProjectEvergreen/wcc/issues/100
    - added to https://github.com/ProjectEvergreen/greenwood/issues/951
1. [x] Explore role / overlap of import assertions - will defer to https://github.com/ProjectEvergreen/greenwood/issues/923
1. [x] figure out if `?type=json` is still useful / needed - breaks in SSR, so should try and cutover to import assertions - #923 
1. [x] research [`shortCircuit`](https://nodejs.org/docs/latest-v16.x/api/esm.html#hooks) to validate correct settings / usage
    - seems fine based on [the code examples](https://nodejs.org/docs/latest-v16.x/api/esm.html#resolvespecifier-context-nextresolve) in the linked docs
    - also matches this project's implementation - https://github.com/DataDog/import-in-the-middle
1. [x] reconcile minimum node version of `16.x` (and `18.x`?) for project + GitHub Actions - standalone PR
1. [x] restore _package.json_ `build` and `develop` commands
1. [x] Add documentation
    - Minimum Node version
    - Experimental


.nvmrc
```diff
@@ -1 +1 @@
-14.17.0
+16.15.0
```
thescientist13, line 1: maybe this should actually be `16.17.0` to align with loaders API

www/pages/docs/server-rendering.md
```diff
@@ -203,6 +203,26 @@ async function getTemplate(compilation, route) {
 }
 ```
 
+### Custom Imports (Experimental)
```
thescientist13, line 206: Should move the _Experimental_ to its own line so deep links don't break when / if the _experimental_ label gets removed.

Issue #948, opened 2022-06-29
Standardize on Web APIs (`Request` / `Response` / `URL` / etc)


thescientist13:
### Discussed in https://github.com/ProjectEvergreen/greenwood/discussions/867

<div type='discussions-op-text'>

<sup>Originally posted by **thescientist13** January 13, 2022</sup>
I would have to do a little more research into Koa, but the [**Fastify**](https://www.fastify.io/) project is looking super interesting and is actively lead / maintained by one the TSC members of NodeJS.  As Greenwood moves towards #708 / #576 , it would be good to evaluate the server solution we are using, as it was originally only considered under the basis of being a _dev_ server, is now going to be put into production situations, and so performance and speed will play a more important role in user experience.

In addition, we should really standardize our [`Request`](https://developer.mozilla.org/en-US/docs/Web/API/Request) / [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) objects.</div>

Should also evaluate native [`fetch` from Node](https://blog.logrocket.com/fetch-api-node-js/) and also keep an eye on [WINTER CG](https://wintercg.org/).

----

So this change would definitely impact plugin resources the most (not necessarily in a bad way) and coming out of #974 , it became clear that `ResourceInterface` may need to be deprecated:
- The assumption of having default implementations on `ResourceInterface` like `shouldResolve` while convenient, has cause [hard to trace issues where plugins were running in lifecycles automatically](https://github.com/ProjectEvergreen/greenwood/pull/974/files#diff-e1c21a418290cc7cfabe43827e44d70e31dc684116aa943bdbd49410e8e1abb3R16) and thus manipulating the req / res unknowingly.  You can see in this PR where i had to explicitly opt-out.  In actuality, it was more of an Abstract Class rather than an interface, so it was even badly named.  I think I would prefer to just get rid of it, but I think `contentType` and `extension` could still be useful maybe?
- With #930 , this plugin will extend past just the browser.  So resource plugins are just generic handlers at that point, environment agnostic.
- Should better explain the nuance of changing the `Request` vs the `Response` and what the order of operations are done by Greenwood when taking the response from resource plugin and setting up the next resource plugin so it can operate on it.  Should it be assumed that `Request` params are immutable, and `Response` objects are mutable?

----

Basically, we should be doing more of:
1. [ ] Leverage `URL` top to bottom internally (instead of `path`)
1. [ ] Use `Request` / `Response` everywhere, at very minimum for plugins / resources
[...]

Issue #946, opened 2022-06-26
bundle and optimize server and API routes and serve from output directory (decouple `serve` task from source code)


thescientist13:
## Type of Change
Feature

## Summary
As it stands right now (I believe) SSR routes are called directly from _src/_, but ideally they should bundle and resolved to JS files by the hybrid server so that running the final result and the CLI can be totally independent of the source code needing to be present.

## Details
This should ensure that the CLI can be used independent of the source code, say if the server crashes or someone wants to write their own server and just build a static distribution through something like Docker.  This means that if `greenwood serve` is called and there is no _workspace_ directory, it should try and find the output directory and just start from that, assuming a pre-generated bundle.  For example, in Next, if you clone a repo and just run `next serve` it will break until you at least run a `next build`.

Most importantly, this avoids false positives thinking something is working when using `serve` because it is hitting source code directly, not bundled output.  Hit that again in https://github.com/ProjectEvergreen/greenwood/issues/986.  Blast you false positives!  😠 

---

This should also probably include API routes too after #1007 lands, and then will need to figure out how we change the shape of _graph.json_, which will have an impact on our graph output.  Maybe something like
```js
const graph = {
  pages: [ /*... */ ],
  apis: [ /*... */ ]
}
```

Additionally, if there are no _pages/_ but an _/api/_ directory, should we even output a default _index.html_ / _404.html_?

PR #1017, opened 2022-12-03
Feature/issue 1007 api routes

### LINKED ISSUE #1007: API Routes
## Overview

Pretty straightforward, but Greenwood should support standalone API routes, along side pages.  So for example
```
src/
  api/
    greeting.js
```

Would be reachable at `/api/greeting`

## Details
Some important details to keep in mind
- Should follow Request / Response model 
- Should play nicely with adapters / server less
- Should be streaming compatible

thescientist13:
## Related Issue
resolves #1007 

## Summary of Changes
1. Introduce support for _src/api/*_ based routes that confirm to a `Request` / `Response` contract (like will also happen for #948)
1. Add test cases and created "API Routes" docs page

## TODO
1. [x] Doesn't change on refresh - ESM caching issue I believe :/ - commented in https://github.com/nodejs/modules/issues/307#issuecomment-1336020135 to show our support for "native" ESM cache busting
1. [x] Validate `build` / `serve` commands
1. [x] Flesh out entire `Request` / `Response` objects
1. [x] Port is hardcoded
1. [x] response assumes `.json()`
1. [x] stop using `path` / url helpers!
1. [x] Testing
1. [x] Documentation
1. [x] new API serve specs are running, but not finishing :/ 
1. [x] Remove demo code

> **Note**:  more of the `Request`/ `Response` will get fleshed out as part of #948

## Questions
1. [x] Should this be _src/api/*_ or _src/pages/api/*_ (like in Next) - will start with _src/pages_ for now
1. [x] Should we add APIs to the graph and / our output some stats?  That will definitely blow up our GraphQL schema as is, so might need to be its own task - tracked in https://github.com/ProjectEvergreen/greenwood/issues/946
1. [x] What if there is just an _/src/api/*_ directory and no _/src/pages/*_, should we not emit default _index.html_ / _404.html_?  (e.g. for an entirely serverless stack) - tracked in https://github.com/ProjectEvergreen/greenwood/issues/946


packages/cli/src/plugins/resource/plugin-api-routes.js
```diff
@@ -0,0 +1,46 @@
...
+    // TODO check it exists first
+    return url.startsWith('/api');
+  }
+
+  async serve(url) {
+    let href = new URL(`${this.getBareUrlPath(url).replace('/api/', '')}.js`, `file://${this.compilation.context.apisDir}`).href;
+
+    // https://github.com/nodejs/modules/issues/307#issuecomment-1165387383
+    if (process.env.__GWD_COMMAND__) { // eslint-disable-line no-underscore-dangle
+      href = `${resolvedUrl}?t=${Date.now()}`;
```
thescientist13, line 23: `resolvedUrl` shouid be `href`, otherwise getting an undefined error

packages/cli/src/plugins/resource/plugin-api-routes.js
```diff
@@ -0,0 +1,46 @@
...
+  async shouldServe(url) {
+    // TODO check it exists first
+    return url.startsWith('/api');
+  }
+
+  async serve(url) {
+    let href = new URL(`${this.getBareUrlPath(url).replace('/api/', '')}.js`, `file://${this.compilation.context.apisDir}`).href;
+
+    // https://github.com/nodejs/modules/issues/307#issuecomment-1165387383
+    if (process.env.__GWD_COMMAND__) { // eslint-disable-line no-underscore-dangle
```
thescientist13, line 22: should make sure to check for the actual value, e.g.
```js
if (process.env.__GWD_COMMAND__ === 'develop') {
  ...
}
```

PR #1020, opened 2022-12-04
handle recursively copying directories for copy plugin

### LINKED ISSUE #1019: nested directories breaking copy plugin
## Details
The copy plugin allow either copying a file or directory, however, if that directory path is deeply nested, then Greenwood will throw an error.  For example, trying to copy PrismJS theme files using something like this
```js
{
  type: 'copy',
  name: 'plugin-copy-prismjs',
  provider: (compilation) => {
    const prismThemesPath = 'node_modules/prismjs/themes';
    const fromThemesDir = path.join(context.projectDirectory, prismThemesPath);
    const toThemesDir = path.join(context.outputDir, prismThemesPath)

    return [{
      from: fromThemesDir,
      to: toThemesDir
    }];
  }
}
```

## Details
The solution is to just make `fs.mkdirSync` recursive in the copy lifecycle.  Same thing can be done from userland for now.
```js
fs.mkdirSync(toThemesDir, {
  recursive: true
});
```

thescientist13:
## Related Issue
resolves #1019 

## Summary of Changes
1. Handle deeply nested directories when copying

## TODO
1. [ ] Add test case

### DIFF
--- a/packages/cli/src/lifecycles/copy.js
+++ b/packages/cli/src/lifecycles/copy.js
@@ -3,13 +3,13 @@ import path from 'path';
 
 async function rreaddir (dir, allFiles = []) {
   const files = (await fs.promises.readdir(dir)).map(f => path.join(dir, f));
-  
+
   allFiles.push(...files);
-  
+
   await Promise.all(files.map(async f => (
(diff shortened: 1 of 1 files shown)


thescientist13:
moved to #1021 where I get the issue number in the branch name correct 🙃

Issue #1048, opened 2023-02-04
handle merging additional `Request` / `Response` instance properties


thescientist13:
## Summary
As part of #948, we created a utility in [_resource-utils.js_](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lib/resource-utils.js) to merge `Response` objects when needed, but it doesn't currently support anything other `body` and `headers`.
```js
return new Response(source.body, {
  headers
});
```

In addition, instances of `Request` are missing critical properties like `body`, which means Greenwood is totally unusable for handling [`POST` requests](https://github.com/thescientist13/greenwood-htmx/pull/3)
https://github.com/ProjectEvergreen/greenwood/blob/v0.28.5/packages/cli/src/lifecycles/serve.js#L37
```js
  const initRequest = new Request(url, {
    method: ctx.request.method,
    headers: new Headers(ctx.request.header)
  });
```

## Details
I would like to make sure that we can merge all relevant [`Response` instance properties](https://developer.mozilla.org/en-US/docs/Web/API/Response#instance_properties) like 
- `status`
- `statusText`
- Custom headers like `Content-Length`

And [`Request` instance properties](https://developer.mozilla.org/en-US/docs/Web/API/Request/body)
- `method`
- `body` (make sure it supports `json`, `formData`, etc)
- integrity
- referrer
- referrerPolicy
- all of them?

Needs to be done in
- the [_serve.js_ lifecycle](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/serve.js) as needed.
- API routes for local dev - https://github.com/ProjectEvergreen/greenwood/pull/1105
- Adapter Plugins
---

> Bonus points for handling the [`Range` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) (good link [here on implementation](https://blog.logrocket.com/streaming-video-in-safari/)) which I think will fix this issue with serving media on Safari
> ```sh
> Error: read ECONNRESET
> at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
> ```
![Screen Shot 2023-03-03 at 9 09 28 PM](https://user-images.githubusercontent.com/895923/222870290-236b7559-75ab-40cc-b11a-173841961322.png)

thescientist13:
See #1146 and #1147 for follows up on this work

PR #1054, opened 2023-02-05
add support for standard `<video>` and `<audio> `formats

### LINKED ISSUE #1038: Support web standard `<audio>` and  `<video>` formats
## Overview
Completely missed that currently there isn't any support for loading resources that could be used in [`<video>`](https://www.w3schools.com/tags/tag_video.asp) or [`<audio>`](https://www.w3schools.com/tags/tag_audio.asp) tags like an _.mp4_ or _.mp3_!  Should probably get on that...

## Details
Would just need to follow pattern of our standard font or images plugins and create two new plugins, one for each format type
1. [ ] Audio - https://www.thoughtco.com/audio-file-mime-types-3469485
1. [ ] Video - https://help.encoding.com/knowledge-base/article/correct-mime-types-for-serving-video-files/
1. [ ] Add a test case for at least one format per type (_.mp3_ and _.mp4_ respectively)

> Would likely want to get this in ASAP after #948 lands since the API for resource plugins is changing a bit

thescientist13:
## Related Issue
resolves #1038 

## Summary of Changes
1. Add support for standard `<video>` and `<audio>` formats
1. Added one-off support for `Content-Length` header which was needed by Safari (will make more robust as part #1048)
1. Found a missing case of needing to "forward" request headers for the `serve` command

## TODO
1. [x] Add support for standard `<audio>` formats
1. [x] Handle conflicts with "competing" extensions like _.ts_ (TypeScript and  Apple [Live Streaming](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/StreamingMediaGuide/DeployingHTTPLiveStreaming/DeployingHTTPLiveStreaming.html)) or _.wav_ and [OGG, since it could be both](https://developer.mozilla.org/en-US/docs/Web/HTTP/Configuring_servers_for_Ogg_media) .  Right now we [`break` on first come, first serve](https://github.com/ProjectEvergreen/greenwood/blob/release/0.28.0/packages/cli/src/lifecycles/serve.js#L69) 🤔 
    - set and filter on request `Accept` header?
    - [Handle OGG format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Configuring_servers_for_Ogg_media)
1. [x] ~~How to best handle Safari for generic media container formats?~~ - solved through above solution
    -   `?type=audio|video`?
1. [ ] Create discussion for  solving the overlapping extensions issues (e.g. _.ts_) 
https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/CreatingVideoforSafarioniPhone/CreatingVideoforSafarioniPhone.html#//apple_ref/doc/uid/TP40006514-SW6
1. [x] ~~Documentation / callout about asset handling on Safari~~ - N / A
1. [x] Although it plays fine, getting this error on the NodeJS side when loading any media file in Safari - deferred to https://github.com/ProjectEvergreen/greenwood/issues/1048
    ```sh
    Error: read ECONNRESET
    at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
    ```


.gitattributes
```diff
@@ -13,4 +13,11 @@
 *.woff    binary
 *.woff2   binary
 *.ttf     binary
-*.eot     binary
\ No newline at end of file
+*.eot     binary
+*.flv     binary
+*.mp4     binary
+*.ts      binary
```
thescientist13, line 19: hmm, will this convert typescript files too? 🤔

packages/cli/src/plugins/resource/plugin-standard-video.js
```diff
@@ -0,0 +1,71 @@
...
+      case 'wmv':
+        contentType = 'video/x-ms-wmv';
+        break;
+      default:
+
+    }
+
+    return new Response(body, {
+      headers: new Headers({
+        'Content-Type': `image/${contentType}`
```
thescientist13, line 59: Whoa... this is really wrong 😬

Wonder how it's still working though?? 🤨

thescientist13:
So maybe I have been over thinking this or misinterpreting some of the data I read, but it seems like _.mp4_ and _.3pg_ are both video formats, so maybe we don't need any of this generic container format handling.

This would just leave _.ogg_ then, which does seem to be a container format and seems like we can just use a generic content type instead of trying to "sniff" it out
https://developer.mozilla.org/en-US/docs/Web/HTTP/Configuring_servers_for_Ogg_media
```
AddType audio/ogg .oga
AddType video/ogg .ogv
AddType application/ogg .ogg
```

That said, the `Sec-Fetch-Dest` header was still an interesting find, but maybe a too prescriptive a solution.

PR #1061, opened 2023-02-09
correctly order Apollo data `<script>` tag inside `<head>` tag

### LINKED ISSUE #1053: GraphQL plugin data `<script>` tag is being inserted before the opening `<head>` tag
## Type of Change
Bug

## Summary
When using the GraphQL plugin, the `<script>` tag placeholder for serialized data is getting inserted _before_ the opening `<head>` tag, instead of after.
```html
  <script data-state="apollo" >
    window.__APOLLO_STATE__ = true;
  </script>
<head>
```

## Details
This one looks pretty obvious by [looking at the code](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/src/index.js#L67) 🤦‍♂️ 
```js
body = body.replace('<head>', `
  <script data-state="apollo" data-gwd-opt="none">
    window.__APOLLO_STATE__ = true;
  </script>
  <head>
`);
```

The `<head>` tag should be first in the `replace` string.

Yasir761:
## Related Issue

resolves #1053

## Summary of Changes
1. Next Apollo data `<script>` block inside `<head>` tag
1. Update specs to ensure this relationship

### DIFF
--- a/packages/plugin-graphql/src/index.js
+++ b/packages/plugin-graphql/src/index.js
@@ -64,13 +64,13 @@ class GraphQLResource extends ResourceInterface {
   async optimize(url, body) {
     return new Promise((resolve, reject) => {
       try {
-        body = body.replace( `
+        body = body.replace('<head>', `
         
+        <head>
           <script data-state="apollo" data-gwd-opt="none">
             window.__APOLLO_STATE__ = true;
           </script>
-          </head>
-        `,'<head>');
+        `);
     
         resolve(body);
       } catch (e) {
--- a/packages/plugin-graphql/src/index.js
+++ b/packages/plugin-graphql/src/index.js
@@ -65,7 +65,6 @@ class GraphQLResource extends ResourceInterface {
     return new Promise((resolve, reject) => {
       try {
         body = body.replace('<head>', `
-        
         <head>
           <script data-state="apollo" data-gwd-opt="none">
             window.__APOLLO_STATE__ = true;
--- a/packages/plugin-graphql/src/index.js
+++ b/packages/plugin-graphql/src/index.js
@@ -64,12 +64,13 @@ class GraphQLResource extends ResourceInterface {
   async optimize(url, body) {
     return new Promise((resolve, reject) => {
       try {
-        body = body.replace('<head>', `
+        body = body.replace( `
+        
           <script data-state="apollo" data-gwd-opt="none">
             window.__APOLLO_STATE__ = true;
           </script>
-          <head>
-        `);
+          </head>
+        `,'<head>');
     
         resolve(body);
       } catch (e) {
--- a/packages/plugin-graphql/test/cases/query-config/query-config.spec.js
+++ b/packages/plugin-graphql/test/cases/query-config/query-config.spec.js
@@ -82,7 +82,7 @@ describe('Build Greenwood With: ', function() {
       });
 
       it('should have one window.__APOLLO_STATE__ <script> with (approximated) expected state', function() {
-        const scriptTags = dom.window.document.querySelectorAll('script');
+        const scriptTags = dom.window.document.querySelectorAll('head script');
         const apolloScriptTags = Array.prototype.slice.call(scriptTags).filter(script => {
           return script.getAttribute('data-state') === 'apollo';
         });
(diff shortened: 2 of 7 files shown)


packages/plugin-graphql/src/index.js
```diff
@@ -64,12 +64,13 @@ class GraphQLResource extends ResourceInterface {
   async optimize(url, body) {
     return new Promise((resolve, reject) => {
       try {
-        body = body.replace('<head>', `
+        body = body.replace( `
+
```
thescientist13, line 68: Hey @Yasir761

What would need here is to just [move the _position_ of `<head>` tag](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/src/index.js#L71) to come _before_ the `<script>` block in the replace function.  From what I see here is looking to replace the `<script>` block _with just_ `<head>`.

I think if you run `yarn test` locally it would likely fail the specs.  I think we can even ensure this change by slightly tweaking [all instances of this kind of spec in the package](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/test/cases/query-config/query-config.spec.js#L84)
```js
// instead of
const scriptTags = dom.window.document.querySelectorAll('script');

// we should ensure the nesting by doing this to ensure it is within the <head> tag
const scriptTags = dom.window.document.querySelectorAll('head script');
```

Let me know if that all makes sense and here are some [docs](https://github.com/ProjectEvergreen/greenwood/blob/master/.github/CONTRIBUTING.md#testing) on how to get into testing for this project.

thescientist13 (commented):
@Yasir761 
Do you mind filling out the PR template too please?  Thanks!

Yasir761:
Can I close this pull request and open a new PR with changes.

packages/plugin-graphql/src/index.js
```diff
@@ -65,10 +65,11 @@ class GraphQLResource extends ResourceInterface {
     return new Promise((resolve, reject) => {
       try {
         body = body.replace('<head>', `
+        
+        <head>
```
thescientist13, line 69: Yeah, this looks good!  My only nitpick would be removing the extra line break but other wise this is desired approach for this fix.  👍

packages/plugin-graphql/test/cases/query-config/query-config.spec.js
```diff
@@ -82,7 +82,7 @@ describe('Build Greenwood With: ', function() {
       });
 
       it('should have one window.__APOLLO_STATE__ <script> with (approximated) expected state', function() {
-        const scriptTags = dom.window.document.querySelectorAll('script');
+        const scriptTags = dom.window.document.querySelectorAll('head script');
```
thescientist13, line 85: This looks good!  If you look in the rest of the test cases for this package, I think you will find a few more test cases with this exact same code, so if you do a search for something file _"should have one window.__APOLLO_STATE__ <script>"_ you should easily be able to find them and the adjust those too.  👍

thescientist13 (commented):
@Yasir761 I think this is looking good so far, so I don't think there's any need to close this or start a new PR.  It's fine to just iterate as you go but if you would prefer we could put the PR in draft mode?
Screen Shot 2023-02-14 at 8 50 43 PM

Like I said, this is looking good far, just a couple minor follow up comments.

packages/plugin-graphql/test/cases/query-config/query-config.spec.js
Yasir761: Hey, I'm not being able to find more such files where I can make changes can you please help out, so I can make change

thescientist13: Yeah, If you look in the other [test case files in the **plugin-graphql** package](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/plugin-graphql/test/cases) you will find other instances of the kind of test you see
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/test/cases/query-children/query-children.spec.js#L155
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/test/cases/query-custom-frontmatter/query-custom-frontmatter.spec.js#L157
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/test/cases/query-custom-schema/query-custom-schema.spec.js#L75
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/test/cases/query-graph/query-graph.spec.js#L153
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-graphql/test/cases/query-menu/query-menu.spec.js#L155

thescientist13 (approved):
Thank you so much, looks good!

PR #1063, opened 2023-02-11
Deleted unwanted card.native.js file

### LINKED ISSUE #1060: remove extra component file from website
## Type of Change
Other

## Summary
Looks like a file was left over from testing a feature but I think it can be removed.
https://github.com/ProjectEvergreen/greenwood/blob/master/www/components/card/card.native.js

## Details
Should just be able to delete it.

loki1205:
## Related Issue
Resolves #1060 

## Summary of Changes

1. [x] Delete card.native.js


thescientist13 (approved):
Thanks!  Will get this merged soon.

PR #1066, opened 2023-02-15
Enhancement/restore rollup local workspace resolution

thescientist13:
## Related Issue
N / A

Noticed some odd paths coming out some projects while upgrading them to **v0.28.0-alpha.1**.  Nothing was broken but I also noticed compared to the [original logic](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/config/rollup.config.js#L13), it was also doing bit more than it should be.

## Summary of Changes
1. Restore Rollup local workspace resolution

## TODO
1. [x] Add / update test case for ESM path that doesn't start with a `.` (Per ESM, imports can not start with a `/`)


thescientist13 (commented):
In light of #1070 , just wanted to capture that after a rebase, all the specs are at least passing locally
```sh
  2582 passing (4m)
  28 pending

PR #1076, opened 2023-02-25
refactor copy lifecycle to use leverage fs.copyFile instead of streams

thescientist13:
## Related Issue
N / A

## Summary of Changes
1. Use `fs.copyFile` instead of complicated streams

## TODO
1. [ ] Can also use `fs.copyFile` in [this spec too](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/develop.default/develop.default.spec.js#L61)


thescientist13 (commented):
Looks like I could also use `fs.copyFile` in [this spec too](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/develop.default/develop.default.spec.js#L61)

PR #1113, opened 2023-05-19
bug/issue 1112 upgrade apollo/client and update entry point specifier path

### LINKED ISSUE #1112: plugin-graphql importing apollo/client file that no longer exists in recent releases
## Type of Change
Bug

## Summary

When running `greenwood develop` and using `@greenwood/plugin-graphql` in my greenwood.config.js I see:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/skynet/workspace/greenwood/rubiks/node_modules/@apollo/client/core/core.cjs.js' imported from /home/skynet/workspace/greenwood/rubiks/node_modules/@greenwood/plugin-graphql/src/core/cache.js
```

## Details

@greenwood/plugin-graphql is importing from `apollo/client/core/core.cjs.js` which no longer exists in more recent releases. Instead the file `apollo/client/core/core.cjs` exists without the `.js` extension.

 The yarn.lock and package.json should be updated with the most recent release of @apollo/client as well

hutchgrant:
## Related Issue
Resolves #1112

## Summary of Changes

* fixed import of `apollo/client/core/core.cjs`
* updated dependency for @apollo/client

### DIFF
--- a/packages/plugin-graphql/src/core/cache.js
+++ b/packages/plugin-graphql/src/core/cache.js
@@ -1,3 +1,5 @@
+// we "deep" link into the Apollo package to avoid pulling in React
+// https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#using-apollo-client-without-react
 import { ApolloClient, HttpLink, InMemoryCache } from '@apollo/client/core/index.js';
 import { checkResourceExists } from '@greenwood/cli/src/lib/resource-utils.js';
 import fs from 'fs/promises';
--- a/packages/plugin-graphql/src/core/cache.js
+++ b/packages/plugin-graphql/src/core/cache.js
@@ -1,4 +1,4 @@
-import ApolloCore from '@apollo/client/core/core.cjs.js';
+import ApolloCore from '@apollo/client/core/core.cjs';
 import { checkResourceExists } from '@greenwood/cli/src/lib/resource-utils.js';
 import fs from 'fs/promises';
 import { gql } from 'apollo-server';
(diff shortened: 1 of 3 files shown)


thescientist13 (approved):
Nice, this looks good!  I'm going to crawl through the Apollo client release neats to see if I can track this source of this down if just for posterity.

Should be able to get this out a into a release over the weekend.

packages/plugin-graphql/src/core/cache.js
```diff
@@ -1,4 +1,4 @@
-import ApolloCore from '@apollo/client/core/core.cjs.js';
+import ApolloCore from '@apollo/client/core/core.cjs';
```
thescientist13, line 1: I found [this thread](https://github.com/apollographql/apollo-client/issues/10319) and [this thread](https://github.com/apollographql/apollo-client/issues/9925#issuecomment-1191045735) which got me looking at the package contents and and maybe there's an alternative we can't, which happens to be a nice ESM export?  🤩 
https://unpkg.com/browse/@apollo/client@3.7.14/core/index.js

```js
// thinking it would be nice if one of these could work
import { ApolloCore } from '@apollo/client';

// or
import { ApolloCore } from '@apollo/client/core/index';
```

Not sure if it's new or a named or default export but might be worth a try.  I can quickly test that over the weekend as well.

thescientist13 (approved):
Cool, so that new ESM entry point seems to work.

As for why this breaking change, based on going through the changelogs, there were quite a few updates to their packaging output, and seems that they have made some changes to better support ESM vs CJS, and in doing so, may have made some "fixes"?  This one seems to touch upon the issue somewhat
https://github.com/apollographql/apollo-client/issues/8592

Either way, the entry point we're using now looks much better now and maybe even a bit of a smaller bundle size now that are we using named exports.

Thanks for the report and fix!

Issue #273, opened 2020-01-17
resolvers and file writing assumes certain route depth


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [x] Other (please clarify below)

## Summary
Technical debt tracking for some of the issue rough and dirty technical design decisions made as part of implementing #115 .

## Details
As parr of the graph querying in its resolvers, and the writing / reading of _cache.json_ files, there are hardcoded uses of things like `split('/')[0]`.  This should be better supported to support any depth of page directories, or made more explicit in some way through the docs / code.

examples:
- [_data/client.js_](https://github.com/ProjectEvergreen/greenwood/blob/release/0.5.0/packages/cli/src/data/client.js#L19)
- [_data/schema/graph.js_](https://github.com/ProjectEvergreen/greenwood/blob/release/0.5.0/packages/cli/src/data/schema/graph.js)
- [_lifecycles/serialize.js_](https://github.com/ProjectEvergreen/greenwood/blob/release/0.5.0/packages/cli/src/lifecycles/serialize.js#L16)
- other instances

Additionally, can this handle deeply nested directories to generate children?

PR #1135, opened 2023-08-05
Feature/issue 1116 base path configuration

### LINKED ISSUE #1116: add support for configuring a base path
## Type of Change
Feature

## Summary
It is not uncommon for websites to be deployed at an "offset" URL from the root domain.  For instance `www.example.com/xyz`.  Most notably, [GitHub Pages](https://pages.github.com/) does this for project based site, as the full URL is a concatenation of the github.com domain and a path segment based on the repository name, and so `<repository>` is effectively the web root in this case.
```sh
http://<username>.github.io/<repository>
```

> _**Note**: by virtue of this feature request, it does require acknowledging that your current guide on using GitHub Pages is both out of date, and thus broken.  Updating that guide should come immediately after introducing this feature._

Thus, it would require all assets to be relative that path segment, e.g.
- `/<repository>/header.js`
- `/<repository>/assers/logo.png`
- etc

## Details
This should be doable but would require a few different parts of the system working together
1. [ ] servers (local and prod) would need to account for this public path when resolving file paths to the filesystem
1. [ ] Will need to handle updating `<link>`, `<script>`, etc tags accordingly when bundling
1. [ ] Others I'm sure 😅 

I wonder if this is something that setting a [`<base href="...">` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) could help with?

thescientist13:
## Related Issue
resolves #1116 (and #273)

## Summary of Changes
1. Add `basePath` support to _greenwood.config.js_ (covers pages, API routes, and adapter routing, `<script>` and `<link>` tags, and GraphQL queries)
1. Persist base path setting in a `<script>` tag
1. Add **Base Path** configuration docs
1. Update **GitHub Pages** guide to call out when to use a base path 

## TODO
1. [x] Enable for `greenwood build` / `greenwood serve`
1. [x] Support  script and style tags
1. [x] Check support for `/graph.json`
    Screenshot 2023-10-23 at 8 57 47 PM
1. [x] Check support for static router
    Screenshot 2023-10-23 at 8 57 38 PM
1. [x] Check support for 404 page
1. [x] Odd `margin: auto` and navigation issues on the docs website
    Screenshot 2023-10-25 at 4 26 13 PM
1. [x] Should this be baked into the graph / URLs by default?
1. [x] Add test cases
    - ~~develop~~
    - ~~serve~~
    - ~~graphql~~
    - ~~static router~~
1. [x] finish and add test cases for - https://github.com/ProjectEvergreen/greenwood/issues/273
1. [x] Check support for dev proxy
1. [x] Check support for API routes / SSR pages
1. [x] Check support in adapters (for host routing)
1. [x] Display `basePath` for server logging output
1. [x] Add documentation
    - clarify this is only for greenwood routing (not for user content like links, forms, etc)
1. [x] test with a project using GitHub pages and update GitHub pages guide - https://github.com/ProjectEvergreen/greenwood-demo-github-pages/issues/1
1. [x] "lift" shared client side base path config to CLI output and document - https://github.com/thescientist13/greenwood-starter-presentation/pull/66
1. [x] Remove demo code
1. [x] Refactor tests to align with #1172 

## Questions
1. [x] Is there actually a way to do it with _just_ `<base href="..."/>` (make discussion) - https://github.com/ProjectEvergreen/greenwood/discussions/1175
1. [x] should track an issue for customizing trailing forward slash - https://github.com/ProjectEvergreen/greenwood/issues/new


www/templates/app.html
```diff
@@ -4,6 +4,7 @@
   <head>
     <title>Greenwood - ${globalThis.page.title}</title>
     <meta charset="utf-8">
+    <base href="/my-subpath/"/>
```
thescientist13, line 7: note: this isn't actually doing anything

packages/cli/src/lib/templating-utils.js
```diff
@@ -157,6 +157,11 @@ async function getAppTemplate(pageTemplateContents, context, customImports = [],
       ? appBody.replace(/<page-outlet><\/page-outlet>/, pageBody)
       : appBody;
 
+    const basePathTag = basePath === ''
```
thescientist13, line 160: note: this isn't actually doing anything

Issue #955, opened 2022-07-01
Layouts (Templates) and Pages


thescientist13:
## Type of Change
RFC

## Summary
Worth thinking about reviewing any DX considerations for Templates and Pages.

## Details
Things that come to mind:
1. [ ] Templates as custom elements
1. [ ] What to do about `getTemplate` and `getBody` if anything, now that we `HTMLElement` SSR support. Does it make sense to have multiple ways to SSR a page?
1. [ ] what to do about current API signature (`(compilation, page, request)`) - Anyway to clean this up?
1. [ ] rename `getBody` -> `getPage`?
1. [ ] rename `getFrontmatter` -> `getMetadata`?
1. [ ] Should templates be renamed to layouts?  Seems more consistent.
1. [ ] Evaluate frontmatter and `<meta>` (a la React helmet)
1. [ ] Per #1088, templates are currently static only for SSR pages, but would be nice to change that - https://github.com/ProjectEvergreen/greenwood/pull/1110/files#diff-036193aa52a1863fc84a1878a4026f259d3605cdc9c0312d3dba861b80b7bb35R199
1. [ ] Use `<slot>` instead of `<content-outlet>`?  Or [maybe not](https://twitter.com/jon_neal/status/1555525473773436932)?
1. [ ] What about custom page extensions, like _.ts_ or _.jsx_?
1. [ ] bundle context plugins for SSR
1. [ ] Page level data - in replace of removing getPages / getTemplate along with #952 
1. [ ] Trailing forward slash (`/`) 
1. [ ] APIs in the _pages/_ directory?
1. [ ] refactor Greenwood context, move all things into _manifest.json_

thescientist13:
Might be relevant - https://www.reddit.com/r/reactjs/comments/110biur/looks_like_headtsx_is_about_to_be_deprecated_on/

Issue #882, opened 2022-02-10
SSR and API routes should support dynamic route naming (with prerendering)


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary

```sh
src/
  artist/
    [name].js/
```

Should support any route like `/artist/${name}/` without needing to know all the URLs up front.

## Details
With this, I think we would also then be able to implement something akin to [Next.js' `getStaticPaths`](https://nextjs.org/docs/pages/api-reference/functions/get-static-paths) API, which could allow users to export a bunch of static content from this dynamic route definition.
```js
export default ArtistDetailsPage extends HTMLElement {
  constructor(request) {
    super();
 
    this.request = request;
  }

  async connectedCallback() {
    const { request } = this;
    const artist = await fetch(request).then(resp => resp.json());
    const { name, bio } = artist;

    this.innerHTML = `
      <h2>${name}</h2>
      <p>${bio}</p>
    `;
  }
}

export const getStaticPaths = (async () => {
  const endpoint = 'http://www.myapi.com';
  const data =  await fetch('${endpoint}/artists').then(resp => resp.json());

  return data.map((artist) => {
    const { id, name } = artist;

    return {
      request: new Request(new URL(`${endpoint}/artists?artistId?=${id}`,
      name: `/artists/${name}/`
    }
}
```

With this data, Greenwood can invoke the page for as many static versions of the page the user wants using:
  - `request` as the "constructor prop" to pass into the page component
  - `name` as the output filename, based on filesystem based routing, e.g. _pages/artist/[name].js_

Effectively this would / could be an alternative and / or complete deprecation of the [sources plugin](https://www.greenwoodjs.io/plugins/source/) and perhaps revisit [workflows around this](https://github.com/ProjectEvergreen/greenwood/discussions/839) overall.

lschierer:
Some other examples of this:
 * [Astro SSR](https://docs.astro.build/en/guides/routing/#server-ssr-mode)
 * [Astro pre-rendered/static](https://docs.astro.build/en/guides/routing/#static-ssg-mode)
 * [Gracile Static Generation](https://gracile.js.org/docs/learn/usage/defining-routes/#doc_staticpaths)  (Gracile can do a fully dynamic SSR page based on path parameters as well, but the documentation is not at all intuitive). 
 
[...]

Issue #1005, opened 2022-11-20
apply isomorphic standards to userland libraries and plugins (`globalThis`)


thescientist13:
## Overview
As called out in [WCC](https://github.com/ProjectEvergreen/wcc/issues/100) and [MDN](), code that needs to run in multiple environments should delegate any "global" references to the new `globalThis`. 
> _The `globalThis` property provides a standard way of accessing the global this value (and hence the global object itself) across environments. Unlike similar properties such as window and self, it's guaranteed to work in window and non-window contexts. In this way, you can access the global object in a consistent manner without having to know which environment the code is being run in. To help you remember the name, just remember that in global scope the this value is `globalThis`._

This means that instead of calling `window.addEventListener` you would want to use `globalThis.addEventListener`. 
 Each environment can choose to polyfill or support some, none, or all of these features, and the code doesn't necessarily have to worry about feature detection, e.g.
```js
if(window) {
  // ... do stuff
}
```

## Details
For Greenwood, this would apply to
- [`staticRouter`](https://github.com/ProjectEvergreen/greenwood/blob/v0.27.0-alpha.6/packages/cli/src/lib/router.js#L45)
- [poyfills](https://github.com/ProjectEvergreen/greenwood/blob/v0.27.0-alpha.6/packages/plugin-polyfills/src/index.js#L30)

Should also update to latest version of WCC to accommodate this.

----

One way to test is to pre-render the website with WCC instead of Puppeteer.

Issue #880, opened 2022-02-09
add support for hydration for compatible renders (`globalThis` state?)


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
At it simplest, in the context of server or pre-rendering, [hydration](https://web.dev/declarative-shadow-dom/#hydration) is the ability for a client side library to re-use the existing HTML (often generated by same client library on the server side) so that on load of the page, the client side library can avoid having to do any DOM manipulation and just build up its internal state / event handers from the existing inert HTML on the page already.  This really help with things like Cumulative Layout Shift (CLI) which is a part of core web vitals.

I think this should be a feature turned off by default, and that can be enabled via _greenwood.config.js_.

## Details
Libraries like [Lit provide such a capability](https://github.com/lit/lit/tree/main/packages/labs/ssr#hydrating-lit-templates) (which will be the pilot for this particular issue though I do want roll our own at some point in service of #548 ) but in the context where _data_ comes into play, that same data must be present in the client as well to ensure that server and client HTML match.

Thus Greenwood needs to expose / inject data somehow into the system we have set up, or alter it to be more accommodating.  For static prerendering, this may require another approach since it won't be coming from a _[route].js_ file.  My initial hunch is to follow a similar convention as per `interpolateFrontmatter` and just stuff everything into `globalThis`, e.g.
```html
<script type="module">
  globalThis = <!-- JSON.stringify(compilation.graph) from the backend -->
</script>
```

PR #1157, opened 2023-10-01
Rfc/issue 952 data loading strategies

### LINKED ISSUE #952: Data Loading Strategies
## Type of Change
Feature

## Summary
Along with #951 , I would like to review data loading strategies with Greenwood.  Also want to revisit the GraphQL plugin now that SSR is in play to see if it makes a difference for `prerender` configuration, which it depends on.

## Details
Take into consideration things like
- For example, having to Prerender data like when sharing templates for SSR - https://www.greenwoodjs.io/docs/configuration/#prerender
- Make graph an actual graph (edges + nodes).  What's the best data structure for this? Otherwise should we not call it a _graph.json_?
- Menus - https://www.greenwoodjs.io/docs/menus/
- Data Sources - https://www.greenwoodjs.io/docs/data/
- rename `imports` field to `resources` in frontmatter?
- How to handle data from the client (eg website header + GraphQL - inline ??) vs SSR.  Can SSR now just use "real" Apollo client? 
- Deprecate Custom Elements flavor of Include Plugin?
- data loaders for SSR pages (e.g. `getServerSideProps` / `getStaticProps`) for SSR / SSG pages and APIs, `export` frontmatter
    - Observed in #1122 , we don't currently have a way to pass request data into SSR pages, be it through an adapter or not.  This would be required to do things like dynamic / slug routing.
    - Should Greenwood or WCC own `getData()` function?
- How to vanilla fetch the graph without GraphQL?  Come in as compilation props?
- Can metadata about the build be passed into an SSR page, like the URL, so that you can dynamically render out static menus on a per page basis, a la - https://github.com/AnalogStudiosRI/www.tuesdaystunes.tv/pull/107
- Change _graph.json_ from `path` -> `url` (instance of URL)
- context plugins as part of SSR
- refactor context, move all things into _manifest.json_

thescientist13:
## Related Issue
resolves #952 

## Summary of Changes
1. Add ~~`async loader` function~~  constructor props for SSR pages request time data (aka `getServerSideProps`)
1. (**breaking**) Refactor / rename `getFrontmatter.static` -> `export const prerender = true` (aka `getStaticProps`)
    - this aligns with [_greenwood.config.js_](https://www.greenwoodjs.io/docs/configuration/#prerender) and #951 
1. Added documentation for ~~`loader`~~ **Web Server Components (WSCs)** and `export const prerender`
1. (**breaking**) Refactor **plugin-graphql**
    - Closer ESM compat (now no more need for Rollup aliases)
    - Document / validate `prerender: true` / custom imports alternative to Puppeteer
1. Upgrade [wcc@0.9.0](https://github.com/ProjectEvergreen/wcc/releases/tag/0.9.0)

## TODO
1. [x] equivalent for `getStaticPaths` - I think this depends on landing #882 , so will need to track there instead
1. [x] Adopt non-loader implementation of data fetching
1. [ ] Align Greenwood and WCC on data loading behavior - https://github.com/ProjectEvergreen/wcc/pull/120
1. [x] Need to do anything for adapters?
1. [x] address PR feedback

## Thoughts / Questions
1. [x] `loaders` pattern vs RSC flavor page components?  - https://github.com/ProjectEvergreen/greenwood/pull/1157/files#r1349748857
1. [x] GraphQL thoughts - should we keep it?
    - can we prerender without puppeteer?  - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/1
    - internal vs external data sources - https://www.greenwoodjs.io/docs/data/ 
    - How to handle data from the client (eg website header + GraphQL - inline ??) vs SSR.  Can SSR now just use "real" Apollo client
    - revisit all currently open "Data" issues - https://github.com/ProjectEvergreen/greenwood/issues?q=is%3Aopen+is%3Aissue+label%3AData
1. [x] Graph refactoring (or own issue?) - https://github.com/ProjectEvergreen/greenwood/issues/1167
    - rename `imports` field to `resources` in frontmatter for graph
    - Change _graph.json_ from `path` -> `url` (instance of URL)
    - Make graph an actual graph (edges + nodes).  What's the best data structure for this? Otherwise should we not call it a _graph.json_?
    - How to vanilla fetch the graph without GraphQL?  Come in as compilation props?
1. [x] ~~Rethink SSR pages API (`getFrontmatter`, `getTemplates`, `getBody`, etc)~~ - will save this for #955 
1. [x] ~~context plugins as part of SSR (for pages and templates work?)~~ - tracking in #955 
1. [x] ~~refactor context, move all things into _manifest.json_~~ - tracking in #955 
1. [x] Deprecate Custom Elements flavor of Include Plugin? - made a note of this to revisit this


packages/cli/test/cases/develop.ssr/src/pages/post.js
```diff
@@ -0,0 +1,26 @@
...
+
+    this.innerHTML = `
+      <h1>Fetched Post ID: ${id}</h1>
+      <h2>${title}</h2>
+      <p>${body}</p>
+    `;
+  }
+}
+
+export async function loader(request) {
```
thescientist13, line 18: Another thought here is to just lean into the `class` based paradigm instead of creating a custom API, e.g. `loaders`.  So basically something like this, wherein the `request` comes through the `constructor` instead, any synchronous operations are done there, and let all the `async` work happen in `connectedCallback`
```js
export default class PostPage extends HTMLElement {
  constructor(request) {
    super();
    
    const params = new URLSearchParams(request.url.slice(request.url.indexOf('?')));
    this.postId = params.get('id');
  }

  async connectedCallback() {
    const { postId } = this;
    const post = await fetch(`https://jsonplaceholder.typicode.com/posts/${postId}`).then(resp => resp.json());
    const { id, title, body } = post;

    this.innerHTML = `
      <h1>Fetched Post ID: ${id}</h1>
      <h2>${title}</h2>
      <p>${body}</p>
    `;
  }
}
```

thescientist13: Per the last Greenwood weekly meeting, a good suggestion was made that the default behavior could be done as in the snippet above, and that possibly for #880 , for users that want to _opt-in_ to hydration, that mechansim could be to export a `loader` function.

This would have a few benefits:
- It helps draw a line between a basic component definition, and a custom bespoke API
- For hydration, we need to know the props.  Having a `loader` function allows us to serialize the props for further hydration at runtime.
- Perhaps to tie it all in together, we can leverage #1005 ?

www/pages/docs/server-rendering.md
```diff
@@ -83,9 +86,10 @@ export default class UsersPage extends HTMLElement {
 }
 ```
 
-In the above example, _card.js_ will automatically be bundled for you on the client side!  🙌
-
-> _**Note**: Keep in mind that for these "page" components, you will likely want to _avoid_ Declarative Shadow for rendering the top level (to avoid wrapping static content in `<template>` tags), but definitely use Declarative Shadow DOM within any dependent custom elements of your page._
+A couple of notes:
+- WSCs run only on the server, thus you have full access to any APIs of the runtime, with the ability to perform one time `async` operations for [data loading](/docs/server-rendering/#data-loading) in `connectedCallback`.
+- In the above example, card.js will automatically be bundled for you on the client side!
```
thescientist13, line 91: Hmm... goes Greenwood actually do this? 🤔 

I feel like yes, but in reality I don't think I've observed it?  Can't say I've directly looked for it since I didn't think it was actually a feature 😅

packages/cli/src/lifecycles/serve.js
```diff
@@ -293,9 +293,9 @@ async function getHybridServer(compilation) {
       const isApiRoute = manifest.apis.has(url.pathname);
       const request = transformKoaRequestIntoStandardRequest(url, ctx.request);
 
-      if (!config.prerender && matchingRoute.isSSR && !matchingRoute.data.static) {
+      if (!config.prerender && matchingRoute.isSSR && !matchingRoute.prerender) {
+        console.log('MATCHING ROUTE -> ', matchingRoute.filename);
```
thescientist13, line 297: Need to delete this console line

packages/plugin-graphql/README.md
```diff
@@ -1,23 +1,15 @@
 # @greenwood/plugin-graphql
 
 ## Overview
-A plugin for Greenwood to support using [GraphQL](https://graphql.org/) to query Greenwood's content graph.  It runs [**apollo-server**](https://www.apollographql.com/docs/apollo-server/) on the backend and provides an [**@apollo/client** _"like"_](https://www.apollographql.com/docs/react/api/core/ApolloClient/#ApolloClient.readQuery) interface for the frontend that you can use.
+A plugin for Greenwood to support using [GraphQL](https://graphql.org/) to query Greenwood's [content graph](https://www.greenwoodjs.io/docs/data/) with our optoinal [pre-made queries](https://www.greenwoodjs.io/docs/menus/).  It runs [**apollo-server**](https://www.apollographql.com/docs/apollo-server/) on the backend at build time and provides a **"read-only"** [**@apollo/client** _"like"_](https://www.apollographql.com/docs/react/api/core/ApolloClient/#ApolloClient.readQuery) interface for the frontend that you can use.
```
thescientist13, line 4: Spelling mistake here on _optional_

thescientist13, line 4: Also a little funky grammar here at the end of the last sentence.

packages/plugin-graphql/README.md
```diff
@@ -1,23 +1,15 @@
...
-# npm
-npm install @greenwood/plugin-renderer-puppeteer --save-dev
-
-# yarn
-yarn add  @greenwood/plugin-renderer-puppeteer --dev
-```
-
-> _We are working on re-evaluating and improving our [data loading](https://github.com/ProjectEvergreen/greenwood/issues/952) and [rendering strategies](https://github.com/ProjectEvergreen/greenwood/issues/951) as part of our [1.0 release](https://github.com/ProjectEvergreen/greenwood/milestone/3)._
+As of now, this plugin requires some form of [prerendering](https://www.greenwoodjs.io/docs/server-rendering/#render-vs-prerender) either through:
+1. Enabling [custom imports](https://www.greenwoodjs.io/docs/server-rendering/#custom-imports)
```
thescientist13, line 11: Clarify that custom imports is related to SSR

Issue #1151, opened 2023-09-09
custom imports bundling breaks when used in API routes and SSR pages


thescientist13:
## Summary
If trying to do something like this with [custom imports](https://www.greenwoodjs.io/docs/server-rendering/#custom-imports)
```js
import styles from './component.css';
```

Rollup throws an error on trying to bundle on the CSS file when used in the aforementioned contexts from the issue title
<details>
  <pre>
        SyntaxError: Unexpected token (1:4)
        at pp$4.raise (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:19420:13)
        at pp$9.unexpected (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:16714:8)
        at pp$9.semicolon (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:16691:66)
        at pp$8.parseExpressionStatement (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:17174:8)
        at pp$8.parseStatement (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:16907:24)
        at pp$8.parseTopLevel (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:16771:21)
        at Parser.parse (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:16543:15)
        at Function.parse (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:16593:35)
        at Graph.contextParse (file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/rollup/dist/es/shared/rollup.js:22959:38)
        at Object.transform (/Users/owenbuckley/Workspace/project-evergreen/greenwood/node_modules/@web/rollup-plugin-import-meta-assets/src/rollup-plugin-import-meta-assets.js:65:24) {
      pos: 4,
      loc: Position { line: 1, column: 4 },
      raisedAt: 5,
      code: 'PLUGIN_ERROR',
[...]

Issue #1004, opened 2022-11-20
support resource plugin based transformations for non "standard" module formats (ex: TypeScript, GraphQL) for SSR


thescientist13:
## Overview
As a companion to https://github.com/ProjectEvergreen/greenwood/issues/878, would like to get support for other current formats working in NodeJS SSR.

### TypeScript
I had started playing around with this as part of #992 and for someone it seems liked the TypeScript plugin `serve` function was not [reading the file path correctly](https://github.com/ProjectEvergreen/greenwood/blob/v0.27.0-alpha.6/packages/plugin-typescript/src/index.js#L40)?

```
{
  moduleURL: URL {
    href: 'file:///Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/plugin-typescript/test/cases/exp-build.prerender/src/components/header.ts',
    origin: 'null',
    protocol: 'file:',
    username: '',
    password: '',
    host: '',
    hostname: '',
    port: '',
    pathname: '/Users/owenbuckley/Workspace/project-evergreen/repos/greenwood/packages/plugin-typescript/test/cases/exp-build.prerender/src/components/header.ts',
    search: '',
    searchParams: URLSearchParams {},
    hash: ''
  }
}

{
  plugins: [
    TypeScriptResource {
      compilation: [Object],
      options: {},
      extensions: [Array],
      contentType: 'text/javascript'
    }
  ]
}

NO URL!!!

      1) "before all" hook in "TypeScript Plugin with static pre-rendering"
{
  moduleContents: 'export default class HeaderComponent extends HTMLElement {\n' +
    '\n' +
    '  async connectedCallback() {\n' +
    '    super.connectedCallback();\n' +
    '\n' +
    '    const navigation = [{\n' +
    "      route: '/about/',\n" +
    "      label: 'About'\n" +
    '    }, {\n' +
    "      route: '/contact/',\n" +
    "      label: 'Contact'\n" +
    '    }];\n' +
    '    \n' +
    '    console.debug({ navigation });\n' +
    '\n' +
    '    /* eslint-disable indent */\n' +
    '    this.innerHTML = `\n' +
    '      <header class="header">\n' +
    '        <nav>\n' +
    '          <ul>\n' +
    '            ${navigation.map((item) => {\n' +
    '              return `\n' +
    '                <li>\n' +
    '                  <a href="${item.route}" title="Click to visit the ${item.label} page">${item.label}</a>\n' +
    '                </li>\n' +
[...]

PR #1166, opened 2023-10-16
Docs/issue 1004 document latest status for non standard import formats

thescientist13:
## Related Issue
"resolves" #1004 

It seems coming out of #952 , using `import` even with _.gql_ files seems to be working just fine, and since this also works with JSX as evidenced in our test cases, I think this means everything is now compatible!
- ✅ _.json_
- ✅ _.css_
- ✅ _.gql_
- ✅ _.jsx_
- ✅ _.ts_ *

> There is a scenario in which we need to handle TypeScript (or any general form of non-standard JavaScript) in WCC - https://github.com/ProjectEvergreen/wcc/issues/122

## Summary of Changes
1. Update documentation (and test case comment) to clarify latest state of custom loaders and configuration
1. Add placeholder TypeScript <> WCC test case

----

Not sure what the specific fix was, but possibly as part #1151 ?

### DIFF
--- a/www/pages/docs/server-rendering.md
+++ b/www/pages/docs/server-rendering.md
@@ -204,10 +204,10 @@ export async function getTemplate(compilation, route) {
 ### Custom Imports
 
 > ⚠️ _This feature is experimental._
+> 
+> _**Note**: At this time, [WCC can't handle non-standard javaScript formats](https://github.com/ProjectEvergreen/greenwood/issues/1004), though we hope to enable this by the 1.0 release._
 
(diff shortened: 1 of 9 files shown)


www/pages/docs/server-rendering.md
```diff
@@ -217,7 +217,7 @@ console.log(json); // { status: 200, message: 'some data' }
 ```
 
 **Steps**
-1. Make sure you are using Node `v18.12.1`
+1. Make sure you are using Node `v18.15.0`
```
thescientist13, line 220: Actually, `18.12.1` I think still works?  So maybe this doesn't need to specifically called out 🤷

Issue #1016, opened 2022-12-02
refactor tests to use `fetch` instead of `request`


thescientist13:
As part of the https://github.com/ProjectEvergreen/greenwood/pull/1014, I realized that we although we have been using **request** to run HTTP calls in our tests, now with native `fetch`, we should use that instead.  That, and also, [**request**](https://www.npmjs.com/package/request) is deprecated.  😅 

So for example, in a few specs we do something like this when testing a server based command
```js
describe('Serve command with API specific behaviors for a JSON API', function() {
  let response = {};

  before(async function() {
    return new Promise((resolve, reject) => {
      request.get(`${hostname}/api/greeting?name=${name}`, (err, res, body) => {
        if (err) {
          reject();
        }

        response.body = JSON.parse(body);

        resolve();
      });
    });
  });

  ...
})
```

Instead, we should now be able to do this
```js
describe('Serve command with API specific behaviors for a JSON API', function() {
  const name = 'Greenwood';
  let response = {};
  let data = {};
  
  before(async function() {
    response = await fetch(`${hostname}/api/greeting?name=${name}`);
    data = await response.json();
  });
  
  ...
})
```

DevLab2425:
@thescientist13 

I gave this one a shot, but getting some failures when running the tests on `master`.

Steps to reproduce:

1. Clone / pull latest from `master`
1. `yarn install --force` to ensure all deps are installed
1. `yarn test`...

The coverage report is created, but the tests fail along with the overall process. Am I missing something?

```
2587 passing (4m)
27 pending
31 failing
...
error Command failed with exit code 31.
```

thescientist13:
@DevLab2425 
So this is just on master without making any changes to the code?  That's odd, I just ran them now myself and all seems fine?
```sh
  2677 passing (3m)
  27 pending
```

> yarn install --force to ensure all deps are installed

Does this change the lockfile at all?  I wonder what version of Yarn you are using?  just `yarn install` in theory should be enough.  (I'm still on 1.x)

One thought is for one of the failing test cases, find it's corresponding _*.spec.js_ file and
1. Change to `describe.only` at the top
1. Pass `true` to [`runner = new Runner()`](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.default/build.config.default.spec.js#L33) to enable logging output

Be curious to see what errors you are getting.

DevLab2425:
@thescientist13 
[...]

PR #1172, opened 2023-10-18
chore/Issue 1016: Replace the use of request with fetch in test cases

DevLab2425:
## Related Issue

https://github.com/ProjectEvergreen/greenwood/issues/1016

## Summary of Changes

- Replace the use of `request` with `fetch`
    - Update related test specs to [match `fetch` response API](https://developer.mozilla.org/en-US/docs/Web/API/Response)

### DIFF
--- a/.mocharc.cjs
+++ b/.mocharc.cjs
@@ -1,5 +1,3 @@
 module.exports = {
-  timeout: 90000,
-  bail: true,
-  exit: true,
+  timeout: 90000
 };
--- a/.mocharc.cjs
+++ b/.mocharc.cjs
@@ -1,3 +1,5 @@
 module.exports = {
-  timeout: 90000
+  timeout: 90000,
+  bail: true,
+  exit: true,
 };
--- a/packages/plugin-typescript/test/cases/develop.default/develop.default.spec.js
+++ b/packages/plugin-typescript/test/cases/develop.default/develop.default.spec.js
@@ -60,7 +60,7 @@ describe('Develop Greenwood With: ', function() {
       let data;
 
       before(async function() {
-        response = await fetch(`http://127.0.0.1:${port}/main.ts`);
+        response = await fetch(`${hostname}:${port}/main.ts`);
         data = await response.text();
       });
 
--- a/packages/plugin-typescript/test/cases/develop.default/develop.default.spec.js
+++ b/packages/plugin-typescript/test/cases/develop.default/develop.default.spec.js
@@ -18,7 +18,6 @@
  */
 import chai from 'chai';
 import path from 'path';
-import request from 'request';
 import { Runner } from 'gallinago';
 import { fileURLToPath, URL } from 'url';
 import { runSmokeTest } from '../../../../../test/smoke-test.js';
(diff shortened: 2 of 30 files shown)


packages/plugin-typescript/test/cases/develop.default/develop.default.spec.js
```diff
@@ -58,37 +58,23 @@ describe('Develop Greenwood With: ', function() {
...
+        response = await fetch(`http://127.0.0.1:${port}/main.ts`);
+        data = await response.text();
       });
 
-      it('should return a 200', function(done) {
-        expect(response.statusCode).to.equal(200);
-
-        done();
+      it('should return a 200', function() {
+        expect(response.status).to.equal(200);
```
DevLab2425, line 69: @thescientist13 
Just wanted to double check on these changes before I start to apply them to the other files. I only really ask because it doesn't seem like anything broke when I changed the spec.

thescientist13: yeah, this looks to be on the right track!  I feel like there was a case I tried that was tricky and couldn't get it work.  Maybe do a "spot" test on the `request.post` usages in [this spec](https://github.com/ProjectEvergreen/greenwood/blob/release/0.29.0/packages/cli/test/cases/develop.default/develop.default.spec.js) on the release/0.29.0 branch?

Also worth testing in any [`serve` based test case](https://github.com/ProjectEvergreen/greenwood/blob/release/0.29.0/packages/cli/test/cases/serve.default/serve.default.spec.js), I noticed in those cases I had to use `.127.0.0.1` and maybe that was where I got stuck?

But if you can get that cross section of test cases working, then I think it should be full steam ahead at that point.

DevLab2425: @thescientist13 

> Maybe do a "spot" test....
Was successful on the spot test: https://github.com/ProjectEvergreen/greenwood/compare/release/0.29.0...DevLab2425:greenwood:issue-1016-release-update

> I had to use .127.0.0.1 .....
Will report back on the IP-based tests.

DevLab2425: I was also successful with replacing the IP with localhost.

Updated compare: https://github.com/ProjectEvergreen/greenwood/compare/release/0.29.0...DevLab2425:greenwood:issue-1016-release-update

DevLab2425: Do you want me to target the v0.29.0 branch for this PR too? It'll have ~90 updates/~25 files with all of the replacements.

thescientist13: Yeah, if you've got it all figured out, feel free to target that release branch! 🙏

thescientist13:
Got a little time, let me see if I can help with the build failures and merge conflicts.

.mocharc.cjs
```diff
@@ -1,3 +1,5 @@
 module.exports = {
-  timeout: 90000
+  timeout: 90000,
+  bail: true,
```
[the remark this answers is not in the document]
thescientist13: I ended up rolling this one back since it was a little frustrating to have to keep re-running specs and then getting halted when one fails, re-starting the runner, bailing at the next failure, etc.  It can be nice to let things keep running to see all the failures in totality during a single run.

Perhaps this could be useful as a CI only flag?  So we don't keep wasting build time after a failure?  🤔

thescientist13 (approved):
Phew, thanks for getting this effort started @DevLab2425 !  After rebasing and resolving the merge conflicts, I was able to get to the rest of the test cases, so this is all looking good to go now!

Feels a lot more lightweight and intuitive using `fetch` 💯

PR #1188, opened 2023-12-05
update test endpoint to fix suddenly failing test cases

thescientist13:
## Related Issue
In the wake of https://github.com/AnalogStudiosRI/www.analogstudios.net/pull/97, forgot to come back and update this project to use the new v2 endpoints.  Otherwise, all related test cases were failing, as observed in https://github.com/ProjectEvergreen/greenwood/pull/1186#issuecomment-1839136253.

## Summary of Changes
1. Update all Analog Studios endpoints from `/api/artists` -> `/api/v2/artists` for impacted test cases

PR #1189, opened 2023-12-05
Issue-1188: Resolve invalid test endpoint

DevLab2425:
## Related Issue
Fix: #1188 

## Summary of Changes

- Created a `test-constants.js` file to store shared test values
- Replace Artists legacy API endpoint with constant value
- Update markdown files that reference the API


thescientist13:
I think #1188 is probably sufficient and the blog posts were merely illustrative with no hostname anyway.  If anything I could see updating all specs to use the same placeholder API I've started using in other tests but not a big deal.

Issue #1232, opened 2024-05-18
Sitemap Generation


thescientist13:
## Summary
A sitemap is an XML file used to tell Search Engines about the content and pages contained within a site, in particular for larger sites and / or where links between pages is maybe not as consistent.
https://developers.google.com/search/docs/crawling-indexing/sitemaps/overview

>  _A sitemap tells search engines which pages and files you think are important in your site, and also provides valuable information about these files. For example, when the page was last updated and any alternate language versions of the page._

Here is a basic example
https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap
```xml
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <url>
    <loc>https://www.example.com/foo.html</loc>
    <lastmod>2022-06-04</lastmod>
  </url>
</urlset>
```

## Details
I think the approach used in Next.js is probably good enough for Greenwood supporting either of this options
1. Static File, e.g. sitemap.xml - will be copied automatically to the output
1. Dynamic File, e.g. sitemap.xml.js - will be provided a copy of the greenwood graph and be expected to return valid XML
    ```js
    export async function sitemap(compilation) {
      const urls = compilation.graph.map((page) => {
        return `
          <url>
            <loc>http://www.example.com${page.route}</loc>
          </url>
        `;
      });

      return `
        <?xml version="1.0" encoding="UTF-8"?>
        <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
          ${urls}
        </urlset>
      `
    }
    ```
    
----

Might want to wait until after #955 is merged since we might want to piggy back off any solutions there re: extending the ability for pages to be more than just markdown (_.md_) or JavaScript (_.js_).

thescientist13:
For now a couple ways to implement this manually could be to:
1. Create / maintain a _src/sitemap.xml_ and use a copy plugin to put into the output directory
1. After the `greenwood build` step, read the contents of _graph.json_ in the output directory and generate the file

jstockdi:
For 2, would it be a copy plugin?  ie, the plugin would generate a temporary file, then pass 

```
     {
        from: tempPath,
        to: new URL(`sitemap.xml`, outputDir)
     }
```

thescientist13:
@jstockdi 
[...]

PR #1240, opened 2024-06-05
feature/issue 1232 static sitemap

jstockdi:
## Related Issue
#1232 - Static sitemap copy plugin

## Summary of Changes
Created a greenwood default copy plugin to copy a sitemap.xml from the source folder

1. [x] Created plugin
1. [x] Updated relevant spec file
1. [x] Provided example sitemap.xml
2. [x] Thought about documentation in configuration.md or plugins.md (Happy to help given direction)

### DIFF
--- a/www/pages/docs/css-and-images.md
+++ b/www/pages/docs/css-and-images.md
@@ -110,3 +110,20 @@ customElements.define('x-header', HeaderComponent);
 >
 > customElements.define('x-header', HeaderComponent);
 > ```
+
+
+### Meta files
+
+The build includes a few [default copy plugins](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/src/plugins/copy) for common meta files.   Typically, they support copying the files from the source directory to the build directory.
+
+* `plugin-copy-favicon.js` - copies favicon.ico.  NOTE:  Does not support multiple favicons.  For multiple support, use the manifest.json or make a custom plugin
+* `plugin-copy-manifest-json.js` - copies manifest.json
+* `plugin-copy-robots.js` - copies robots.txt
+* `plugin-copy-sitemap.js` - copies sitemap.xml
+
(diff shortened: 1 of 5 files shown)


www/pages/docs/css-and-images.md
```diff
@@ -105,4 +105,21 @@ customElements.define('x-header', HeaderComponent);
...
-> ```
\ No newline at end of file
+> ```
+
+
+### Meta files
+
+The build includes a few [default copy plugins](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/src/plugins/copy) for common meta files.   Typically, they support copying the files from the source directory to the build directory.
+
+* `plugin-copy-favicon.js` - copies favicon.ico.  NOTE:  Does not support multiple favicons.  For multiple support, use the manifest.json or make a custom plugin
```
thescientist13, line 115: Nice, I think I will tweak this as it interweaves public API / private implementation / development + testing details and so I think for the sure of readers, probably just listing out the supported files are is good.  Though you make a good callout that if you need _more_ support, then linking to the copy plugin is a good idea, so will definitely keep that in.

As a side note, _manifest.json_ is more related to Greenwood's own build output, so not really something for users per se, but I do have a PR that is documenting more of that as we speak, as it pertains to [adapter plugins](https://github.com/ProjectEvergreen/greenwood/pull/1237/files#diff-adb56e4ca10fb02f0c5a8695f668985a028ad038a5bab81bd85359167770e638R37) anyway, and will likely get its own "Build Output" section when migrated to the new docs website soon.

thescientist13 (approved):
Awesome, thanks for this!  I am going to queue this up for the next release coming out soon (next 4 - 5 weeks) and is in progress at #1208 .  I will get this out in an alpha release before then so will keep you posted.  It's a clean enough change so I don't expect any issues rebasing onto that release branch so for now tagging it for **alpha.3**, which I expect to get by next weekend at the latest.

Looking forward to collaborating on the dynamic version of this next, and can make some time to discuss the approach on our next weekly meeting if you're interested.

----

edit: also just a reminder you are a collaborator within this org, so you don't have to fork and can commit branches right to upstream, just to help save on the hassle of the forking workflow if we want.  👍

PR #1245, opened 2024-06-14
Dynamic Sitemap Copy Plugin (#1232)

jstockdi:
## Submitting a Pull Request
Submitting a copy plugin based approach for #1232

## Summary of Changes
Adds a package with unit test example

1. [x] Built plugin
1. [x] Created test
1. [x] Added jsdoc

### DIFF
--- a/packages/plugin-dynamic-sitemap/src/index.js
+++ b/packages/plugin-dynamic-sitemap/src/index.js
@@ -1,37 +1,30 @@
-import { checkResourceExists } from "@greenwood/cli/src/lib/resource-utils.js";
 import fs from 'fs/promises';
 
-
-
-
-  
 const greenwoodPluginDynamicExport = (options = {}) => [{
   type: 'copy',
   name: 'plugin-dynamic-sitemap',
   provider: async (compilation) => {
-    
-    const { base_url}  = options;
+
+    const { baseUrl } = options;
     const { outputDir } = compilation.context;
-    
+
     let sitemapXML = '<?xml version="1.0" encoding="UTF-8"?>\n';
-    sitemapXML += '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n'
+    sitemapXML += '<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n';
 
     compilation.graph.forEach(page => {
-      sitemapXML += `<url><loc>${base_url}${page.outputPath}</loc></url>\n`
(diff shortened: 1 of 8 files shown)


packages/plugin-dynamic-sitemap/src/index.js
```diff
@@ -0,0 +1,30 @@
+import fs from 'fs/promises';
+
+const greenwoodPluginDynamicExport = (options = {}) => [{
```
thescientist13, line 3: Interesting approach and maybe it's just something I'm missing here but the main concern / limitation I see with this approach is that there is no way to programmatically build your _own_ sitemap.  To me, the advantage of the dynamic sitemap option was that could automatically build it up whatever way you need, even tapping into your own frontmatter / page metadata.

Referencing my back of the napkin example in #1232, which is basically just taking a page out of the [Next.js playbook](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap), I would expect users to be able to make a _sitemap.xml.(js|ts)_ file in the root of their workspace (e.g. _src/sitemap.xml.js_, which would just export a function in which we pass the compilation to them, and they can build it up themselves and set their own values for all [supported properties](https://www.sitemaps.org/protocol.html), since there are things they might want
```js
    const { data, route } = route;

    return `
      <url>
        <changefreq>${data.freq}</changefreq>
        <priority>https://www.sitemaps.org/protocol.html#prioritydef>${data.priority ?? '1.0'}</priority>
      </url>
    `;
  });
  const sitemap = `
      ${urls.join('\n')}
    </urlset>
  `;

  return new Response(body, { headers: 'Content-Type' : 'text/xml' })
}
```

----

As far as how to do this through greenwood, I think [the API Resource](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/plugins/resource/plugin-api-routes.js#L15) as a reference is probably going to get us halfway there at least for development, such that hitting `https://localhost:1984/sitemap.xml` returns something.

For build time, we maybe add something to the build step [like we do for adapters](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/commands/build.js#L120), where if we have a sitemap (perhaps the graph can track this on init?) we `import` the that function from the user's file and write out the file for them?

thescientist13 (commented):
Left a comment regarding technical approach, and I think this should also be in core, if we're going to support the static version through core too.

Let's chat during tomorrow's meeting though!

PR #1260, opened 2024-07-21
feat(cli): #1232 dynamic sitemap generation

### LINKED ISSUE #1232: Sitemap Generation

jstockdi:
Continuing work on https://github.com/ProjectEvergreen/greenwood/pull/1245 — reworked and rebased onto current master (July 2026).

## Related Issue

Resolves #1232

## Documentation

If the approach looks right, I'll add a docs section for `sitemap.xml.js` / `generateSitemap(compilation)` alongside the existing static sitemap docs.

## Summary of Changes

1. [x] `packages/cli/src/plugins/copy/plugin-copy-sitemap.js`: when no static `src/sitemap.xml` exists and the workspace has a `src/sitemap.xml.js` module, write its `generateSitemap(compilation)` output to the scratch dir and copy it to the output. The static file wins when both exist, so existing projects are unaffected.
2. [x] `packages/cli/src/plugins/resource/plugin-resource-sitemap.js` (new): serves `/sitemap.xml` from the same `generateSitemap` hook during `greenwood develop`.
3. [x] Tests, modeled on current test case conventions:
   - `packages/cli/test/cases/build.default.sitemap-dynamic/` — build emits the generated `sitemap.xml` with a `<loc>` per page in the graph
   - `packages/cli/test/cases/develop.default.sitemap-dynamic/` — dev server serves the generated `/sitemap.xml` as `text/xml`

### Changes from the previous revision

- Dropped the adapter-based approach per review — everything is in the out-of-the-box copy/resource plugins now.
- Reconciled with the static sitemap support that landed in #1240 (static copy behavior unchanged).
- `generateSitemap(compilation)` remains the user hook: Greenwood hands over the compilation and the user maps `compilation.graph` to XML. Pages generated via dynamic routing / `getStaticPaths` appear in the graph like any other page, so they flow into the sitemap with no extra API.

### DIFF
--- a/packages/cli/src/plugins/copy/plugin-copy-sitemap.js
+++ b/packages/cli/src/plugins/copy/plugin-copy-sitemap.js
@@ -1,4 +1,5 @@
 import { checkResourceExists } from "../../lib/resource-utils.js";
+import fs from "node:fs/promises";
 
 const greenwoodPluginCopySitemap = [
   {
@@ -6,13 +7,27 @@ const greenwoodPluginCopySitemap = [
     name: "plugin-copy-sitemap",
     provider: async (compilation) => {
       const fileName = "sitemap.xml";
-      const { outputDir, userWorkspace } = compilation.context;
+      const { outputDir, scratchDir, userWorkspace } = compilation.context;
       const sitemapPathUrl = new URL(`./${fileName}`, userWorkspace);
+      const sitemapModuleUrl = new URL(`./${fileName}.js`, userWorkspace);
       const assets = [];
 
       if (await checkResourceExists(sitemapPathUrl)) {
         assets.push({
           from: sitemapPathUrl,
+          to: new URL(`./${fileName}`, outputDir),
+        });
+      } else if (await checkResourceExists(sitemapModuleUrl)) {
+        // generate a dynamic sitemap from the user's sitemap.xml.js module
+        // https://github.com/ProjectEvergreen/greenwood/issues/1232
+        const { generateSitemap } = await import(sitemapModuleUrl.href);
+        const sitemap = await generateSitemap(compilation);
+        const sitemapScratchUrl = new URL(`./${fileName}`, scratchDir);
+
+        await fs.writeFile(sitemapScratchUrl, sitemap, "utf-8");
+
+        assets.push({
+          from: sitemapScratchUrl,
           to: new URL(`./${fileName}`, outputDir),
         });
       }
--- a/packages/cli/src/plugins/resource/plugin-resource-sitemap.js
+++ b/packages/cli/src/plugins/resource/plugin-resource-sitemap.js
@@ -1,0 +1,39 @@
+import { checkResourceExists } from "../../lib/resource-utils.js";
+
+class SitemapResource {
+  constructor(compilation) {
+    this.compilation = compilation;
+    this.contentType = "text/xml";
+  }
+
+  async shouldServe(url) {
+    const { userWorkspace } = this.compilation.context;
+
+    return (
+      url.pathname === "/sitemap.xml" &&
+      (await checkResourceExists(new URL("./sitemap.xml.js", userWorkspace)))
+    );
+  }
+
+  async serve() {
+    const { userWorkspace } = this.compilation.context;
+    // generate a dynamic sitemap from the user's sitemap.xml.js module
+    // https://github.com/ProjectEvergreen/greenwood/issues/1232
+    const { generateSitemap } = await import(new URL("./sitemap.xml.js", userWorkspace).href);
+    const body = await generateSitemap(this.compilation);
+
+    return new Response(body, {
+      headers: new Headers({
+        "Content-Type": this.contentType,
+      }),
+    });
+  }
+}
+
+const greenwoodPluginResourceSitemap = {
+  type: "resource",
+  name: "plugin-resource-sitemap",
+  provider: (compilation) => new SitemapResource(compilation),
+};
+
+export { greenwoodPluginResourceSitemap };
--- a/packages/cli/test/cases/build.default.sitemap-dynamic/build.default.sitemap-dynamic.spec.js
+++ b/packages/cli/test/cases/build.default.sitemap-dynamic/build.default.sitemap-dynamic.spec.js
@@ -1,0 +1,77 @@
+/*
+ * Use Case
+ * Run Greenwood build command with a dynamic sitemap module (src/sitemap.xml.js) and no static
+ * sitemap.xml file in the workspace.
+ *
+ * User Result
+ * Should generate a Greenwood build with a sitemap.xml in the output directory, generated from
+ * the generateSitemap export of the user's sitemap.xml.js module.
+ *
+ * User Command
+ * greenwood build
+ *
+ * User Config
+ * (default)
+ *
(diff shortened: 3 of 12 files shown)


thescientist13 (commented):
Hey @jstockdi , thanks for plugging away at this.

Not sure if you had a chance  to review [my thoughts](https://github.com/ProjectEvergreen/greenwood/pull/1245#discussion_r1645268168) from the last PR?  This one seems to be going in a couple different directions and so just want to make sure we're on the same page.

For example a couple observations that are unclear to me at an immediate glance:
- [adapters](https://www.greenwoodjs.io/plugins/adapter/) are more intended for preparing "standard" Greenwood output to other hosting targets, like Netlify and Vercel (mainly for SSR pages and API routes)
- I think if a static sitemap is supported out of the box, why not the dynamic one too?  (as opposed to through a plugin).  Or maybe all sitemap behaviors should be a plugin?

The main thing I see that I think we need to consider is giving the _user_ a hook to generate their _own_ sitemap, as opposed to us trying to it all for them.

Although not in Greenwood yet, one future facing feature we hope to get around to soon is [dynamic routing](https://github.com/ProjectEvergreen/greenwood/issues/882), e.g. _src/pages/blog/[slug].js_ and so I wonder how Greenwood (or Next for that matter) handle such an open ended sitemap (if not using [`getStaticPaths`](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-paths))

So would be good to consider how this might interplay with that.  🤔 

----

I think I might put this one draft though for now and then we can discuss and add more details after our next meeting sync.

jstockdi:
Rebased onto current master (4d23087f) — the branch was v0.29-era, so this is a rework rather than a mechanical rebase. Addressing the open review points from last time:

- **Adapters dropped.** Agreed they're for hosting targets; the dynamic sitemap now lives in the same out-of-the-box plugins as the static one: `plugin-copy-sitemap.js` (build) plus a new `plugin-resource-sitemap.js` (develop, serves `/sitemap.xml`).
- **Static + dynamic both out of the box.** Since #1240 landed, a static `src/sitemap.xml` already copies through. This PR adds the dynamic case on top: if there's no static file and the workspace has a `src/sitemap.xml.js` module, its `generateSitemap(compilation)` export is written to the scratch dir and copied out. Static wins when both exist, so existing projects are unaffected.
- **User hook, not magic.** `generateSitemap(compilation)` is exactly the "let the user generate their own" shape — Greenwood hands over the compilation and the user maps `compilation.graph` (or whatever they want) to XML. Since dynamic routing and `getStaticPaths` have landed since the last review, generated static routes appear in the graph like any other page, so they flow into the sitemap with no extra API.

Tests: `build.default.sitemap-dynamic` and `develop.default.sitemap-dynamic`, modeled on current case conventions; both pass locally along with `build.default.meta-files` (static sitemap regression).

packages/cli/src/plugins/copy/plugin-copy-sitemap.js
```diff
@@ -1,20 +1,35 @@
...
 const greenwoodPluginCopySitemap = [
   {
     type: "copy",
     name: "plugin-copy-sitemap",
     provider: async (compilation) => {
       const fileName = "sitemap.xml";
-      const { outputDir, userWorkspace } = compilation.context;
+      const { outputDir, scratchDir, userWorkspace } = compilation.context;
       const sitemapPathUrl = new URL(`./${fileName}`, userWorkspace);
+      const sitemapModuleUrl = new URL(`./${fileName}.js`, userWorkspace);
```
thescientist13, line 12: Just curious if a developer were to `import` something in their _sitemap.xml.js_ file, then it probably wouldn't work, right?  I'm thinking cases where a user may be pulling from a CMS or database for dynamic sitemap generation, I don't think this approach would work for that.  Same thing would go for adapter plugins too, I would think.

That's why I was thinking instead of copy, this should effectively be considered an SSR page like any other page.  Still not sure if we should support it as `sitemap.(js|ts)`, with or without the _.xml_ in it.
https://nextjs.org/docs/app/api-reference/file-conventions/metadata/sitemap#generating-a-sitemap-using-code-js-ts

packages/cli/src/plugins/resource/plugin-resource-sitemap.js
```diff
@@ -0,0 +1,39 @@
...
+      url.pathname === "/sitemap.xml" &&
+      (await checkResourceExists(new URL("./sitemap.xml.js", userWorkspace)))
+    );
+  }
+
+  async serve() {
+    const { userWorkspace } = this.compilation.context;
+    // generate a dynamic sitemap from the user's sitemap.xml.js module
+    // https://github.com/ProjectEvergreen/greenwood/issues/1232
+    const { generateSitemap } = await import(new URL("./sitemap.xml.js", userWorkspace).href);
```
thescientist13, line 22: we would probably want to support typescript files too. 

----

Also, just thinking outloud, I think given Greenwood's conventions, it might be good to return a `Response` object from the sitemap file?  Sveltekit is a good example
https://svelte.dev/docs/kit/seo#Manual-setup-Sitemaps

WDYT?

jstockdi:
Agreed with the direction from both review threads — treating this as a `Response`-returning route is the better convention, and rather than rework this branch a third time I've reimplemented it fresh from current master: **#1747** supersedes this PR.

Answering the two open threads here for the record:

- **"if a developer were to `import` something in their sitemap.xml.js file, then it probably wouldn't work"** — small correction: for static builds it does work, since the module executes in Node at build time. The structural gap was adapters/on-demand SSR: the raw file was never bundled, so a serverless target couldn't run it. #1747 closes that by bundling the module through the same Rollup path as API routes (node export conditions, chunks tracked in the manifest) and generating `public/sitemap.xml` by executing the *bundled* handler — so CMS/database imports are proven to survive bundling, and adapters have a mountable artifact.
- **TypeScript + returning a `Response` (SvelteKit-style)** — both adopted in #1747: the contract is now `handler(request, { compilation })` returning a `Response`, identical in shape to API routes, with a `sitemap.xml.ts` test case.

Closing this one in favor of #1747.

Issue #1223, opened 2024-04-26
nested SSR pages (and API routes) are throwing a module not found error (needs additional patch for adapters)


thescientist13:
## Summary
Haven't tested for production yet, but at least when running `greenwood develop` and having a nested SSR page route, e.g.
```
src/
  pages/
    blog/
      index.js
```

Will output
```sh
Started local development server at http://localhost:1984
      Develop command with HTML route response using default export and nested SSR page
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/test/cases/develop.ssr/src/pages/index.js' imported from /Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lib/execute-route-module.js
    at new NodeError (node:internal/errors:405:5)
    at finalizeResolution (node:internal/modules/esm/resolve:327:11)
    at moduleResolve (node:internal/modules/esm/resolve:980:10)
    at defaultResolve (node:internal/modules/esm/resolve:1193:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:404:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:373:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:250:38)
    at ModuleLoader.import (node:internal/modules/esm/loader:336:34)
    at importModuleDynamically (node:internal/modules/esm/translators:143:35)
    at importModuleDynamicallyCallback (node:internal/modules/esm/utils:176:14) {
  url: 'file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/test/cases/develop.ssr/src/pages/index.js',
  code: 'ERR_MODULE_NOT_FOUND'
```

## Details
If you notice in the path shown in the error, the _blog/_ segment is missing
```sh
Cannot find module '/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/test/cases/develop.ssr/src/pages/index.js' 
```

thescientist13:
Argh, completely forgot to test for adapters and this is definitely not working based on some local testing I just conducted for this, so need to re-open 😞

Issue #552, opened 2021-04-11
Graph schema consistency enhancements


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Make some enhancements / cleanups to the the graph schema
1. Unique IDs
1. Double `MenuQuery` Item

### Unique IDs
We should make sure all [content gets a unique ID](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/cli/src/lifecycles/graph.js#L98) since right now it is just the filename without the extension.  Could probably be based on a hash of the entire file page.

### Double MenuQuery Item 
The [`MenuQuery`](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/plugin-graphql/src/queries/menu.gql) seems to return nested [`item.item`](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/www/components/header/header.js#L40) which makes it  a little clunky to interact with.
```js
query($name: String, $route: String, $order: MenuOrderBy) {
  menu(name: $name, pathname: $route, orderBy: $order) {
    item {
      label,
      route
    }
    children {
      item {
        label,
        route
      },
      children {
        item {
          label,
          route
        }
      }
    }
  }
}
```

Wonder if there is a way to add some ergonomics [here](https://github.com/ProjectEvergreen/greenwood/blob/release/0.10.0/packages/plugin-graphql/src/schema/graph.js#L22), or maybe give the sub items a better, more semantic name, like `children`?

thescientist13:
We now have a hashing util if that helps
https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lib/hashing-utils.js

thescientist13:
Not sure if this is really that valuable on first refresh of this?  I suppose can revisit post 1.0.

thescientist13:
Did a big graph refactor as part of #1167 / #1266 and now `id` is supported, so this can be resolved as part of that one.

PR #1266, opened 2024-08-11
rfc/issue 1167 content as data

### LINKED ISSUE #1167: Content as Data
## Type of Change
RFC

## Summary
Coming out of #952 (and now while working on the new Greenwood website), wanted to track some specific thoughts and ideas around accessing content as data with Greenwood and making it easier to work in HTML / markdown for happy path cases, and mostly to provide an alternative to using GraphQL as built into Greenwood.

In addition, the "graph", if that's even a good name for it, is showing its age by now and in need of some TLC.

### Collections

Although Greenwood has the concept of [menus](https://www.greenwoodjs.io/docs/menus/), which is a way to group content in the context of a navigation and whatnot, perhaps a more general purpose name going forward may a "Collection", as inspired by Astro's (Content Collections](https://docs.astro.build/en/guides/content-collections/).   Ultimately it should be whatever meaning the user wants to give, be it navigation, a bunch of blog posts, or something else entirely.

Things like:
- exposing the menu and children query as JavaScript as standalone functions / helpers
- making it possible to do next / prev links (like for sequential steps for tutorials, blogs, etc)
- listing out a table of contents

(I think Menu's may have added extra functionality on top of that like ordering and what not, but maybe that should just be considered a "type" of collection?)

For example, say listing out the table of contents for a blog using constructor props instead of the [**Children** GraphQL query](https://www.greenwoodjs.io/docs/data/#internal-sources)
```js
// notice this is not a .gql file, but perhaps the logic could be shared?
import { getChildren } from '@greenwood/cli/src/queries/menu.js';

export default class BlogPostsPage extends HTMLElement {
  constructor(request, graph) {
    super();

    this.graph = graph;
  }

  async connectedCallback() {
    const { graph } = this;
    const blogPosts = getChildren(graph, { parent: 'blog' });

    this.innerHTML = `
      ${
         blogPosts.map((post) => {
           return `<h2>${post.title}</h2>
         }).join('');
        }
    `;
  }
}
```

#### Frontmatter

We may also want to rename it from `index` to `order` since these aren't arrays, and the starting index thing may be confusing / misleading.
```md
---
title: 'Docs'
collection: 'nav'
order: 1
---

## Docs

Welcome to the docs page
```

#### HTML

We already have frontmatter support for markdown and JS
```js
export async function getFrontmatter(compilation) {
 return { /* ... */
}
```

But might be good to see if we can extend frontmatter support to HTML as well, assuming it doesn't break all the HTML parsers
```html
---
collection: 'nav'
order: 2
---
<html>
  <head>
    <title>Guides</title>
  </head>
  <body>
    <h2>Guides Page
  </body>
</html
```

### Rich Frontmatter

As reported in #1229 , would be good to see how we can support "rich" frontmatter data, and to the use case presenting point, what more could we do with frontmatter in this use case, as it would be really nice to leverage this data somehow combined with a custom element
```md
---
name: 'My Playlist'
songs:
  - title: You Can Close Your Eyes
    url: http://archive.org/download/james-taylor-montreux-jazz-fest-1988-rsr/01 You Can Close Your Eyes.mp3
  - title: Country Road
    url:  http://archive.org/download/james-taylor-montreux-jazz-fest-1988-rsr/09 Country Road.mp3
---

## My Playlist

<x-playlist
  title="${globalThis.page.name}"
  songs="${JSON.stringify(globalThis.page.songs)}"
></x-playlist>
```        

> Might also be worth revisiting this pattern as well - https://www.greenwoodjs.io/docs/configuration/#interpolate-frontmatter

### Rendering Strategies / Contexts

Somewhat related to #951 , one caveat around content as data is what to about CSR use cases, like when authoring HTML.  How can you get content as data into a custom element using only HTML as you will be coupled to API calls and have to handle them somehow for production builds, since the plan isn't to support an actual Greenwood GraphQL server for production, and so all data is pre-generated at build time, this making it pretty much read-only.

```html
<!doctype html>
  <head>
    <title>Greenwood</title>
    <script type="module" src="../components/footer/footer.js" data-gwd-opt="static"></script>
    <script type="module" src="../components/header/header.js"></script>

    <script type="application/json">
      /* graph JSON.stringified here??? */
    </script>
    
    <script type="module">
      const graph = await fetch('./graph.json').then(resp => resp.json());
      const nav = graph
        .filter(page => page.data.collection === 'nav')
        .sort((a, b) => a.data.index > b.data.index ? 1 : -1);

      console.log({ nav });
    </script>
  </head>

  <body>
    <!-- header component my also have a nav component to share the data with too --> 
    <app-header nav="???"></app-header>
    <h1>Welcome to my website!</h1>
    <app-footer></app-footer>
  </body>
</html>
```

### Graph Structure

There are definitely some breaking changes to be made to the graph, to better align on terminology and overall usefulness to Greenwood and users.  So given a current item in the graph
```js
{
  page: {
    data: { menu: 'side', index: 7, linkheadings: 0, tableOfContents: [] },
    filename: 'source.md',
    id: 'source',
    label: 'Source',
    imports: [],
    resources: [],
    outputPath: '/plugins/source/index.html',
    path: 'www/pages/plugins/source.md',
    route: '/my-subpath/plugins/source/',
    template: 'page',
    title: 'Source',
    isSSR: false
  }
}
```

- In the graph, there seems to be a lot of cruft that isn't actually useful
    - it doesn't "scale", e.g. for nested pages - see #1223 
    - What is the meaningful difference  between `id` and `label`?  `path` and `filename`?  anything else worth cleaning up?
- Rename `imports` field to `resources` in frontmatter config
- Rename `index` -> `order`
- Change  `path` -> `url` (instance of URL or at least the `href` property for serialiazing)
- Something more predictable for `outputPath` / `id`
    - https://github.com/ProjectEvergreen/greenwood/pull/1224/files#diff-7bec4c1c598f208ba9ec341822f12eace13bede43b5b4d07a9502faf68b3b95dR194
    - https://github.com/ProjectEvergreen/greenwood/pull/1237

I wonder if there is a case to rename this to `pages` instead?  A graph implies nodes, some sort of linking structure, etc which it is definitely not in its current state.

-----

Would be good to go through and evaluate all currently [open and related issues](https://github.com/ProjectEvergreen/greenwood/issues?q=is%3Aissue+is%3Aopen+label%3AData)

thescientist13:
## Related Issue
resolves #1167 

> For a demo of all these features see https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/49/

## Summary of Changes
1. Introduced _Collections_ as the successor to Menus with test cases
1. Add support for "rich" frontmatter with test cases
1. Introduce a first party "content" fetching helper query utilities
1. Put content as data and active frontmatter behind the same config key (`activeContent`)
1. Refactor and clean up graph properties, with better defaults (and updated website frontmatter accordingly)
1. Apply appropriate Netlify redirects in _netlify.toml_
1. Refreshed content as data and adjacent docs, moved more of the GraphQL docs to the plugin's README
1. Handles https://github.com/ProjectEvergreen/greenwood/issues/552

## TODOs
1. [x] Rename `interpolateFrontmatter` -> `activeFrontmatter`
1. [x] Frontmatter / graph node properties refactoring (and update website as needed)
    - ~~rename `linkheadings` -> `toc`~~
    - ~~What is the meaningful difference  between `id` and `label`?  `path` and `filename`?  anything else worth cleaning up?~~ - better generated `id`, and gave better definitions to everything else
    - ~~Rename `index` -> `order`~~
    - ~~rename `menu` -> `collection`~~
    - ~~default `title` and `label` should be the same and actually meaningful)~~
    - ~~Change  `path` -> `url` (instance of URL or at least the `href` property for serializing)~~
    - ~~Something more predictable for `outputPath`~~ - renamed to `outputHref` 
      - https://github.com/ProjectEvergreen/greenwood/pull/1224/files#diff-7bec4c1c598f208ba9ec341822f12eace13bede43b5b4d07a9502faf68b3b95dR194
      - https://github.com/ProjectEvergreen/greenwood/pull/1237
    - ~~it (`id`) doesn't "scale", e.g. for nested pages~~ - see #1223
1. [x] GraphQL refactoring
    - Delete config schema for GraphQL
    - align with CLI collection utils (ideally)
1. [x] Make more / less content as data available through active frontmatter + test cases
    ```html
    <!-- so would be nice if we could use other content as data, e.g. -->
    <app-toc route="${globalThis.page.route}"></app-toc>

    <!-- for now, just using userland frontmatter -->
    <app-toc route="${globalThis.page.route}"></app-toc>
    ```
1. [x] downstream observations - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/94
    - `title` not populating as active frontmatter (is this because we set it to `null`?)
    - why do run an `unshift` on content as data in the graph? 
1. [x] realized in https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/84 that we were never checking for `activeFrontmatter` when rendering [page layout `<title>`](https://github.com/ProjectEvergreen/greenwood/pull/1266/files#diff-8e2bb543875ed80566eb644ab64798d8e0b71232f1a8cd47898877239f5f3236L196)
1. [x] prune collections serializing to only the bare essentials for CSR (`data`, `route`, ???)
1. [x] use content server port as dev server `+= 1` 
1. [x] how to sync _queries.js_ (rename to client)  host and port with greenwood config / content server
1. [x] hashing / hydration techniques from GraphQL world, e.g. APOLLO DATA + docs
1. [x] looks like import map merging was broken when no import map exists? - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/84
    - also make sure only one import map exists
    - needs to account for import map configuration now
1. [x] collection APIs `develop` specs?
1. [x] Documentation
    - rename menus -> collections
    - ~~rename `interpolateFrontmatter` -> `activeFrontmatter`~~
    - graph property changes
    - `${graph.page.data.foo}`
    - _data/client.js_
    - update GraphQL plugin README
    - external sources plugin
    - adapters
1. [x] we already get frontmatter contents in the graph phase, but we re-get it in html resource plugin
1. [x] need relaxed node modules plugin intercept patch? - seems like we're fine without, maybe landed in another commit?
1. [x] how come `Response` merging isn't working? - seems to work fine 🤷‍♂️ 
1. [x] make sure client.js and GraphQL resolvers are handling `basePath`
1. [x] ~~build graphQL resolvers off of greenwood internal functions / content API endpoint~~ - I think would rather wait until #1278 
1. [x] content as data work should probably happen in the plugin (instead of in HTML plugin)
1. [x] clean up TODOs / console logging
1. [x] ~~move active frontmatter (back) to standard HTML plugin?~~ - will leave it where it is
1. [x] handle hidden files / undefined error - https://github.com/ProjectEvergreen/greenwood/pull/1266/files#r1792691572
1. [ ] (nice to have) GraphQL plugin enhancements - this might require work (including for `contentAsData` to link queries to specific page)
    - fix apollo data bug - https://github.com/ProjectEvergreen/greenwood/issues/349
    - https://github.com/ProjectEvergreen/greenwood/issues/347

## Thoughts / Observations
1. [x] alternate namespace to `globalThis` ? - https://github.com/ProjectEvergreen/greenwood/discussions/1286
    ```html
    <title>${globalThis.page.title}</title>
    ```
1. [x] Is a content server OK for now? - https://github.com/ProjectEvergreen/greenwood/discussions/1286
    - Could we just intercept API requests instead using a resource plugin instead?
    - should we inline for production builds instead of using `fetch` requests, e.g. `<script type="application/json" GWD-DATA ...></script>`
1. [x] leave no trace, at least make it configuration option to avoid unnecessary work?  🏕️ - Yes
1. [x] I wonder if there is a case to rename `graph` to `pages` instead?  A graph implies nodes, some sort of linking structure, etc which it is definitely not in its current state. - https://github.com/ProjectEvergreen/greenwood/discussions/1286
1. [x] should we consolidate `activeFrontmatter` + `contentAsData` options?  Maybe boil it down to `activeContent`? - we shall!
1. [x] Messy / noisy output for “rich” activeFrontmatter :/ - https://github.com/ProjectEvergreen/greenwood/discussions/1286
    - can we at least prune _some_ stuff?
    - at least recommend prerender + static opt (e.g. in JS) 👇 
1. [x] #951 - will handle this as part of https://github.com/ProjectEvergreen/www.greenwoodjs.dev/issues/31
    - `prerender` caveat - only your _components_ run once, not your generic inline `<script>` tags!
    - When to use active frontmatter vs CSR + prerendering vs hydration?
    - with active frontmatter, it's there in the HTML.  With CSR prerender, no remnants left behind, a la carte via `<script type="application/json">`
    - This all feels somewhat "RSC-y" (e.g. fetch data close to your components).  I think if we can come up with some sort of hydration / islands story here, this could become a really nice pattern, at least for static sites (what happens when you're graph "gets huge" though?  Just read from the DOM to hydrate?
    - to the above point, instead of `data-gwd-opt="static"`, should we create something like `use static`?  🤔 
    - or could WCC "unwrap" a light DOM component?  (e.g. `<my-heading><h1>Hello World!</h1></my-heading>` -> `<h1>Hello World!</h1>` 🤔 🤔 
1. [x] as far as renaming things, I suppose we could potentially align with the xxxHref naming convention to - https://github.com/ProjectEvergreen/greenwood/discussions/1286
    - context - e.g. `outputDir`, `scratchDir`, `userWorkspace` and append _Href_
    - bundled resources - e.g. `optimizedFileName`, etc


packages/cli/src/lifecycles/graph.js
```diff
@@ -227,66 +206,100 @@ const generateGraph = async (compilation) => {
...
+                if (!collections[pageCollection]) {
+                  collections[pageCollection] = [];
+                }
+
+                collections[pageCollection].push(page);
+              }
+
+              compilation.collections = collections;
             } else {
               console.debug(`Unhandled extension (${extension}) for route => ${route}`);
```
thescientist13, line 304: [Noticed in this PR](https://github.com/ProjectEvergreen/greenwood/pull/1285/commits/a5ac177ca9edec948c289cec8ee19305db452cc8), we has a couple issues here
1. `route` is undefined
    ```js
    console.debug(`Unhandled extension (${extension}) for page => ${filename}`);
    ```
1. we should probably ignore hidden files
    ```sh
    building from local sources...
    { extension: '.DS_Store' }
    file:///Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lifecycles/graph.js:292
                  console.debug(`Unhandled extension (${extension}) for route => ${route}`);
    ```

PR #1269, opened 2024-08-24
feature/issue 1268 import map and attribute polyfill configuration

### LINKED ISSUE #1114: deprecate import maps shim for development
## Type of Change
Enhancement

## Summary
With both [Chrome and Firefox ](https://twitter.com/btopro/status/1616174143031885830) now natively supporting import maps, there is less of a need to autoload [**es-module-shims**](https://github.com/guybedford/es-module-shims) when running `greenwood develop`.
https://caniuse.com/?search=importmap
Screen Shot 2023-05-19 at 7 41 26 PM

## Details
So we would want to do the following:
1. [ ] Remove the shim JavaScript from being added to the page
1. [ ] Remove the logic for adding  / remove the shim attributes from `<script>` tags 
1. [ ] Remover the dependency from _package.json_
1. [ ] Update the [_How It Works_ section](https://www.greenwoodjs.io/about/how-it-works/#cli) of the docs to call out support level?
1. [ ] Could we get by supporting this if it doesn't cover Safari `15.x`?  Perhaps we could put it into our [polyfills plugin](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/plugin-polyfills)?

### LINKED ISSUE #1268: polyfills and configuration for import maps and attributes
## Summary
Shockingly, it was only just recently learned that [import attributes are not baseline across browsers](https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/57#issuecomment-2295349811) and using the `with` syntax will just break outright in anything but Chrome at the moment.
```js
import sheet from './card.css' with { type: 'css' };
```

image
image

----

It seems that we need to give the ability to backfill this capability especially if we want to promote this capability on the new website, which involve a couple things

1. Give user's the ability to opt-in to Greenwood doing the polyfilling
    ```js
    export default {
      polyfills: {
        importAttributes: ['css', 'json']
      }
    }
    ```
1. And then polyfill (technically downgrade to standard ESM)
    ```js
    // strip the offending syntax
    import sheet from './card.css';

    // and have Greenwood return a Constructable Stylesheet for card.css
    const sheet = new CSSStyleSheet();sheet.replaceSync(`p { color: red }`);
    export default sheet;
    ```
       
## Details

Interestingly, returning a `CSSStyleSheet` was our [original implementation](https://github.com/ProjectEvergreen/greenwood/pull/1215), but then we in hindsight, mistakingly went [all in on the `with` syntax](https://github.com/ProjectEvergreen/greenwood/pull/1259), not realizing it was not supported yet.  😅 

One thing to keep in mind as part of exposing this `polyfill` configuration is draw a line on what can go behind this, as it should only be a way to bridge the gap for a spec that is not baseline yet.  So it should just be based on a matter of _when_ this feature lands, not if it's still in the _if_ stage, otherwise it would be better off as a plugin.  Eventually, user's should be able to just disable the polyfill since eventually it will just be in all browsers.

----

Based on some initial testing, I noticed is that our use of [**es-module-shims**](https://github.com/guybedford/es-module-shims) is getting in the way, giving false positives in development since it seems to enable CSS and JSON modules that make these work in development for Safari / FF, which is obviously not desirable, so we should put that behind a flag as well.
Screenshot 2024-08-23 at 3 20 01 PM
Screenshot 2024-08-23 at 3 20 25 PM

So we should also disable import maps by default as well since it's now so widely available having now landing in [Safari >= 16.4](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#browser_compatibility) now, and instead should probably cancel #1115 and just create a new issue moving it to our polyfills plugin.

thescientist13:
## Related Issue
resolves #1268

## Summary of Changes
1. Add opt-in polyfill configuration for import maps (and set to off by default)
1. Add opt-in polyfill configuration for import attributes (and set to off by default)
1. Document new polyfill configurations
1. Minor cleanup

Screenshot 2024-08-24 at 1 24 27 PM
Screenshot 2024-08-24 at 1 24 42 PM

## TODO
1. [x] what about special content type handling needed for `serve` now, will this work on netlify / vercel???
    - tap into query params?
    - oh, should just _not_ externalize these instead (this was just the "old way" right?)
1. [x] testing
    - polyfill configuration options
    - import maps (`develop` + `serve` test cases)
    - import attributes (`develop` + `serve` test cases)
1. [x] add **acorn-import-attributes** as a dependency
1. [x] why were we hardcoding `text/css` in _rollup.config.js_, think we can remove make this dynamic to the type
1. [x] clean up uses of `Sec-Fetch-Dest` since I don't think it actually does anything
1. [x] Documentation
    - config option
    - Call out browser support
1. [x] Test against impacted upstream projects (website, analog studios)
    - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/87
    - https://github.com/AnalogStudiosRI/www.analogstudios.net/pull/98/commits/5b7faa3929c55e77e4241937bd78d123cdbcf6fd
1. [x] clean up demo code / _greenwood.config.js_ 

## Follow Up Issues
1. [x] Since imports maps are now in Safari >= 16.4, let's close #1114 and instead make an issue moving this behavior to the polyfills plugin
    - only for dev 
    - `type="module-shim`, `type="module"`, `es-module-shims.js`, `module-shim`
    - https://github.com/guybedford/es-module-shims?tab=readme-ov-file#browser-support
    - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type/importmap#browser_compatibility 
1. [x] discussion on dropping support for import attributes polyfill / status tracking (and back-link within docs) - https://github.com/ProjectEvergreen/greenwood/discussions/1275
1. [x] comment about these developments in https://github.com/ProjectEvergreen/greenwood/discussions/1216
1. [x] Support bundling a `<link>` tag within a shadow root? - https://github.com/ProjectEvergreen/greenwood/discussions/1276


packages/cli/src/plugins/resource/plugin-standard-javascript.js
```diff
@@ -28,6 +31,40 @@ class StandardJavaScriptResource extends ResourceInterface {
       }
     });
   }
+
+  async shouldPreIntercept(url) {
+    return url.protocol === 'file:' && this.extensions.includes(url.pathname.split('.').pop());
```
thescientist13, line 36: I think we only want to `preIntercept` if `polyfills.importAttributes` is true... could save a slight amount of processing here

PR #1285, opened 2024-10-07
feature/issue 1233 CSS Modules ™️ plugin

### LINKED ISSUE #1233: CSS Modules ™️ plugin
## Summary
The new [Greenwood website](https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/11) is piloting a plugin for supporting a very basic implementation of [CSS Modules](https://github.com/css-modules/css-modules) ™️ optimized for providing a nice co-located CSS DX in particular for Light DOM Web Components used for static templating.

Like a footer, for example
```css
.wrapper {
  text-align: center;
  border-radius: var(--radius-4);
  padding: var(--size-fluid-3);
  box-shadow: var(--shadow-4);
}
```

```js
import styles from "./footer.module.css";

export default class Footer extends HTMLElement {
  connectedCallback() {
    const year = `${new Date().getFullYear()}`;

    this.innerHTML = `
      <footer class="${styles.wrapper}">
        <h4>GreenwoodJS &copy;${year}</h4>
      </footer>
    `;
  }
}

customElements.define("app-footer", Footer);
```

Would strip out the `import` statement and add the CSS as a `<style>` tag to the HTML page, including for SSR, leaving you with something like this in your final HTML
```html
<html>
  <head>
   <style>
     .footer-aca12vdvv-wrapper {
        text-align: center;
        border-radius: var(--radius-4);
        padding: var(--size-fluid-3);
        box-shadow: var(--shadow-4);
      }
   </style>
  </head>
  <body>
    <app-footer>
      <footer class="footer-aca12vdvv-wrapper">
        <h4>GreenwoodJS &copy;2024</h4>
      </footer>
    </app-footer>
  </body>
```

> There are [some patches](https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/11/files#diff-07efaa3e5ee0019350946cd86d28956c5014e12d754ac21ac8b932fd828a405a) needed for this too, so we should try and vet those first against Greenwood's test cases.

## Details
As CSS Modules are a spec and can get quite complex, this should definitely be marked as experimental and YMMV, but did want something nice to compliment a _.css_ first authoring experience with minimal fuss / happy path only support.  The below are probably a couple things we can handle for the PR

1. Make sure this is covered as a test case - https://github.com/ProjectEvergreen/greenwood/pull/1262
1. DX considerations / prototyping carry overs
    - list unused class names
    - build race condition bug
1. documentation around only being useful for Light DOM use cases?
1. should we shim / conform to [Constructable Stylesheets API](https://web.dev/articles/css-module-scripts)?  This actually gets us the CSS rules 🤩 . 
    Screenshot 2024-01-21 at 6 13 09 PM
1. Any flexibility of the import usage API, e.g.
    ```js
    // hardcoded import
    import styles from './something.module.css'
    
    // vs
    import myCss from './something.module.css'
   ```
1. Any flexibility in the usage of the import?
    ```js
     // needs to reference the import name
     <h1 class="${styles.heading}">Welcome to Greenwood!</h1>

     // as opposed to
     const { heading } = styles;
     <h1 class="${heading}">Welcome to Greenwood!</h1>
    ```

## Discussions
And these are probably better left to putting into a discussion
1. refactor to avoid having to use intermediate output file (I think this is needed for `prerendering` because things run in a Worker Thread? 🤔 )
1. conform to [icss](https://github.com/css-modules/icss) specification (leverage postcss-modules package?)
    - Re-evaluate Lightning CSS - https://lightningcss.dev/css-modules.html
    - https://github.com/ProjectEvergreen/greenwood/discussions/763
1. Check out known issue here - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/74

thescientist13:
## Related Issue
resolves #1233 

## Summary of Changes
1. Create a CSS Modules ™️ (mostly spec compliant) plugin
1. Add to list of plugins on the website
1. Added some narrowing to Rollup bundling to only exclude JS assets from using Greenwood plugins for resolving

## TODO
1. [x] handle module shims hardcoding check
1. [x] fix windows test cases
1. [x] confirm if `response.clone()` / mergeResponse in prerender lifecycle is actually needed - yes, and we should include it anyway
1. [x] why do we need to return a CSS module via HTTP in the plugin? - ah, yeah I think we need this prerendering support 
1. [x] logging / comments / misc refactoring + cleanup
1. [x] website repo validation - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/107
1. [x] I think _graph.js_ change should go over to #1266 - https://github.com/ProjectEvergreen/greenwood/pull/1266/files#r1792691572
1. [x] we're only recursively handling _.js_ files, what about TS, etc? - https://github.com/ProjectEvergreen/greenwood/pull/1285#issuecomment-2405869888

## Nice to have
1. [x] ~~better implementation than having to write to a file on disk (_cssModulesMap.json_)~~ - the main issue here is that for SSR / prerender, we can't save to `globalThis` since memory is not shared in `Worker` threads, so will have to leave it for now
1. [ ] warn (or eliminate) unused / `undefined` styles (e.g. selector from the CSS is not found in the HTML).  maybe with option to fail on build?  otherwise by default this will just "break" the client, which is good / better I guess?
      Screenshot 2024-10-07 at 6 07 49 PM
1. [x] allow additional import identifier
    ```js
    /* works ✅ */
    import styles from './header.module.css';

    /* does not work 🚫 */
    import header from './header.module.css';
    ```
1. [x] will this handle recursive references to components called using `import` (or only top level `<script>` tags) - seems to be working just fine per test cases verification


thescientist13:
So the topic of custom formats, we are principally limited to what we can reasonably cover with acorn due to the fact that, at least as of now, our implementation here depends on us crawling through all `<script>` tags in the HTML and extracting out all the styles, and needing to do that recursively.

Unfortunately, acorn on supports walking synchronously, which is in conflict with processing all these files first using Greenwood plugins (like say we do in _rollup.config.js_) since plugins are `async`.

So for now, just doing what I did in WCC in adding [basic support for TypeScript stripping](https://github.com/ProjectEvergreen/wcc/pull/162) and called out whats supported in the plugin's README.

Perhaps there is a better implementation to be found, so definitely open to PRs that could help work around this problem, but i think this is good enough for now.  😇

PR #1291, opened 2024-11-02
docs/Update PostCSS plugin README.md to correct case sensitive name

lschierer:
The export for the @greenwood/plugin-postcss was incorrectly capitalized causing the import to fail if copied directly from the documentation.


thescientist13 (approved):
Ah, good catch.  Thanks!

Will get this out at as a patch release soon to make sure NPM docs get updated too.

PR #1300, opened 2024-11-06
enhancement/bump plugins peer dependencies to match latest minor release range

lschierer:
peer dependencies need to match dev dependencies or package managers will occasionally (pnpm will always) complain

## Related Issue
I didn't file an issue about this, because it was easy enough to fix the couple of plugins I've used so far

## Summary of Changes

I updated the peer dependency line to match the dev dependency line in the package.json for several of the plugins.  I should have gone through and done them all, but I don't currently have time.

### DIFF
--- a/packages/plugin-babel/package.json
+++ b/packages/plugin-babel/package.json
@@ -23,7 +23,7 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@greenwood/cli": "^0.4.0"
+    "@greenwood/cli": "^0.30.0"
   },
   "dependencies": {
     "@babel/core": "^7.10.4",
--- a/packages/plugin-css-modules/package.json
+++ b/packages/plugin-css-modules/package.json
@@ -22,7 +22,7 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@greenwood/cli": "^0.4.0"
+    "@greenwood/cli": "^0.30.0"
   },
   "dependencies": {
     "acorn": "^8.0.1",
--- a/packages/plugin-google-analytics/package.json
+++ b/packages/plugin-google-analytics/package.json
@@ -23,7 +23,7 @@
     "access": "public"
   },
   "peerDependencies": {
-    "@greenwood/cli": "^0.4.0"
+    "@greenwood/cli": "^0.30.0"
   },
   "devDependencies": {
     "@greenwood/cli": "^0.30.0"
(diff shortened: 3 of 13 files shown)


thescientist13 (commented):
Yeah, this is currently something lacking with our Lerna publishing setup since [Lerna doesn't handle bumping `peerDependencies`](https://github.com/lerna/lerna/issues/955) unfortunately.  (That said, I never understood why npm complains about this since even with a range of `^0.4.0` shouldn't that support any <= `1.0.0` 🤷‍♂️ )

We do try and surface this now in our [new docs](https://greenwoodjs.dev/docs/plugins/) and we also have it so the init command now drops a _.npmrc_ file file when scaffolding a new project so user's wont have to worry about the flag when starting a new project.

> This is something I've wanted to revisit in our next batch of working in the context of updating our overall monorepo / publishing hygiene (like moving off Yarn 1.x, using conventional commits, general release management enhancements, etc)

----

That all said, would you be open to making the same bumps for all of other plugins too?  Would be nice to level them all up at once until we can hopefully solve this through our publishing workflow.  🙏

lschierer:
PR updated to include more plugins per https://github.com/ProjectEvergreen/greenwood/pull/1300#pullrequestreview-2421259103

thescientist13 (approved):
Thank you for this and making all the updates.

thescientist13:
Just did a quick test and looks like the peer dependency warnings are gone now! 🎉 
```sh
➜  my-app rm .npmrc
➜  my-app npm i -D @greenwood/plugin-typescript
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated gulp-header@1.8.12: Removed event-stream from gulp-header
npm warn deprecated trim@0.0.1: Use String.prototype.trim() instead
npm warn deprecated glob@8.1.0: Glob versions prior to v9 are no longer supported
npm warn deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)

added 346 packages, and audited 347 packages in 3s

94 packages are looking for funding
  run `npm fund` for details

6 high severity vulnerabilities

To address issues that do not require attention, run:
  npm audit fix

To address all issues (including breaking changes), run:
  npm audit fix --force

Run `npm audit` for details.
```

PR #1304, opened 2024-11-07
Renaming domain name

jstockdi:
Rename links after domain name change?

## Related Issue
No issue

## Summary of Changes
Renamed domain, checked links.


thescientist13 (commented):
Yeah, I made a bunch of changes for the new domain in #1297 but opted to not change the tests since it's not public facing in any way and well, is just for the sake of testing (could have been any domain).

That said, If you want to update anything in the test cases that would be fine, but I would avoid touching any of the files already touched in my PR (like _app.html_) since you'll probably end up with a merge conflict.  👍

thescientist13:
I think this one has gone a little stale, but if still interested in fixing it up, please feel free to reopen!

Issue #1202, opened 2024-02-01
Upgrade to Node `20.x` (latest LTS)


thescientist13:
## Summary
It's that time of the release cycle again :)
https://nodejs.org/en/about/previous-releases
![Screenshot 2024-01-31 at 9 35 12 PM](https://github.com/ProjectEvergreen/greenwood/assets/895923/6eb47e84-9b77-403a-afd9-1eacf22eea07)

## Details
Node 20.x is the latest LTS, so we will want to start making the moves on upgrading CI and making sure Greenwood works well on it.  This will also be super helpful for things like #923 .

thescientist13:
Looks like Netlify can support any version? (at least as implied by the statement that it supports any version nvm can install)
https://docs.netlify.com/configure-builds/available-software-at-build-time/#languages
![Screenshot 2024-11-08 at 6 12 42 PM](https://github.com/user-attachments/assets/8fe91ee2-13e4-486c-877e-f0d8ee6408ea)

Looks like Vercel is capped  at `20.x` at this time
https://vercel.com/docs/functions/runtimes/node-js/node-js-versions
![Screenshot 2024-11-08 at 6 12 27 PM](https://github.com/user-attachments/assets/750066f7-eab3-4de9-a6cb-9ceea103c7db)

thescientist13:
Here's a current breakdown of key feature we're using and where they stand in the NodeJS release / API status

|Feature| Min. Node Version|Changelog|Notes / Status|
|--------|------------------|-----------|------|
|`import.meta.resolve` | 18.19.0, 20.6.0 | https://nodejs.org/docs/latest-v22.x/api/esm.html#importmetaresolvespecifier | Release Candidate |
| JSON import attributes | 18.20.5, 22.12.0  | https://nodejs.org/docs/latest-v22.x/api/esm.html#json-modules | Stable, awaiting a 20.x backport |
|New Loader Hooks | 20.6.0 |  https://nodejs.org/docs/latest-v22.x/api/module.html#customization-hooks | Release Candidate |

PR #1330, opened 2024-11-28
feature/issue 1328 customize Vercel adapter serverless runtime nodejs version

### LINKED ISSUE #1329: allow for customizable serverless nodejs runtime version for Vercel adapter
## Type of Change

Feature

## Summary

In the context of #1202 and that Greenwood works as expected with all current versions of NodeJS
- `18.x`
- `20.x`
- `22.x`

We should allow this configuration for the Vercel adapter, since this is a configuration item for Vercel's build output API.

## Details

I think coming out of #1326 , looks like we might need to also bump the minimum version to `20.x`.  We should also document this in plugin's README.

thescientist13:
## Related Issue

resolves #1329 

## Documentation 

N / A

## Summary of Changes

1. Add for a **runtime** option for the Vercel adapter
1. Set new default version to `nodejs20.x`
1. Document **runtime** option in the plugin's README

## TODO

1. [x] Should align with final minimum version coming out of #1202 / #1326
   - let's stick with Node 18 for the default - going to stick with `20.x` for now, actually - https://github.com/ProjectEvergreen/greenwood/pull/1330#pullrequestreview-2526726633

### DIFF
--- a/packages/plugin-adapter-vercel/README.md
+++ b/packages/plugin-adapter-vercel/README.md
@@ -61,7 +61,26 @@ export default {
 }
 ```
 
+## Options
+
+### Runtime
+
+Vercel supports [multiple semver major NodeJS versions](https://vercel.com/docs/functions/runtimes/node-js/node-js-versions#default-and-available-versions) for the serverless runtime as part of the [build output API](https://vercel.com/docs/build-output-api/v3/primitives#serverless-functions).  With the **runtime** option, you can configure your functions for any supported NodeJS version.  Current default version is `nodejs20.x`. 
+
+```javascript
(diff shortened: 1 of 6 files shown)


thescientist13 (commented):
Looks like we were able to stay on Node 18 in #1326 , so we should reset the default version back to `nodejs18.x`

thescientist13 (commented):
Based on #1372 / #1373 it looks like we can keep the minimum version for Greenwood as `18.x`, but not sure if that's useful to what people are using?  Given where NodeJS is in the LTS cycle, perhaps `20.x` makes sense, right in the middle?

Mostly just thinking of this from what makes a good default value for this plugin?  Or just always keep it aligned with Greenwood's minimum version for consistency?  🤔

PR #1332, opened 2024-11-30
enhancement/issue 1312 ESM-only PostCSS configuration file

### LINKED ISSUE #1312: ESM only configuration for PostCSS plugin
## Type of Change
Feature

## Summary

Related to #824 , it looks like PostCSS now supports [ESM configuration files](https://github.com/postcss/postcss-cli/issues/387) which means we should now be able to remove the requirement to have a CJS and an ESM configuration file.

## Details

It seems from [this comment](https://github.com/postcss/postcss-cli/issues/387#issuecomment-2346077550) it would technically have to be a _.mjs_ file but we'll take what we can get!

In addition to updating the plugin's README file, we'll also want to post a PR to the website repo to update the docs page
https://greenwoodjs.dev/docs/plugins/postcss/#installation

thescientist13:
## Related Issue

resolves #1312 

## Summary of Changes

1. Only check for an ESM based configuration file
1. Update PostCSS related dependencies
1. Update plugin README

> Breaking change will require renaming _postcss.config.mjs_ -> _postcss.config.js_

## Documentation 

1. [x] Create issue to update PostCSS website docs - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/issues/145

## TODO
1. [x] Sanity check, since in retrospect, seems like we never needed this, since we programmatically loaded the configuration file ourselves? - https://github.com/AnalogStudiosRI/www.blissri.com/pull/54
    - works just fine as is lol 😅 
1. [x] restore custom configuration docs - https://github.com/ProjectEvergreen/greenwood/pull/1332/files#r1864679805


packages/plugin-postcss/README.md
```diff
@@ -52,32 +52,6 @@ export default {
 
 ## Options
 
-### Configuration
-
-To use your own PostCSS configuration, you'll need to create _two (2)_ config files in the root of your project, by which you can provide your own custom plugins / settings that you've installed.
```
thescientist13, line 57: Whoops, still need to keep this documentation, just minus the CJS part.

lschierer:
I think the difference is that https://github.com/postcss/postcss-load-config now supports a zillion different formats but your plugin documentation never got updated when it did, as it points at a really old issue :-)

thescientist13:
Yeah, I think this was just an implementation error on our part since we don't actually rely on PostCSS to find the configuration file, we feed the options directly to PostCSS ourselves so in retrospect the _.cjs_ variant was never needed probably 🤦‍♂️ 😅

Issue #1336, opened 2024-12-02
build error with CSS style sheets imported by lit component


lschierer:
## Type of Change

Bug

## Summary

build and dev mode show different behaviors with typescript.   Something that works in dev should at least build.  

## Details

* if you check out <https://github.com/lschierer/greenwood-repoductions/tree/bare-css-typescript> (note the branch) and run ```pnpm run dev``` this project will work and display more or less as expected.   if you take the same project at the same commit and branch, and run ```pnpm run build``` it will fail saying that HTMLElement is not defined in header.ts.   This commit currently shows no errors with ```pnpm tsc```  (note that running tsc will create a bunch of .d.ts files that cause *different* errors for greenwood cli unless you then run ```find src/ -iname '*.d.ts' -delete```).   

* also using the 0.31.0 alpha.0 release, in my HP_Stuff repo that I've used in other reports, I see that the line
```
import SpectrumCard from "@spectrum-css/card/dist/index.css" with { type: "css" };
``` 
works in dev mode, but results in a module not found error at build.  I've not yet reproduced that one in the greenwood-reproductions repo because I keep hitting various other type not defined errors at build like noted just above.

thescientist13:
So there are a couple of issues here but to try and break them down

>  it will fail saying that HTMLElement is not defined in header.ts.

So the issue here is that Lit's SSR capability does not [support native `HTMLElement`](https://github.com/lit/lit/discussions/2092), which is what is happening when the `prerender: true` flag is being set in your _greenwood.config.js_.  So basically if you are using Lit + SSR, you would always have to be using `LitElement` base class.


I could have sworn I had this called out in the caveat's section but it is not there, so I have opened [a PR to document this](https://github.com/ProjectEvergreen/greenwood/pull/1337), so thanks for catching this!  👀 

----

> also using the 0.31.0 alpha.0 release, in my HP_Stuff repo that I've used in other reports, I see that the line
> `import SpectrumCard from "@spectrum-css/card/dist/index.css" with { type: "css" };`
> works in dev mode, but results in a module not found error at build. I've not yet reproduced that one in the greenwood-reproductions repo because I keep hitting various other type not defined errors at build like noted just above.

Based on the demo repo I have been validating in, I was able to confirm all the Spectrum JS and CSS imports work in both dev and prod builds and you can see a production build here hosted on Vercel.
https://greenwood-lit-ssr-yp27-git-demo-74cdbd-thescientist13s-projects.vercel.app
[...]

PR #1337, opened 2024-12-02
docs/better document and clarify Lit SSR plugin usage caveats

thescientist13:
## Related Issue

Was called out in https://github.com/ProjectEvergreen/greenwood/issues/1336 that this caveat was missing.  Could have sworn it was there 🤦 

## Documentation 

N / A

## Summary of Changes

1. Properly callout native `HTMLElement` caveat in Lit SSR
1. Share tip on Lit SSR styles work around using Raw plugin + `<style>` tag
1. Document pnpm hoisting strategy for **lit-labs** dependencies

## TODO
1. [x] Incorporate additional caveats feedback around rendering strategies
1. [x] Lit SSR styles work around (raw plugin + style tag)
1. [x] pnpm hoisting strategy
1. [x] make `HTMLElement` limitation more up-front on the website - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/147

### DIFF
--- a/packages/plugin-renderer-lit/README.md
+++ b/packages/plugin-renderer-lit/README.md
@@ -41,6 +41,7 @@ $ pnpm add -D @greenwood/plugin-renderer-lit
 1. Please familiarize yourself with some of the [caveats](https://lit.dev/docs/ssr/overview/#library-status) called out in the Lit docs, like:
     - Lit SSR [**only** renders into declarative shadow roots](https://github.com/lit/lit/issues/3080#issuecomment-1165158794), so you will have to keep browser support and polyfill usage in mind.
     - At this time, `LitElement` does not support `async` work.  You can follow along with this issue [in the Lit repo](https://github.com/lit/lit/issues/2469).
+1. Lit SSR does [not support native `HTMLElement`](https://github.com/lit/lit/discussions/2092) so you will need to use `LitElement` as your base class in all instances.
 1. Lit only supports templates on the server side for HTML generated content, thus Greenwood's `getBody` API must be used.  We would love for [server only components](https://github.com/lit/lit/issues/2469#issuecomment-1759583861) to be a thing though!
 1. Lit does not support [`CSSStyleSheet` (aka CSS Modules) in their SSR DOM shim](https://github.com/lit/lit/issues/2631#issuecomment-1065400805).
 1. Full hydration support is not available yet.  See [this Greenwood issue](https://github.com/ProjectEvergreen/greenwood/issues/880) to follow along with when it will land.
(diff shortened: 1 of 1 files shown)


lschierer:
I'm partly confused about what is possible because <https://greenwoodjs.dev/docs/plugins/lit-ssr/> links to <https://greenwoodjs.dev/docs/pages/server-rendering/> which has a number of examples using html components, but no mention on either page that these examples won't work with the lit plugin enabled (or they will, but only with the right combination of options?).

lschierer:
Per the discussion in discord, a caveat that explained that the examples using HTMLElements would need a search/replace substitution to use LitElement would suffice to prevent nearly all the confusion I experienced and resolve the other comments I've left on this PR.   The caveat would, unfortunately, probably need to be on *both* the website on the /docs/plugins/lit-ssr/ page and in the plugin's own readme (or overly literal people like me are likely to miss it).

thescientist13:
Yeah, good call.  I don't expect us to maintain docs examples for every possible WC based library that could be used with Greenwood, Lit was just the main one that also provides good SSR support, but we can at least make this more obvious both from the Lit SSR plugin docs page and the README for sure to make sure developers know they can _only_ use `LitElement` based classes when opting in to Lit+SSR

lschierer:
No, you certainly don't want to try to have a caveat that either encompasses all possible web components nor one per web component.  That's why I was envisioning the caveat being on the two lit specific documentation pages.

thescientist13 (commented):
Hey @lschierer 

Thanks for all the good feedback here, and I have now updated the plugin's README with following additional call-outs:
- Move `HTMLElement` to be the first bullet of caveats section
- Suggest hoisting patterns for PNPM users
- Update GitHub issue link for `CSSStyleSheets` limitation and provide link to alternative (interim) work around

I also opened a PR to the website docs to also make the `HTMLElement` caveat more up front - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/147

----

Also, the observation of having multiple `prerender` options between Greenwood config and this plugin is indeed confusion, so I think now with this clearer caveat around `HTMLElement` / `LitElement`, I think it would just make sense to remove the `prerender` option from this plugin, and have it just be set from _greenwood.config.js_, so I opened an issue for that here and will open a standalone PR for that change - https://github.com/ProjectEvergreen/greenwood/issues/1343 .

----

Thanks again for all your suggestions and feedback and let me know what you think of these changes.  👍

lschierer:
awesome, looks good :-)

thescientist13:
Great, appreciate the extra set of eyes on it.  👀

PR #1342, opened 2024-12-11
bug/issue 1338 handle bundling of bare CSS `@import` specifiers

### LINKED ISSUE #1338: bare CSS `@import` specifiers are not resolving and bundling correctly
## Type of Change

Bug

## Summary

Observed as part of testing for #1309 that when we try and leverage Spectrum's CSS files through an `@import`, e.g.
```css
/* src/styles/theme.css */
@import url('../../node_modules/@spectrum-web-components/styles/all-large-dark.css');
```

This file will load
Screenshot 2024-12-02 at 5 05 27 PM

BUT, it will lead to errors on other files it references
Screenshot 2024-12-02 at 5 05 14 PM

## Details

In theory I thought this was covered by https://github.com/ProjectEvergreen/greenwood/issues/1257 but I guess we are not recursively handling `@import` statements within `@import` statements?  Here's what the file looks like from unpkg
https://unpkg.com/browse/@spectrum-web-components/styles@1.0.2/all-large-dark.css

Perhaps this is specific to `@import` in _node_modules_ specifically though?  Perhaps we should do something similar to #684 and prefix things as such?
```css
/* src/styles/theme.css */
@import url('~@spectrum-web-components/styles/all-large-dark.css');
```

Or check if things do / don't have a `./` or `../` and automatically resolve that way?

thescientist13:
## Related Issue

resolves #1338 

## Documentation 

N / A

## Summary of Changes

1. Resolve and bundle bare CSS `@import` specifiers
1. Add test case

## TODO
1. [x] should probably rename url variables for better naming for better clarity (e.g. `url` -> `sourceUrl`)
1. [x] rebase after #1341 
1. [x] remove demo code
1. [ ] (nice to have or make a standalone issue) wondering if we could use this technique to handle this case for `url(...)` - https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/plugins/resource/plugin-standard-css.js#L75
    - will make an issue once I put together a minimal reproduction for the issue

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-standard-css.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-css.js
@@ -31,7 +31,7 @@ function bundleCss(body, sourceUrl, compilation, workingUrl) {
           if (value.indexOf('.') === 0 || value.indexOf('/node_modules') === 0) {
             const resolvedUrl = value.startsWith('/node_modules')
               ? new URL(getResolvedHrefFromPathnameShortcut(value, projectDirectory))
-              : new URL(value, url);
+              : new URL(value, sourceUrl);
 
             const importContents = fs.readFileSync(resolvedUrl, 'utf-8');
 
--- a/packages/cli/src/plugins/resource/plugin-standard-css.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-css.js
@@ -11,7 +11,7 @@ import { ResourceInterface } from '../../lib/resource-interface.js';
 import { hashString } from '../../lib/hashing-utils.js';
 import { getResolvedHrefFromPathnameShortcut } from '../../lib/node-modules-utils.js';
 
-function bundleCss(body, url, compilation, resolvedUrl) {
+function bundleCss(body, sourceUrl, compilation, workingUrl) {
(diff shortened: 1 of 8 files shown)


packages/cli/src/plugins/resource/plugin-standard-css.js
```diff
@@ -26,13 +26,23 @@ function bundleCss(body, url, compilation) {
...
         const { value } = node;
 
-        if (value.indexOf('.') === 0 || value.indexOf('/node_modules') === 0) {
-          const resolvedUrl = value.startsWith('/node_modules')
-            ? new URL(`.${value}`, projectDirectory)
-            : new URL(value, url);
-          const importContents = fs.readFileSync(resolvedUrl, 'utf-8');
+        if (!value.startsWith('http')) {
+          if (value.indexOf('.') === 0 || value.indexOf('/node_modules') === 0) {
+            const resolvedUrl = value.startsWith('/node_modules')
```
thescientist13, line 31: this will need to align with https://github.com/ProjectEvergreen/greenwood/pull/1341/files#diff-e20a60e7b88e17250746babe9a6f80b88def96d34e3901541b6ced7dd668b378R32

PR #1358, opened 2024-12-22
bug/issue 1354 support isomorphic data client APIs without `prerender` configuration option

### LINKED ISSUE #1354: data client not working on production builds (when not using `prerender` option)
## Type of Change

Bug

## Summary

While working on [this project](https://github.com/thescientist13/HP_Stuff/pull/1) realized that there was a significant oversight in how the data client would work.  It technically won't work as is without `prerender` enabled for a production build, since without prerendering, the custom JSON files per query will not be generated, leading to 404s in a production build
image

- For now have captured this in the discussion here - https://github.com/ProjectEvergreen/greenwood/discussions/1286
- Also, realized the docs don't capture this nuance well either - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/154

## Details

It seems for now we can get away with making a non-prerender fallback for the data client, where if prerendering was _not_ used, we can just `fetch` _graph.json_ and do the filtering client side.

thescientist13:
## Related Issue

resolves #1354

## Documentation 

https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/154

## Summary of Changes

1. Add "fallback" support for using the isomorphic data client APIs without setting the `prerender` flag
1. Add some test cases for content as data `<script>` setup blocks

### DIFF
--- a/packages/cli/src/data/client.js
+++ b/packages/cli/src/data/client.js
@@ -1,13 +1,32 @@
+import { filterContentByCollection, filterContentByRoute } from '@greenwood/cli/src/lib/content-utils.js';
+
 const CONTENT_STATE = globalThis.__CONTENT_AS_DATA_STATE__ ?? false; // eslint-disable-line no-underscore-dangle
-const PORT = globalThis?.__CONTENT_SERVER__?.PORT ?? 1984; // eslint-disable-line no-underscore-dangle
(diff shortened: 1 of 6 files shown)


thescientist13 (commented):
Hmm, looks like this is failing for SSR pages... (no `CONTENT_STATE` detected)
Screenshot 2024-12-27 at 1 41 40 PM

Issue #1327, opened 2024-11-24
refresh TypeScript plugin configuration and dependency management


thescientist13:
## Type of Change

Enhancement

## Summary

Its been a couple of years since we originally authored the [TypeScript plugin](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-typescript/) and so after all this time, it probably warrants a refresh on some of its assumptions around configuration and dependency management

## Details

Things we should probably refresh / refactor
1. Default compiler configuration (based on recent projects
    - https://github.com/thescientist13/greenwood-lit-ssr/pull/25)
    - https://twitter.com/mattpocockuk/status/1819660693794111660
    - import attributes - https://github.com/thescientist13/greenwood-lit-ssr/pull/24/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0
    ```json
    {
      "compilerOptions": {
        "target": "es2022",
        "module": "es2022",
        "lib": ["es2022", "DOM", "DOM.Iterable", "esnext"],
        "allowSyntheticDefaultImports": true,
        "moduleResolution": "node",
        "allowImportingTsExtensions": true,
        "noEmit": true
      },
    }
    ```
1.  Should probably make `tsc` a `peerDependency` so users can manage their own versioning (and make it a dev dependency for plugin dev)
1.  Documentation on how to best setup type checking and error handling / logging for development and CI - https://github.com/ProjectEvergreen/greenwood/issues/658
1. Recommendations for running `tsc` in CI

----

Projects for reference
- https://github.com/AnalogStudiosRI/www.analogstudios.net
- https://github.com/AnalogStudiosRI/api

Issue #1296, opened 2024-11-04
Typescript support for External Source Plugins


lschierer:
## Type of Change
New Feature

## Summary
Please support using TypeScript to write an External Source Plugin 

## Details

I attempted to follow https://www.greenwoodjs.io/plugins/source/ changing the .js extension out for .ts and get an ERR_UNKNOWN_FILE_EXTENSION that persists even after I was assisted in solving these errors for other sections of my configuration [here](https://github.com/ProjectEvergreen/greenwood/discussions/1292).   

I suspect that its actually possible to do this, but that it would be (currently) a multi-step process, where I'd need to transpile the plugin to javascript to then be able to run the greenwood develop or build commands.  This request is to support, and document, a configuration that will consolidate that into a single command line that could be included as the develop script in a package.json file.  

I further suspect that this is related to  [#1250](https://github.com/ProjectEvergreen/greenwood/issues/1250) where in the comments I requested support for the main configuration file to be supported as a typescript file.  My intuition is that if you solved that, this would be mostly solved as well.

thescientist13:
> I further suspect that this is related to https://github.com/ProjectEvergreen/greenwood/issues/1250 where in the comments I requested support for the main configuration file to be supported as a typescript file. My intuition is that if you solved that, this would be mostly solved as well.

Yeah, pretty much.  Right now the TypeScript plugin is only geared towards transforming "web" code, e.g. anything in your _src/_ directory, but not really anything else.

Let me align this with #1250 though since I think this would be something we would apply to all plugins at large.  My guess is that this would likely try and tackle this in two phases:
1. Support exposing JS Docs style types (#1250)
1. Have Greenwood support TypeScript files first party for its APIs (would probably re-purpose this issue for that effort)

I've also been experimenting with Bun and Deno and they are working pretty well out of the box with some Greenwood demo apps as part of initial testing, and so now with NodeJS also supporting TS, I may even considering moving TS support right into Greenwood itself, probably with a flag though.  We shall see!

Issue #658, opened 2021-07-08
improve TypeScript plugin compilation and error handling


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Coming out of #650 , wanted to track a couple thoughts / suggestions that would be helpful to get right for our TypeScript plugin.

This was based on adding a type to the _www/_ project like so
```js
// src/utils/greeting.d.ts
```

and using it like so
```js
// src/components/greeting.ts
import { html, css, LitElement, customElement, property } from 'lit-element';

@customElement('app-greeting')
export class GreetingComponent extends LitElement {
  static styles = css`p { color: blue }`;

  @property()
  name = 'Somebody';

  render() {
    const greeting: Greeting = {
      message: html`<p>Hello, ${this.name}!</p>`
    };

    return greeting.message;
  }
}
```

And the following _tsconfig.json_
```json
{
  "compilerOptions": {
    "experimentalDecorators": true,
    "noImplicitAny": true
  }
}
```

### VS Code Integration
If I just use the above TSConfig, I get an warnings about `Greeting`
![Screen Shot 2021-07-08 at 12 40 49 PM](https://user-images.githubusercontent.com/895923/124959922-c1feb000-dfe9-11eb-8974-400bf8bd6f73.png)

If I add something like `includes` to _tsconfig.json_ I get more errors :/
```json
{
  "include": [
    "./www/**/**/*.d.ts"
  ],
  "compilerOptions": {
    "experimentalDecorators": true,
    "noImplicitAny": true
  }
}
```

![Screen Shot 2021-07-08 at 12 42 06 PM](https://user-images.githubusercontent.com/895923/124960109-f96d5c80-dfe9-11eb-9954-dedc1bc8c7ba.png)

### Rollup TypeScript Plugin
While observing an issue with the [`[hash]` value in filenames being different between Windows and Linux / macOS](https://github.com/ProjectEvergreen/greenwood/pull/650#issuecomment-877614947), I found there is a newer contender for a TS plugin for Rollup, a bit slower, but maybe it will fix this issue and help solve for the compilation errors we need, below?  👇 
https://github.com/ezolenko/rollup-plugin-typescript2

[...]

PR #1420, opened 2025-02-14
feature/issue 1250 Greenwood types

### LINKED ISSUE #1250: typing support for public capabilities of Greenwood
## Summary

While Greenwood is a build tool without much of a public API other than its CLI and plugins, there are some areas that should probably provide typing to users through JSDocs, maybe TS but I'm not super keen on TypeScript in libraries.

## Details
Most obvious ones that come to mind are:

1. [_greenwood.config.js_](https://greenwoodjs.dev/docs/reference/configuration/), even if just [JS Docs style](https://nextjs.org/docs/pages/building-your-application/configuring/typescript#type-checking-nextconfigjs), ex.
    ```js
    /** @type {import('@greenwood/cli').GreenwoodConfig} */
    export default {
      /* config options here */
    }
    ```
1. [Plugins API](https://greenwoodjs.dev/docs/reference/plugins-api/) (and get rid of that silly `ResourceInterface` lol)
    - https://github.com/ProjectEvergreen/greenwood/issues/1296
1. [Content as Data](https://greenwoodjs.dev/docs/content-as-data/) - like the data client
1. others?

We would also want to make sure we validate these through CI somehow, ideally using these config files in our own projects and / or test cases.

----

Raised in https://github.com/ProjectEvergreen/greenwood/discussions/1292#discussioncomment-11147829, a good point was raised about _authoring_ Greenwood plugins (and probably also _greenwood.config.js_ in TypeScript.  Might need to make another issue for that once we can land this one first, unless its easy to do it all in one go with something like [**sucrase**](https://www.npmjs.com/package/sucrase), which we are using now in WCC.

thescientist13:
## Related Issue

resolves #1250 

## Documentation 

1. [ ] TODO
    - need to document all new typings and usage; Configuration, Content as Data, compilation, Plugin APIs (theme packs?).  (create a page in the Reference docs?)
    - no more references to `ServerInterface` / `ResourceInterface`

## Summary of Changes

1. Created declaration files for Greenwood's
    - configuration
    - plugins
    - compilation
    - content as data
1. Added declaration exports for the CLI and all plugin packages
1. Refactored / removed `ServerInteface` / `ResourceInterface`
1. Added `tsc` for type checking to CI
1. Standardize all module related fields for all workspace _package.json_ files

> ℹ️ **Note**
> - There will be a standalone PR to update the TypeScript _plugin
>    - https://github.com/ProjectEvergreen/greenwood/issues/1327
>    - https://github.com/ProjectEvergreen/greenwood/issues/658
> - As for authoring Greenwood APIs in TypeScript, follow along with https://github.com/ProjectEvergreen/greenwood/issues/1296

## TODO
1. [ ] Types for SSR APIs - `getBody`, `getLayout`, `getFrontmatter`
1. [ ] Types for GraphQL APIs
1. [ ] Update all plugin READMEs for types
    - CSS Modules declaration type
1. [ ] Contributing docs update?

## Nice to have / new issues

1. [ ] handle all cases of `// @ts-nocheck` (or make as a good first issue)
1. [ ] we're way behind on our node-html-parser version
1. [ ] copy over this config to WCC
1. [ ] JSDocs for the entire project?
    - `noImplicitAny`, etc
1. [ ] do we need rollup babel plugin?


packages/cli/src/lib/walker-package-ranger.js
```diff
@@ -257,9 +257,10 @@ async function walkPackageJson(packageJson = {}, walkedPackages = new Set()) {
         const resolvedRoot = derivePackageRoot(resolved);
 
         if (resolvedRoot) {
-          const resolvedPackageJson = (
-            await import(new URL("./package.json", resolvedRoot), { with: { type: "json" } })
-          ).default;
+          const resolvedPackageJson =
+            // @ts-expect-error see https://github.com/microsoft/TypeScript/issues/42866
+            (await import(new URL("./package.json", resolvedRoot), { with: { type: "json" } }))
```
lschierer, line 262: a .toString() here avoids the TS error about URLs instead of string

lschierer: This comment repeats in a few other places.  It's a very slight hack to do it that way, but avoiding the ts-expect-error prevents the disable from masking other potential errors on the same line.

thescientist13: good to know, but I would prefer to not have to alter code just to fit TS, especially when TS is wrong (in this case 😄 )

packages/cli/src/types/api.d.ts
lschierer, line 1: Interesting that these aren't generated files?  aren't .d.ts files normally outputs of a build, not part of the repository?

lschierer: Is this because all of your source files are actually javascript, and you do not have typescript to generate them from?  Would you be open to defining types in something like Zod if I took these .d.ts and created a series of Zod types that represented them? Obviously not intendeding to block this PR with that.

lschierer (commented):
minor notes as I was curious about what it took to do this and why it was labeled as a "breaking" change.

packages/cli/src/types/api.d.ts
thescientist13: Yes, it's because Greenwood's source code is (and will be) JavaScript, so the types are hand made in this case.

> Would you be open to defining types in something like Zod if I took these .d.ts and created a series of Zod types that represented them? 

I'm not really sure what that would look like, but I'm not sure how Zod would / does fit into this at the framework level though or why we would want to couple to anything more than we would be with TS / JSDocs.  

But if you want to start a thread in the #typescript channel of the Discord to chat about it, I would be open to that.  

(FYI I planned to share a more formal thread in the Discord on all the TS PRs in a couple days once I finished a couple more things around this)

thescientist13:
@lschierer 
> and why it was labeled as a "breaking" change.

Because I added an `exports` field in each _package.json_ for the CLI and all plugins, which technically alters the resolution logic used by Node (and most tools in the space) so generally a good practice to highlight this, though from an end user perspective, should be no observable difference.

PR #1471, opened 2025-03-26
Allow source plugins to access existing graph

### LINKED ISSUE #1470: Allow source plugins to access existing graph
### Current State

Source plugins are not able to access graph generated from files.

### Desired State

Source plugins could access existing graph, allowing them to use pages as data when creating other pages.

For example, I create a page for each tag defined in frontmatter:
https://github.com/EskiMojo14/neocities/blob/main/src/greenwood/tags.ts
https://eskimojo.neocities.org/packages/
https://eskimojo.neocities.org/packages/tags/testing/

### Additional Context

_No response_

EskiMojo14:
## Related Issue

resolves #1470 

## Summary of Changes

- Moved assignment of graph into compilation *before* source plugins run, meaning plugins can access existing graph.

### DIFF
--- a/packages/cli/src/lifecycles/graph.js
+++ b/packages/cli/src/lifecycles/graph.js
@@ -360,6 +360,9 @@ const generateGraph = async (compilation) => {
 
       const sourcePlugins = compilation.config.plugins.filter((plugin) => plugin.type === "source");
 
+      compilation.graph = graph;
+      compilation.manifest = { apis };
+
       if (sourcePlugins.length > 0) {
         console.debug("building from external sources...");
         for (const plugin of sourcePlugins) {
@@ -385,9 +388,6 @@ const generateGraph = async (compilation) => {
           }
         }
       }
-
-      compilation.graph = graph;
-      compilation.manifest = { apis };
 
       resolve(compilation);
     } catch (err) {


packages/cli/src/lifecycles/graph.js
```diff
@@ -360,6 +360,9 @@ const generateGraph = async (compilation) => {
 
       const sourcePlugins = compilation.config.plugins.filter((plugin) => plugin.type === "source");
 
+      compilation.graph = graph;
```
thescientist13, line 363: maybe just worth leaving a small comment here re: order matters?

thescientist13 (commented):
Cool, this looks pretty good at first glance.  Let me get all the tests running and assuming they all pass, then this should be good to merge.

EskiMojo14:
thanks! yeah, some help with tests would be appreciated, the strategy in this repo is pretty unfamiliar for me

thescientist13:
So each folder in _packages/*/src/test/cases/_ is basically just a mini Greenwood project, with the general strategy being to test that (for example) if I have a _src/pages/index.html_, that after running Greenwood, I should have a _public/index.html_.  Most of our validation usually just comes down to using **glob-fs** to verify the expected number of output files were created, or for testing the actual content of those pages, just reading the contents of that file and putting the HTML content into something like JSDOM, and than making sure the contents are expected.  (like querying for an `<h1>` tag and checking `textContent`).

So for this particular change, I would probably recommend altering this test case
https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.plugins.source/

Should be able to just add an additional entry here that pulls some data from the compilation, which you can put into the `body` contents of the page, then you can test that the generated page contains that content.  You can also alter the _src/_ directory to add a similar use case like you did that uses some content from frontmatter.
https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.plugins.source/greenwood.config.js#L6

For some [testing](https://github.com/ProjectEvergreen/greenwood/blob/master/CONTRIBUTING.md#testing), here are some some [tips](https://github.com/ProjectEvergreen/greenwood/blob/master/CONTRIBUTING.md#notes)
- Use `describe.only` for isolating that test case - https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.plugins.source/build.plugins-source.spec.js#L46
- Enabling debug output for the test command runner `new Runner(true)` - https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.plugins.source/build.plugins-source.spec.js#L57 
- Comment out the public output cleanup function to help debug the generated output - https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.plugins.source/build.plugins-source.spec.js#L147

Hope that helps!

packages/cli/test/cases/build.plugins.source/build.plugins-source.spec.js
```diff
@@ -54,7 +55,7 @@ describe("Build Greenwood With: ", function () {
     this.context = {
       publicDir,
     };
-    runner = new Runner();
+    runner = new Runner(true);
```
thescientist13, line 58: not a big deal but we can remove the debug output when done.

thescientist13 (approved):
Thanks!  I made some tweaks in these files as well for another PR so apologies on that.  I don't mind handling the merge conflict though unless you feel like giving it a try, but otherwise this is looking good to go!

Issue #1424, opened 2025-02-27
Built-In TypeScript support


thescientist13:
#### Overview

So I had some time to play around with what first party TypeScript support might look like in Greenwood, and I think there might be a good path forward here, in particular now that NodeJS provides [built-in support for type stripping (and some transformations)](https://nodejs.org/en/learn/typescript/run-natively#running-typescript-code-with-nodejs).  With this, a lot of the transformation work can be handled for us for free from the runtime itself.  (and would also apply to Deno / Bun if and when we get to them!)

```sh
$ NODE_OPTIONS='--experimental-strip-types' greenwood <command>
```

In particular, now that the ecosystem has come to a good place around having a distinction of type _stripping_ (which makes live reload / HMR / general DX much faster by not having to send everything through the TypeScript compiler) vs type _checking_ (which is what actually provides the red squiggles in your IDE), combined with more support from all (server) JS runtimes, I think the cost bringing this in first party becomes a very small effort for the value.

And just as importantly, users of Greenwood are very much [interested in using TypeScript](https://github.com/ProjectEvergreen/greenwood/discussions/1422), so it felt like it was really worth re-evaluating our original decision around externalizing this capability in the first place.  We've already started [adding types for Greenwood](https://github.com/ProjectEvergreen/greenwood/issues/1250), and make it all "first party" would also help [simplify usage / expectations around using TypeScript with Greenwood](https://github.com/ProjectEvergreen/greenwood/discussions/1292#discussioncomment-11147829).

#### NodeJS Support

As far as [NodeJS support](https://nodejs.org/api/typescript.html#type-stripping) goes, it breaks downs like so:

- v22.6.0 (LTS) - `--experimental-strip-types` flag was added
- v22.7.0 (LTS) - `--experimental-transform-type` flag was added
- v23.0.0 (Non-LTS) - `--experimental-strip-types` unflagged

The recommended compiler config is:
```json
{
  "compilerOptions": {
    "target": "ES2020",
    "module": "preserve",
    "moduleResolution": "bundler",
    "allowImportingTsExtensions": true,
    "noEmit": true,
    "verbatimModuleSyntax": "false"
  }
}
```

We might also want to recommend TypeScript >=5.8 with [`erasableSyntaxOnly`](https://egghead.io/use-the-erasable-syntax-only-type-script-compilation-flag~8m293).

[...]

Issue #1248, opened 2024-06-21
dynamic layouts for SSR


thescientist13:
## Summary
Coming out of #955 / #1212 , one follow up feature to track is supporting dynamic layouts during SSR.  So while this has been possible
```sh
src/
  layouts/
    app.js
    blog.js
```

The layout was executed only once at bundle time which means it was only a one time render.
https://github.com/ProjectEvergreen/greenwood/blob/v0.30.0-alpha.2/packages/cli/src/lifecycles/bundle.js#L243

## Details
Main trick is that I think this this will require bundling layouts into SSR pages like we do the body.  Not sure how tricky this will be though.  🤔 

----

Additionally, would be nice to support dynamic context layouts too, as currently they are hardcoded to only support HTML.
https://github.com/ProjectEvergreen/greenwood/blob/v0.30.0-alpha.2/packages/cli/src/lib/templating-utils.js#L13

lschierer:
if I'm doing things correctly (never a sure thing) then dynamic layouts partly work.  with a dynamic page, I got the dynamic layout working by importing the getLayout function into the dynamic page then re-exporting in the same object as getFrontmatter():

```
import { getLayout } from "../../../layouts/standard.ts";
export { getFrontmatter, getLayout };
```

however either in the dynamic page or in a markdown page, using the "layout" attribute of the frontmatter to refer to the dynamic layout file appears to be nonfunctional.

thescientist13:
A minimal reproduction would help but if I am understanding you correctly, doing this
https://greenwoodjs.dev/docs/resources/markdown/#layouts
```md
---
layout: blog
---

# My First Blog Post

This is my first blog post, I hope you like it!
```

Will map to a matching _filename_ in the layouts _directory_, e.g.
```sh
src/
  layouts/
    blog.html # or blog.js
```

Looking at the docs, I can see how this is not as clear as it could be, so if that's what you're referring to, I can definitely make that clarification as part of https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/160

lschierer:
I have layouts/blog.ts instead of .js

thescientist13:
Ah, I see now.

Yeah, in that case that is just something we need to implement, see #1395 .

(FWIW, the next batch of ecosystem compat work will primarily focus around TypeScript enhancements for the project, so expect a lot of those related issues to get resolved as part of the 0.32.0 release which should be starting development soon, so hang tight! 🙏 )

Issue #1395, opened 2025-01-23
support custom resource plugins / formats for layouts


thescientist13:
## Type of Change

Feature

## Summary

Realized that coming out of https://github.com/ProjectEvergreen/greenwood/issues/955 /  we overlooked supporting custom formats for layouts.

```sh
src/
  layouts/
    blog.ts
```

## Details

I'm not sure if the issue was that split off https://github.com/ProjectEvergreen/greenwood/issues/1248 and this got lost in the shuffle, but we should definitely apply the same logic as https://github.com/ProjectEvergreen/greenwood/pull/1265 for layouts as well.

lschierer:
would the changes in alpha.5 have addressed this esssentially as a side effect?

thescientist13:
Going to de-scope this for the current release in favor of at least achieving parity with having TS based layouts, per https://github.com/ProjectEvergreen/greenwood/pull/1474

thescientist13:
Since the example above (_src/layouts/blog.ts_) is already supported now per #1474 , let me update the example to something not already supported out of the box, and then push this out, since I think with _.js_ and _.ts_ supported as extensions, that should get us pretty far.

PR #1473, opened 2025-03-28
Feat typescript layouts

lschierer:
## Related Issue

This partially addresses #1248 
this fixes a gap in #1424

## Documentation 

I didn't change any of the documentation. 

## Summary of Changes

1. [x] Allow users to have Server rendered layouts using Web Components in TypeScript files 
1. [x] This checks for .ts, .mts, .mjs and .js file extensions for layouts. 
1. [x] I only changed packages/cli/src/lib/layout-utils.js, #1424 seemed to indicate a different file would be required.  

I tested this in <https://github.com/lschierer/GreenwoodSpectrumTheme> by coping the modified file into place.


thescientist13:
Yeah, I was working on #1395 and came to a similar conclusion of let's just focus on filling the gap around TS based layouts at least, since the more I got into it, the more I realized it might require a bit more work, and the logic around app / page layouts is a bit clunky at the moment.

I got at least app layouts working and then got a bit side tracked on page layouts over something that I think needs better documenting on the website.  (that a _src/pages/index.html_ will ignore a _src/layouts/page.html_ but will honor a _src/layouts/app.html).

Let me get my PR up with what I have so far since it includes some tests as well and let me get back to just fast tracking TS based page layouts, and then let me reconcile that against what is here just to make sure I'm not forgetting anything and then we can go from there.

lschierer:
sure, I just happened to notice the hard coded .js at line 51 when I was trying to understand something that I was in fact doing wrong (I remember that I learned something, I forget what).  So I thought, this looked like low hanging fruit, and it might save some time for yall.  Or you'd tell me I missed something and I'd learn from that.

thescientist13:
Thanks, yeah good looking out!

Yeah, I deferred the full scope of #1395 for now and so I think https://github.com/ProjectEvergreen/greenwood/pull/1474 should cover the basics for now, at least
- _app.ts_
- _page.ts_
- _{custom-layout}.ts_

A couple things I plan to look into in the next release line though, just as a heads up to you
- #1475 
- #1476

Issue #71, opened 2019-05-01
Have develop command pick next randomly available port


thescientist13:
## Type of Change
- [x] New Feature Request
- [ ] Documentation / Website
- [ ] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
By default, instead of hardcoding the port for `devServer` it might be really nice to have it just pick the next available port on the user's local machine?

## Details

Something similar to  [**webpack-dev-server**](https://stackoverflow.com/questions/42055201/webpack-dev-server-dynamic-port)

> _Also, if we enable this, we might / should be able to run all our tests in parallel?  Doesn't have to be in this PR, but a good fast follow if so!_

thescientist13:
I think we can probably get that output as part of running **webpack-dev-server** maybe? 

Also, probably for a related issue, would be adding `open` as default?

hutchgrant:
To my knowledge, that output doesn't show by default. We're showing it on purpose after build because otherwise user would have no idea. `open` didn't work in my OS, under every test I ran with it.  Yet, it does work in CRA so I don't know.

thescientist13:
ah, interesting.  Maybe in most other projects I'm always using the CLI, so always passed the flag that way
```
$ webpack-dev-server --open
```

hutchgrant:
yea might be the difference between webpack node api and cli. There are small things such as that which are different.

thescientist13:
I wonder if [this trick](https://twitter.com/matteocollina/status/1255888871684014080) would work?
>  would not hardcode the port and set it to zero so it is dynamically allocated, and then start the client and communicate to it. It's the only sane way I've found to avoid flaky tests.

PR #1482, opened 2025-04-04
enhancement/issue 1481 pre clean build output directories

### LINKED ISSUE #1481: pre-clean all build output directories
### Current State

Currently, Greenwood has two build output related directories when running development and build tasks
- _.greenwood/_ ([scratch directory](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/context.js#L11)) - Primarily used for pre-rendered / pre-built bundles but before final bundling and any optimizations needed before going into the output directory
- _public/_ ([output directory](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/context.js#L12)) - The final build output directory 

In particular the scratch directory can often be used during development runs by plugins or the CLI for misc operations, like with CSS Modules plugin
https://github.com/ProjectEvergreen/greenwood/blob/master/packages/plugin-css-modules/src/index.js#L180

The main issue is that artifacts from previous runs / build could get built up over time and potentially leak from the _.greenwood/_ folder into the build output folder, leading to unexpected deployment artifacts.  In fact, in pretty much every project I have, I always add a `clean` task that runs rimraf over these directories anyway
- https://github.com/ProjectEvergreen/www.greenwoodjs.dev/blob/main/package.json#L15
- https://github.com/AnalogStudiosRI/www.analogstudios.net/blob/main/package.json#L23
- https://github.com/thescientist13/greenwood-lit-ssr/blob/main/package.json#L23

### Desired State

Coming out of this, it would be nice if the following was automatically done by Greenwood
1. Clean the scratch directory
1. Clean the output directory
1. Clean the output directory for adapter plugins
1. (nice to have) could we "push" the scratch directory down into _node_modules_ instead of having it put into the user's root directory?

### Additional Context

_No response_

thescientist13:
## Related Issue

resolves #1481 

## Documentation 

N / A

## Summary of Changes

1. Pre-clean (and consolidate logic for) output and scratch directories
1. Pre-clean adapters build output
1. Optimize GraphQL query caching for build command only
1. Optimize CSS Modules scratch output

## TODO
1. [x] ~~(nice to have / new issue) output scratch directory within node modules~~ - deferred for now https://github.com/ProjectEvergreen/greenwood/pull/1482#issuecomment-2905643391
    - pre-test in a project 
    - would need / want to align with docs and init .gitignore

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   },
   "scripts": {
     "lerna": "lerna",
-    "clean": "rimraf --glob ./**/.greenwood/** ./**/public/** ./coverage",
+    "clean": "rimraf --glob ./packages/**/.greenwood/** ./packages/**/public/** ./coverage",
     "clean:deps": "rimraf **/node_modules/**",
     "build": "cross-env __GWD_ROLLUP_MODE__=strict node . build",
     "serve": "node . serve",
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -1,8 +1,6 @@
 import { bundleCompilation } from "../lifecycles/bundle.js";
-import { checkResourceExists } from "../lib/resource-utils.js";
 import { copyAssets } from "../lifecycles/copy.js";
 import { getDevServer } from "../lifecycles/serve.js";
-import fs from "fs/promises";
 import {
   preRenderCompilationWorker,
   preRenderCompilationCustom,
@@ -11,19 +9,12 @@ import {
 
 const runProductionBuild = async (compilation) => {
   const { prerender, activeContent, plugins } = compilation.config;
-  const outputDir = compilation.context.outputDir;
   const prerenderPlugin = compilation.config.plugins.find((plugin) => plugin.type === "renderer")
     ? compilation.config.plugins.find((plugin) => plugin.type === "renderer").provider(compilation)
     : {};
   const adapterPlugin = compilation.config.plugins.find((plugin) => plugin.type === "adapter")
     ? compilation.config.plugins.find((plugin) => plugin.type === "adapter").provider(compilation)
     : null;
-
-  if (!(await checkResourceExists(outputDir))) {
-    await fs.mkdir(outputDir, {
-      recursive: true,
-    });
-  }
 
   if (prerender) {
     // start any of the user's server plugins if needed
--- a/packages/cli/src/lifecycles/bundle.js
+++ b/packages/cli/src/lifecycles/bundle.js
@@ -391,7 +391,9 @@ async function bundleSsrPages(compilation, optimizePlugins) {
         });
       }
 
-      // better way to write out this inline code?
+      // would be nice to find out a better way to write out / generate this inline code "facade"
+      // and how to calculate the relative path to the src/ page's entry point in the user's workspace
+      // https://github.com/ProjectEvergreen/greenwood/pull/1482#issuecomment-2905643391
       /* eslint-disable no-useless-escape */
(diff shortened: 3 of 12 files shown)


thescientist13:
So one slight complication with trying to "tuck away" the scratch (_.greenwood/_) directory is due to how bundle SSR pages.  Basically to bundle them with all the correct information, we have to product a custom "entry" file that itself effectively `import`s the user's actual page file.
https://github.com/ProjectEvergreen/greenwood/blob/v0.32.0/packages/cli/src/lifecycles/bundle.js#L394


This file is put into the _.greenwood/_ directory, and has to relatively find the user's actual SSR page in their workspace.  Here is a simplified example of what that output would look
```js
// .greenwood/artists.js
import { executeRouteModule } from "/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-renderer-lit/src/execute-route-module.js";

const moduleUrl = new URL("../src/pages/artists.js", import.meta.url);

export async function handler(request) {
  const compilation = JSON.parse(`{"graph": [ /* ... */ ]} config: {}`);
  const page = JSON.parse(`{/* ... */}`);
  const data = await executeRouteModule({ moduleUrl, compilation, page, request });
  let staticHtml = `<!DOCTYPE html>
        
        <! -- -->
     </html>
    `;

  if (data.body) {
    staticHtml = staticHtml.replace(/<content-outlet>(.*)<\/content-outlet>/s, data.body);
  }

  return new Response(staticHtml, {
    headers: {
      "Content-Type": "text/html",
    },
  });
}
```

In particular, this line has to be relative from the location of the user's workspace
```js
const moduleUrl = new URL("../src/pages/artists.js", import.meta.url);
```

So now that the location has moved to "inside" the package, it should really be calculated somehow, as opposed to how we do it now 😅 
```js
for (const page of ssrPages) {
  const { id, route, pageHref } = page;
  const pagePath = new URL(pageHref).pathname.replace(pagesDir.pathname, "./");
  const entryFileUrl = new URL(pageHref);
  const entryFileOutputUrl = new URL(
    `file://${entryFileUrl.pathname.replace(pagesDir.pathname, scratchDir.pathname)}`,
  );

  const outputPathRootUrl = new URL(`file://${path.dirname(entryFileOutputUrl.pathname)}/`);
  const pagesPathDiff = context.pagesDir.pathname.replace(
    context.projectDirectory.pathname,
    "",
  );
  const relativeDepth = "../".repeat(pagePath.split("/").length - 1);

  // ...

  // better way to write out this inline code?
  /* eslint-disable no-useless-escape */
  await fs.writeFile(
    entryFileOutputUrl,
    `
    import { executeRouteModule } from '${normalizePathnameForWindows(executeModuleUrl)}';

    const moduleUrl = new URL('${relativeDepth}${pagesPathDiff}${pagePath.replace("./", "")}', import.meta.url);

    //...
    `
  )
}
```

----

So let me play around with it for a bit, and if nothing materializes after a bit, I will have to eave a comment to handle the tucking away at a later time.  Is there a reason we can't just pass `pageHref`?  🤔 
```js
  await fs.writeFile(
    entryFileOutputUrl,
    `
    import { executeRouteModule } from '${normalizePathnameForWindows(executeModuleUrl)}';

    const moduleUrl = new URL('${pageHref}');

    //...
    `
  )
```

One thing to consider though, is that for all test cases, the _.greenwood/_ folder will be shared for all runs now, which if after solving #71  (which would in theory allow our tests run in parallel), that sharing could become a liability.  (but only for our monorepo, to be clear).

PR #1535, opened 2025-07-15
enhancement/issue 991 reduce SSR page instantiations

### LINKED ISSUE #991: eliminate SSR page instantiation during graph lifecycle
## Overview

In order to build a [graph of all pages](https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/graph.js), Greenwood gets the metadata of SSR based pages, calling functions [like `getFrontmatter`](https://www.greenwoodjs.io/docs/server-rendering/#api) or reading page level configurations like `prerender` or `isolation`.

However the SSR route worker is not efficient as it could be since, it will just automatically run all the things, including just blindly generating the page using WCC
```js
async function executeRouteModule({ modulePath, compilation, route, label, id, prerender, htmlContents, scripts }) {
  const parsedCompilation = JSON.parse(compilation);
  const data = {
    template: null,
    body: null,
    frontmatter: null,
    html: null
  };

  if (prerender) {
    ...
  } else {
    const module = await import(pathToFileURL(modulePath)).then(module => module);
    const { getTemplate = null, getBody = null, getFrontmatter = null } = module;

    if (module.default) {
      const { html } = await renderToString(pathToFileURL(modulePath));

      data.body = html;
    } else {
      if (getBody) {
        data.body = await getBody(parsedCompilation, route);
      }
    }

    if (getTemplate) {
      data.template = await getTemplate(parsedCompilation, route);
    }

    if (getFrontmatter) {
      data.frontmatter = await getFrontmatter(parsedCompilation, route, label, id);
    }
  }

  parentPort.postMessage(data);
}

executeRouteModule(workerData);
```

## Details
Basically, all your pages will get alllllllll those functions called, making your API requests, etc.  Just for building the graph, nothing other than frontmatter is needed.  All this stuff will get built anyone during the `build` so this is just duplicate / wasted effort.

This is related to some other issues as well
- I imagine this will help contribute to improving #970
- I think we might want to rethink templates and pages and maybe clean up some of these APIs - #955
- Might make sense to do in tandem with #988 

How do we account for options coming out of a page though?
https://github.com/ProjectEvergreen/greenwood/pull/1206/
```js
export const isolation = true;
```
----

So we'll definitely want to think of a way to refactor this so `develop` is a no-op, and this only happens once on `build`.

Further thinking, I think this could / should apply anywhere we invoke a Worker thread for this kind of work, and so instead of invoking a module with a worker and automatically calling _all_ of its exports (body, layout, etc) by default, we might want to explicitly pass in what we _want_, e.g.
```js
// before
worker.postMessage({
  executeModuleUrl: routeWorkerUrl.href,
  moduleUrl: filenameUrl.href,
  compilation: JSON.stringify(compilation),
  page: JSON.stringify({
    servePage: isCustom,
    route,
    root,
    label,
  }),
  request,
});

// after
worker.postMessage({
  executeModuleUrl: routeWorkerUrl.href,
  moduleUrl: filenameUrl.href,
  compilation: JSON.stringify(compilation),
  page: JSON.stringify({
    servePage: isCustom,
    route,
    root,
    label,
  }),
  {
    frontmatter: true,
    layout: false,
    body: false,
  }
  request,
});
```

thescientist13:
## Related Issue

resolves #991 

## Documentation 

1. [x] Document new conventions for [renderer plugins](https://greenwoodjs.dev/docs/reference/plugins-api/#renderer) and [constructor props for dynamic layouts](https://greenwoodjs.dev/docs/pages/layouts/#server-rendering) - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/issues/224

## Summary of Changes

1. Introduce `contentOptions` argument for _execute-route-module.js_ and integrate with call sites in the lifecycles and plugins of the CLI
1. Update Lit Renderer Plugin

## TODO
1. [x] Get CLI test cases passing
1. [x] why does `getLayout` display `undefined` for the HTML in our test page? (and add test case if applicable)
    - related to this? - https://github.com/ProjectEvergreen/greenwood/pull/1535/files#r2211946240
    <details>
      Screenshot 2025-07-16 at 9 20 40 PM
    </details>
1. [x] update Lit SSR render plugin
1. [x] better way to check for default export as a page vs a dynamic layout? - https://github.com/ProjectEvergreen/greenwood/pull/1535/files#diff-00c80dffc8c29f51577f933b981748aff2f0f1bcaed2d38aad9a4b324f9cf93bR47
1. [x] confirm this is what we want - https://github.com/ProjectEvergreen/greenwood/pull/1535/files#r2211946240
1. [x] address / clean up `TODO` references - `https://github.com/ProjectEvergreen/greenwood/issues/991`
1. [x] add test case for dynamic layout constructor props
1. [x] any types that need to be updated?
1. [x] clean up test code / restore _greenwood.config.js_


packages/cli/src/lib/layout-utils.js
```diff
@@ -357,9 +363,9 @@ async function getAppLayout(pageLayoutContents, compilation, matchingRoute) {
       const worker = new Worker(new URL("./ssr-route-worker.js", import.meta.url));
 
       worker.on("message", (result) => {
-        if (result.body) {
-          dynamicAppLayoutContents = result.body;
-        }
+        // if (result.layout) {
+        dynamicAppLayoutContents = result.body ?? result.layout;
```
thescientist13, line 367: should this just use `result.layout`?

PR #1543, opened 2025-08-02
enhancement/issue 1379 stabilize nodejs version 22.18.0

### LINKED ISSUE #1379: stabilize to a single minumum NodeJS LTS version
## Type of Change

Enhancement

## Summary

Coming out of #1202 / #1372, we opened up the range for NodeJS LTS version to try and accommodate all minimum versions with the least amount of experimental features [based on the current compatibility matrix at the time](https://github.com/ProjectEvergreen/greenwood/issues/1202#issuecomment-2566656021)
```json
"engines": {
  "node": "^18.20.5 || ^20.10.0 || ^22.12.0"
}
```

Per research into opening #1372 , here's a current breakdown of key feature we're using and where they stand in the NodeJS release / API status

|Feature| Min. Node Version|Changelog|Notes / Status|
|--------|------------------|-----------|------|
|`import.meta.resolve` | 18.19.0, 20.6.0 | https://nodejs.org/docs/latest-v22.x/api/esm.html#importmetaresolvespecifier | Release Candidate |
| JSON import attributes | 18.20.5, 20.10.0*, 22.12.0  | https://nodejs.org/docs/latest-v22.x/api/esm.html#json-modules | Stable, [* awaiting a 20.x backport, though](https://github.com/nodejs/node/pull/55961) |
|New Loader Hooks | 20.6.0 |  https://nodejs.org/docs/latest-v22.x/api/module.html#customization-hooks | Release Candidate |

## Details

Prior to 1.0, would like to go back to a single minimum version, using `>=` so we don't have to constantly be chasing the LTS release schedule, e.g.

The main items we're waiting on
- Stable `import.meta.resolve` - currently RC as of `20.18.x`
- Stable JSON Import Attributes - https://github.com/nodejs/node/pull/55961
- Stable loader hooks * - currently RC as of `20.18.x`

I think eventually we can go to something like this for the engines field
```json
"engines": {
  "node": ">=20.?.0"
}
```

thescientist13:
## Related Issue

resolves #1379 

## Documentation 

1. [x] document recommended minimum version of NodeJS - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/issues/226

## Summary of Changes

1. Update GitHub Actions, _.nvmrc_, engines, etc
1. Removed strip types flag from init TS scaffolding starter
1. Adopt [**amaro**](https://www.npmjs.com/package/amaro) for internal type-stripping in Greenwood (instead of **sucrase**)

## TODO
1. [x] 🚨 Need to wait for GitHub Actions to support latest version
    - just waiting on windows now - https://github.com/ProjectEvergreen/greenwood/pull/1543#pullrequestreview-3102316695
1. [x] Greenwood <> Amaro (swap out sucrase)
1. [x] Check adapter <> Node versions compat
    - not much we can do about `24.x` support at this time - #1533 
1. [x] WCC <> Amaro
    - I don't think, since we need Sucrase for JSX too
1. [x] review if `case "CIRCULAR_DEPENDENCY":` in Rollup is still needed
    - I think this is fine for now
1. [x] See if we can start using [`fs.cp` + w / recursive flag](https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisescpsrc-dest-options) - that was easy, we're already using it!  😅 
1. [ ] Update PR checks to capture latest workflows after merge

## Nice to have / own issues
1. [x] See if we can start using [`fs.glob` (still experimental)](https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisescpsrc-dest-options) - https://github.com/ProjectEvergreen/greenwood/issues/1551
    - https://github.com/ProjectEvergreen/greenwood/pull/1543#issuecomment-3169540821
1. [x] Any chance [URL Pattern](https://nodejs.org/api/url.html#class-urlpattern) is available in Node 22 now? (needed for #882)
    - nope, not yet, only in `>= 23.8.0`
1. [x] ~~Also, coming out of #1519, we could also consider enforcing our ESLint rule for node prefixes to match a NodeJS version.~~
    - I think it's fine, since we can only specify a single specific version and no ranges.  But our test coverage should keep us honest here
1. [x] We can probably adopt [`parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig
) for the CLI and init packages now, so should make an issue - https://github.com/ProjectEvergreen/greenwood/issues/1552
    - that said, it doesn't handle the "nice" things like a description, wonder if we could / should just roll that on our own?
1. [ ] GitHub Actions still has a versioned ubuntu image


thescientist13:
Just checked and GitHub Actions are still on `22.17.x`
https://github.com/ProjectEvergreen/greenwood/actions/runs/16737214358/job/47378407422
Screenshot 2025-08-04 at 8 20 33 PM

thescientist13 (commented):
we're so back 🚀 
https://github.com/ProjectEvergreen/greenwood/actions/runs/16842547704/job/47716432732
Screenshot 2025-08-08 at 7 32 30 PM

thescientist13:
For an example of how to do the `fs.glob` conversion
```js
// before
it("should contain one javascript file in the output directory", async function () {
  expect(await glob.promise(path.join(this.context.publicDir, "*.js"))).to.have.lengthOf(1);
});
```

```js
// after
it("should contain one javascript file in the output directory", async function () {
  const files = await Array.fromAsync(fs.promises.glob(
    "*.js",
    { cwd: new URL(".", import.meta.url)},
  ));

  expect(files.length).to.equal(1);
});
```

Issue #823, opened 2021-12-18
make Greenwood CLI and plugins `async` through and through


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Coming out of #532 , where I had tried to do some [`async` operations in the `provider` of a plugin](https://github.com/ProjectEvergreen/greenwood/commit/16c6694420821dbe2a31be25a20946579900f21d), that it dawned on me that should we really be better committed to, and consistently, implementing `async` in Greenwood.

## Details
The main issue is that to do something like this, _everything_ has to become `async`, which is fine, but we just have to design around it so that we can support `async` and non-async functions alike, with one API.  This is because `await` will automatically wrap return values in `Promise`, so in that way `async` is backwards compatible, unlike a sync version

This would also be a good time to fix "faux" async where we use it inside for `forEach` and `reduce` functions, which when actually forced into an `async` context, fall apart, since those Array methods are sync.  We should probably use more `for ... in`.

Bonus, the `async` variants of `fs` methods can take an instance of `URL` natively, so we wouldn't have to use `pathToFileURL` in those situation.

thescientist13:
Another case for why this would be useful - https://github.com/ProjectEvergreen/greenwood/pull/1331

KaiPrince:
Starting this work in https://github.com/ProjectEvergreen/greenwood/pull/1546. You can assign it to me.

thescientist13:
I think another issue coming out of this work is that the website project now seems to show a max listeners warning - https://github.com/ProjectEvergreen/www.greenwoodjs.dev/pull/211#issuecomment-3263954548

Tracking this new issue in #970

PR #1545, opened 2025-08-04
chore: refactor async callbacks

KaiPrince:
## Related Issue

https://github.com/ProjectEvergreen/greenwood/issues/823

## Documentation 

N/A

## Summary of Changes

1. wrap async `map` callbacks with sync function to return array of pending promises instead of making the map itself run asynchronously
2. convert async `reduce` callbacks to `for..in` loops
3. remove unnecessary `Promise.resolve` calls

### DIFF
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -38,13 +38,7 @@ const runProductionBuild = async (compilation) => {
       });
     }
 
-    await Promise.all(
-      servers.map(async (server) => {
-        await server.start();
-
-        return Promise.resolve(server);
-      }),
-    );
+    await Promise.all(servers.map((server) => server.start()));
 
     if (prerenderPlugin.executeModuleUrl) {
       await preRenderCompilationWorker(compilation, prerenderPlugin);
--- a/packages/cli/src/lifecycles/bundle.js
+++ b/packages/cli/src/lifecycles/bundle.js
@@ -123,36 +123,40 @@ async function optimizeStaticPages(compilation, plugins) {
           (page.isSSR && page.prerender) ||
           (page.isSSR && compilation.config.prerender),
       )
-      .map(async (page) => {
-        const { route, outputHref } = page;
-        const outputDirUrl = new URL(outputHref.replace("index.html", "").replace("404.html", ""));
-        const url = new URL(`http://localhost:${compilation.config.port}${route}`);
-        const contents = await fs.readFile(
-          new URL(`./${outputHref.replace(outputDir.href, "")}`, scratchDir),
-          "utf-8",
-        );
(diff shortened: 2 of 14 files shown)


packages/cli/src/commands/build.js
KaiPrince, line 1: I recommend hiding whitespace changes for easier reviewing

packages/cli/src/lifecycles/bundle.js
```diff
@@ -202,84 +206,58 @@ async function bundleStyleResources(compilation, resourcePlugins) {
...
-            intermediateResponse.clone(),
-            currentResponse.clone(),
-          );
+          const mergedResponse = mergeResponse(response.clone(), currentResponse.clone());
 
           if (mergedResponse.headers.get("Content-Type").indexOf(contentType) >= 0) {
-            return Promise.resolve(mergedResponse.clone());
+            response = mergedResponse.clone();
           }
         }
```
KaiPrince, line 221: Each stage is pretty much the same, so a future refactoring might be to extract each of the stages and let them share some code through a common function. It would also reduce the size of this massive function.

packages/cli/src/lifecycles/serve.js
```diff
@@ -69,9 +69,8 @@ async function getDevServer(compilation) {
       for (const plugin of resourcePlugins) {
         if (plugin.shouldServe && (await plugin.shouldServe(url, request))) {
           const current = await plugin.serve(url, request);
-          const merged = mergeResponse(response.clone(), current.clone());
 
-          response = merged.clone();
+          response = mergeResponse(response.clone(), current.clone());
```
KaiPrince, line 73: I notice that some of these use `current.clone()` and some don't. Is that intentional, or should they all be the same way? I've left it as is for now, but I can make them all the same while I'm here.

packages/plugin-renderer-puppeteer/src/puppeteer-handler.js
```diff
@@ -45,8 +47,7 @@ export default async function (compilation, callback) {
     console.info(
       "For more information please see this guide - https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md",
     );
-
-    return Promise.reject();
+    throw e;
```
KaiPrince, line 50: This one is technically a change, because you can't throw nothing syntactically in JS, but I think rejecting with the actual error instead of nothing is better here anyways.

test/smoke-test.js
```diff
@@ -269,7 +269,7 @@ function serve(label) {
 }
 
 async function runSmokeTest(testCases, label) {
-  testCases.forEach(async (testCase) => {
+  testCases.forEach((testCase) => {
```
KaiPrince, line 272: There are no `await` statements in the callback so I just removed the `async`.

Issue #970, opened 2022-08-05
performance benchmarking / optimizing Greenwood


thescientist13:
## Type of Change
- [ ] New Feature Request
- [ ] Documentation / Website
- [x] Improvement / Suggestion
- [ ] Bug
- [ ] Other (please clarify below)

## Summary
Can't say I've focused to heavily on build time performances but since Zach Leatherman of 11ty put together a bench mark [blog post](https://www.zachleat.com/web/build-benchmark/) and [repo](https://github.com/zachleat/bench-framework-markdown) of some of the top SSGs, and so I figured I would test Greenwood just at least get a sense of where it sat amongst the rest.

I made [a fork](https://github.com/thescientist13/bench-framework-markdown) and decided to add Greenwood.  😬 🤞 

I tried it on two machines:
- MBP 2016 (Intel) - Interestingly enough, had no problem chruning through all 4000 pages, albeit it took a while 🐢 😁 
- MBA 2020 (M1) - Had some OOO memory issues sometimes when doing 50 / 100 pages intermittently 😞 
    <details>
      <pre>
          % sh ./bench-greenwood.sh 100
          [bench] Clean complete
          [bench] Setup complete 100 pages
          -------------------------------------------------------
          Welcome to Greenwood (v0.26.0) ♻️
          -------------------------------------------------------
          Initializing project config
          Initializing project workspace contexts
          Generating graph of workspace files...
          building from local sources...
          Running Greenwood with the build command.
          pages to generate 
           /
           /posts/ad-deserunt-cillum-consectetur-occaecat copy/
           /posts/ad-deserunt-cillum-consectetur-occaecat/
           /posts/ad-in-id-ex-sunt copy/
           /posts/ad-in-id-ex-sunt/
           /posts/ad-in-laborum-anim-ex copy/
           /posts/ad-in-laborum-anim-ex/
           /posts/ad-nulla-tempor-ipsum-nostrud copy/
           /posts/ad-nulla-tempor-ipsum-nostrud/
           /posts/adipisicing-dolore-ea-do-magna copy/
           /posts/adipisicing-dolore-ea-do-magna/
           /posts/adipisicing-dolore-ullamco-in-amet copy/
           /posts/adipisicing-dolore-ullamco-in-amet/
           /posts/adipisicing-exercitation-do-nisi-consectetur copy/
           /posts/adipisicing-exercitation-do-nisi-consectetur/
[...]

PR #1546, opened 2025-08-04
Enhancement/issue 823 ensure greenwood cli and plugins are favoring idiomatic usages of async

KaiPrince:
## Related Issue

https://github.com/ProjectEvergreen/greenwood/issues/823

## Documentation 

N/A

## Summary of Changes

1. wrap async `map` callbacks with sync function to return array of pending promises instead of making the map itself run asynchronously
2. convert async `reduce` callbacks to `for..in` loops
3. remove unnecessary `Promise.resolve` calls

### DIFF
--- a/packages/cli/src/commands/build.js
+++ b/packages/cli/src/commands/build.js
@@ -38,13 +38,7 @@ const runProductionBuild = async (compilation) => {
       });
     }
 
-    await Promise.all(
-      servers.map(async (server) => {
-        await server.start();
-
-        return Promise.resolve(server);
-      }),
-    );
+    await Promise.all(servers.map((server) => server.start()));
 
     if (prerenderPlugin.executeModuleUrl) {
       await preRenderCompilationWorker(compilation, prerenderPlugin);
--- a/packages/cli/src/lifecycles/bundle.js
+++ b/packages/cli/src/lifecycles/bundle.js
@@ -123,36 +123,40 @@ async function optimizeStaticPages(compilation, plugins) {
           (page.isSSR && page.prerender) ||
           (page.isSSR && compilation.config.prerender),
       )
-      .map(async (page) => {
-        const { route, outputHref } = page;
-        const outputDirUrl = new URL(outputHref.replace("index.html", "").replace("404.html", ""));
-        const url = new URL(`http://localhost:${compilation.config.port}${route}`);
-        const contents = await fs.readFile(
-          new URL(`./${outputHref.replace(outputDir.href, "")}`, scratchDir),
-          "utf-8",
-        );
-        const headers = new Headers({ "Content-Type": "text/html" });
-        let response = new Response(contents, { headers });
+      .map((page) =>
+        (async () => {
+          const { route, outputHref } = page;
+          const outputDirUrl = new URL(
+            outputHref.replace("index.html", "").replace("404.html", ""),
+          );
+          const url = new URL(`http://localhost:${compilation.config.port}${route}`);
+          const contents = await fs.readFile(
+            new URL(`./${outputHref.replace(outputDir.href, "")}`, scratchDir),
+            "utf-8",
+          );
+          const headers = new Headers({ "Content-Type": "text/html" });
+          let response = new Response(contents, { headers });
 
-        if (!(await checkResourceExists(outputDirUrl))) {
-          await fs.mkdir(outputDirUrl, {
-            recursive: true,
-          });
-        }
-
-        for (const plugin of plugins) {
-          if (plugin.shouldOptimize && (await plugin.shouldOptimize(url, response.clone()))) {
-            const currentResponse = await plugin.optimize(url, response.clone());
-
-            response = mergeResponse(response.clone(), currentResponse.clone());
+          if (!(await checkResourceExists(outputDirUrl))) {
+            await fs.mkdir(outputDirUrl, {
+              recursive: true,
+            });
           }
-        }
 
-        // clean up optimization markers
-        const body = (await response.text()).replace(/data-gwd-opt=".*?[a-z]"/g, "");
+          for (const plugin of plugins) {
+            if (plugin.shouldOptimize && (await plugin.shouldOptimize(url, response.clone()))) {
+              const currentResponse = await plugin.optimize(url, response.clone());
 
-        await fs.writeFile(new URL(outputHref), body);
-      }),
+              response = mergeResponse(response.clone(), currentResponse.clone());
+            }
+          }
+
+          // clean up optimization markers
+          const body = (await response.text()).replace(/data-gwd-opt=".*?[a-z]"/g, "");
+
+          await fs.writeFile(new URL(outputHref), body);
+        })(),
+      ),
   );
 }
 
@@ -202,84 +206,58 @@ async function bundleStyleResources(compilation, resourcePlugins) {
       const request = new Request(url, { headers });
       const initResponse = new Response(contents, { headers });
 
-      let response = await resourcePlugins.reduce(async (responsePromise, plugin) => {
-        const intermediateResponse = await responsePromise;
+      let response = initResponse;
+
+      for (const plugin in resourcePlugins) {
         const shouldServe = plugin.shouldServe && (await plugin.shouldServe(url, request));
 
         if (shouldServe) {
           const currentResponse = await plugin.serve(url, request);
-          const mergedResponse = mergeResponse(
-            intermediateResponse.clone(),
-            currentResponse.clone(),
-          );
+          const mergedResponse = mergeResponse(response.clone(), currentResponse.clone());
 
           if (mergedResponse.headers.get("Content-Type").indexOf(contentType) >= 0) {
-            return Promise.resolve(mergedResponse.clone());
+            response = mergedResponse.clone();
           }
         }
+      }
 
-        return Promise.resolve(responsePromise);
-      }, Promise.resolve(initResponse));
-
-      response = await resourcePlugins.reduce(async (responsePromise, plugin) => {
-        const intermediateResponse = await responsePromise;
+      for (const plugin in resourcePlugins) {
         const shouldPreIntercept =
           plugin.shouldPreIntercept &&
-          (await plugin.shouldPreIntercept(url, request, intermediateResponse.clone()));
+          (await plugin.shouldPreIntercept(url, request, response.clone()));
 
         if (shouldPreIntercept) {
-          const currentResponse = await plugin.preIntercept(
-            url,
-            request,
-            intermediateResponse.clone(),
-          );
-          const mergedResponse = mergeResponse(
-            intermediateResponse.clone(),
-            currentResponse.clone(),
-          );
+          const currentResponse = await plugin.preIntercept(url, request, response.clone());
+          const mergedResponse = mergeResponse(response.clone(), currentResponse.clone());
 
           if (mergedResponse.headers.get("Content-Type").indexOf(contentType) >= 0) {
-            return Promise.resolve(mergedResponse.clone());
+            response = mergedResponse.clone();
           }
         }
+      }
 
-        return Promise.resolve(responsePromise);
-      }, Promise.resolve(response.clone()));
-
-      response = await resourcePlugins.reduce(async (responsePromise, plugin) => {
-        const intermediateResponse = await responsePromise;
+      for (const plugin in resourcePlugins) {
         const shouldIntercept =
-          plugin.shouldIntercept &&
-          (await plugin.shouldIntercept(url, request, intermediateResponse.clone()));
+          plugin.shouldIntercept && (await plugin.shouldIntercept(url, request, response.clone()));
 
         if (shouldIntercept) {
-          const currentResponse = await plugin.intercept(
-            url,
-            request,
-            intermediateResponse.clone(),
-          );
-          const mergedResponse = mergeResponse(
-            intermediateResponse.clone(),
-            currentResponse.clone(),
-          );
+          const currentResponse = await plugin.intercept(url, request, response.clone());
+          const mergedResponse = mergeResponse(response.clone(), currentResponse.clone());
 
           if (mergedResponse.headers.get("Content-Type").indexOf(contentType) >= 0) {
-            return Promise.resolve(mergedResponse.clone());
+            response = mergedResponse.clone();
           }
         }
+      }
 
-        return Promise.resolve(responsePromise);
-      }, Promise.resolve(response.clone()));
-
-      response = await resourcePlugins.reduce(async (responsePromise, plugin) => {
-        const intermediateResponse = await responsePromise;
+      for (const plugin in resourcePlugins) {
         const shouldOptimize =
-          plugin.shouldOptimize && (await plugin.shouldOptimize(url, intermediateResponse.clone()));
+          plugin.shouldOptimize && (await plugin.shouldOptimize(url, response.clone()));
 
-        return shouldOptimize
-          ? Promise.resolve(await plugin.optimize(url, intermediateResponse.clone()))
-          : Promise.resolve(responsePromise);
-      }, Promise.resolve(response.clone()));
+        if (shouldOptimize) {
+          response = await plugin.optimize(url, response.clone());
+        }
+      }
 
       optimizedFileContents = await response.text();
 
--- a/packages/cli/src/lifecycles/config.js
+++ b/packages/cli/src/lifecycles/config.js
@@ -15,20 +15,24 @@ const greenwoodPlugins = (
       new URL("./renderer/", greenwoodPluginsDirectoryUrl),
       new URL("./resource/", greenwoodPluginsDirectoryUrl),
       new URL("./server/", greenwoodPluginsDirectoryUrl),
-    ].map(async (pluginDirectoryUrl) => {
-      const files = await fs.readdir(pluginDirectoryUrl);
+    ].map((pluginDirectoryUrl) =>
+      (async () => {
+        const files = await fs.readdir(pluginDirectoryUrl);
 
-      return await Promise.all(
-        files.map(async (file) => {
-          const importUrl = new URL(`./${file}`, pluginDirectoryUrl);
-          // @ts-expect-error see https://github.com/microsoft/TypeScript/issues/42866
-          const pluginImport = await import(importUrl);
-          const plugin = pluginImport[Object.keys(pluginImport)[0]];
+        return await Promise.all(
+          files.map((file) =>
+            (async () => {
+              const importUrl = new URL(`./${file}`, pluginDirectoryUrl);
+              // @ts-expect-error see https://github.com/microsoft/TypeScript/issues/42866
+              const pluginImport = await import(importUrl);
+              const plugin = pluginImport[Object.keys(pluginImport)[0]];
 
-          return Array.isArray(plugin) ? plugin : [plugin];
-        }),
-      );
-    }),
+              return Array.isArray(plugin) ? plugin : [plugin];
+            })(),
+          ),
+        );
+      })(),
+    ),
   )
 )
   .flat(PLUGINS_FLATTENED_DEPTH)
--- a/packages/cli/src/lifecycles/config.js
+++ b/packages/cli/src/lifecycles/config.js
@@ -129,13 +129,13 @@ const readAndMergeConfig = async () => {
     // workspace validation
     if (workspace) {
       if (!(workspace instanceof URL)) {
-        return Promise.reject("Configuration error: workspace must be an instance of URL");
+        throw new Error("Configuration error: workspace must be an instance of URL");
       }
 
       if (await checkResourceExists(workspace)) {
         customConfig.workspace = workspace;
       } else {
-        return Promise.reject(
+        throw new Error(
           "Configuration error: Workspace doesn't exist! Please double check your configuration.",
         );
       }
@@ -147,14 +147,14 @@ const readAndMergeConfig = async () => {
     ) {
       customConfig.optimization = optimization;
     } else if (optimization) {
-      return Promise.reject(
+      throw new Error(
         `Configuration error: provided optimization "${optimization}" is not supported.  Please use one of: ${optimizations.join(", ")}.`,
       );
     }
 
     if (activeContent) {
       if (typeof activeContent !== "boolean") {
-        return Promise.reject("Configuration error: activeContent must be a boolean");
+        throw new Error("Configuration error: activeContent must be a boolean");
       }
       customConfig.activeContent = activeContent;
     }
@@ -164,7 +164,7 @@ const readAndMergeConfig = async () => {
 
       flattened.forEach((plugin) => {
         if (!plugin.type || pluginTypes.indexOf(plugin.type) < 0) {
-          return Promise.reject(
+          throw new Error(
             `Configuration error: plugins must be one of type "${pluginTypes.join(", ")}". got "${plugin.type}" instead.`,
           );
         }
@@ -172,7 +172,7 @@ const readAndMergeConfig = async () => {
         if (!plugin.provider || typeof plugin.provider !== "function") {
           const providerTypeof = typeof plugin.provider;
 
-          return Promise.reject(
+          throw new Error(
             `Configuration error: plugins provider must be a function. got ${providerTypeof} instead.`,
           );
         }
@@ -180,7 +180,7 @@ const readAndMergeConfig = async () => {
         if (!plugin.name || typeof plugin.name !== "string") {
           const nameTypeof = typeof plugin.name;
 
-          return Promise.reject(
+          throw new Error(
             `Configuration error: plugins must have a name. got ${nameTypeof} instead.`,
           );
         }
@@ -208,7 +208,7 @@ const readAndMergeConfig = async () => {
         if (typeof devServer.hud === "boolean") {
           customConfig.devServer.hud = devServer.hud;
         } else {
-          return Promise.reject(
+          throw new Error(
             `Configuration error: devServer hud options must be a boolean.  Passed value was: ${devServer.hud}`,
           );
         }
@@ -216,7 +216,7 @@ const readAndMergeConfig = async () => {
 
       if (devServer.port) {
         if (!Number.isInteger(devServer.port)) {
-          return Promise.reject(
+          throw new Error(
             `Configuration error: devServer port must be an integer.  Passed value was: ${devServer.port}`,
           );
         } else {
@@ -232,7 +232,7 @@ const readAndMergeConfig = async () => {
         if (Array.isArray(devServer.extensions)) {
           customConfig.devServer.extensions = devServer.extensions;
         } else {
-          return Promise.reject(
+          throw new Error(
             "Configuration error: provided extensions is not an array.  Please provide an array like ['txt', 'foo']",
           );
         }
@@ -246,9 +246,7 @@ const readAndMergeConfig = async () => {
 
     if (port) {
       if (!Number.isInteger(port)) {
-        return Promise.reject(
-          `Configuration error: port must be an integer.  Passed value was: ${port}`,
-        );
+        throw new Error(`Configuration error: port must be an integer.  Passed value was: ${port}`);
       } else {
         customConfig.port = port;
       }
@@ -256,7 +254,7 @@ const readAndMergeConfig = async () => {
 
     if (basePath) {
       if (typeof basePath !== "string") {
-        return Promise.reject(
+        throw new Error(
           `Configuration error: basePath must be a string.  Passed value was: ${basePath}`,
         );
       } else {
@@ -267,7 +265,7 @@ const readAndMergeConfig = async () => {
     if (pagesDirectory && typeof pagesDirectory === "string") {
       customConfig.pagesDirectory = pagesDirectory;
     } else if (pagesDirectory) {
-      return Promise.reject(
+      throw new Error(
         `Configuration error: provided pagesDirectory "${pagesDirectory}" is not supported.  Please make sure to pass something like 'docs/'`,
       );
     }
@@ -275,7 +273,7 @@ const readAndMergeConfig = async () => {
     if (layoutsDirectory && typeof layoutsDirectory === "string") {
       customConfig.layoutsDirectory = layoutsDirectory;
     } else if (layoutsDirectory) {
-      return Promise.reject(
+      throw new Error(
         `Configuration error: provided layoutsDirectory "${layoutsDirectory}" is not supported.  Please make sure to pass something like 'layouts/'`,
       );
     }
@@ -284,7 +282,7 @@ const readAndMergeConfig = async () => {
       if (typeof prerender === "boolean") {
         customConfig.prerender = prerender;
       } else {
-        return Promise.reject(
+        throw new Error(
           `Configuration error: prerender must be a boolean; true or false.  Passed value was typeof: ${typeof prerender}`,
         );
       }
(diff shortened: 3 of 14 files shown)


packages/cli/src/commands/build.js
KaiPrince, line 1: I recommend hiding whitespace changes for easier reviewing

packages/cli/src/lifecycles/bundle.js
```diff
@@ -202,84 +206,58 @@ async function bundleStyleResources(compilation, resourcePlugins) {
...
-            intermediateResponse.clone(),
-            currentResponse.clone(),
-          );
+          const mergedResponse = mergeResponse(response.clone(), currentResponse.clone());
 
           if (mergedResponse.headers.get("Content-Type").indexOf(contentType) >= 0) {
-            return Promise.resolve(mergedResponse.clone());
+            response = mergedResponse.clone();
           }
         }
```
KaiPrince, line 221: Each stage is pretty much the same, so a future refactoring might be to extract each of the stages and let them share some code through a common function. It would also reduce the size of this massive function.

packages/cli/src/lifecycles/serve.js
```diff
@@ -69,9 +69,8 @@ async function getDevServer(compilation) {
       for (const plugin of resourcePlugins) {
         if (plugin.shouldServe && (await plugin.shouldServe(url, request))) {
           const current = await plugin.serve(url, request);
-          const merged = mergeResponse(response.clone(), current.clone());
 
-          response = merged.clone();
+          response = mergeResponse(response.clone(), current.clone());
```
KaiPrince, line 73: I notice that some of these use `current.clone()` and some don't. Is that intentional, or should they all be the same way? I've left it as is for now, but I can make them all the same while I'm here.

packages/plugin-renderer-puppeteer/src/puppeteer-handler.js
```diff
@@ -45,8 +47,7 @@ export default async function (compilation, callback) {
     console.info(
       "For more information please see this guide - https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md",
     );
-
-    return Promise.reject();
+    throw e;
```
KaiPrince, line 50: This one is technically a change, because you can't throw nothing syntactically in JS, but I think rejecting with the actual error instead of nothing is better here anyways.

test/smoke-test.js
```diff
@@ -269,7 +269,7 @@ function serve(label) {
 }
 
 async function runSmokeTest(testCases, label) {
-  testCases.forEach(async (testCase) => {
+  testCases.forEach((testCase) => {
```
KaiPrince, line 272: There are no `await` statements in the callback so I just removed the `async`.

packages/cli/src/lifecycles/bundle.js
```diff
@@ -123,36 +123,40 @@ async function optimizeStaticPages(compilation, plugins) {
...
-        const headers = new Headers({ "Content-Type": "text/html" });
-        let response = new Response(contents, { headers });
-
-        if (!(await checkResourceExists(outputDirUrl))) {
-          await fs.mkdir(outputDirUrl, {
-            recursive: true,
-          });
-        }
+      .map((page) =>
+        (async () => {
```
thescientist13, line 127: Just at a surface glance anyway, (no benchmarking per se) I think this approach may actually get us further away from the idiomatic usage I had in mind, since this seems like it would go from one (`async) function in the map 
```js
.map(async (page) => {
```

to two functions on each loop, all closures
```
.map((page) =>
        (async () => {
```

While performance wasn't necessarily the main goal here, I am a little sensitive to adding more overhead.  Plus, and what is a reoccuring theme in this codebase that I wanted to fix as part of this issue, is that I think `map` might have just been a bad choice in the first place, especially with the `async` added in there.  (FWIW I've been trying to get better at avoiding this pattern in newer code 😇 )

Instead here, I think it's fine to break this down into two operations, first filter, then run a `for...of`
```js
const pagesToRender = compilation.graph.filter(/* ... */)

for(const page of pagesToRender) {
  /* everything that was in the map goes here */
}
```

what do you think?

packages/cli/src/lifecycles/config.js
```diff
@@ -125,13 +129,13 @@ const readAndMergeConfig = async () => {
...
       if (!(workspace instanceof URL)) {
-        return Promise.reject("Configuration error: workspace must be an instance of URL");
+        throw new Error("Configuration error: workspace must be an instance of URL");
       }
 
       if (await checkResourceExists(workspace)) {
         customConfig.workspace = workspace;
       } else {
-        return Promise.reject(
+        throw new Error(
```
thescientist13, line 138: same as my comment above, let's avoid any additional refactorings in this PR not directly related to fixing all the loops and things just to try and keep the diff as small and focused as possible .  (we can always just open a separate PR for this if we feel strongly about it)

packages/cli/src/lifecycles/prerender.js
```diff
@@ -187,23 +187,23 @@ async function staticRenderCompilation(compilation) {
...
-
-      await trackResourcesForRoute(body, compilation, route);
-      await createOutputDirectory(route, new URL(scratchUrl.href.replace("index.html", "")));
-      await fs.writeFile(scratchUrl, body);
-
-      console.info("generated page...", route);
-
-      return Promise.resolve();
-    }),
+    pages.map((page) =>
```
thescientist13, line 190: This seems like it could easily just be a `for ... of`

packages/cli/src/lifecycles/serve.js
thescientist13: I _think_ they should all be using `clone`.

packages/plugin-renderer-puppeteer/src/plugins/server.js
```diff
@@ -15,8 +15,6 @@ class PuppeteerServer {
       (await getDevServer(this.compilation)).listen(offsetPort, async () => {
         console.info(`Started puppeteer prerender server at http://localhost:${offsetPort}`);
       });
-    } else {
```
thescientist13, line 18: are we sure this should / needs to be deleted?  I would test with `yarn develop` but I think it's OK to leave this one out of this PR since it's a little out of scope, IMO.

packages/plugin-graphql/src/schema/graph.js
```diff
@@ -66,7 +66,7 @@ const getParsedHeadingsFromPage = (tableOfContents = [], headingLevel) => {
 };
 
 const getPagesFromGraph = async (root, query, context) => {
-  return Promise.resolve(context.graph);
+  return context.graph;
```
thescientist13, line 69: this is probably fine, but as mentioned, maybe better to take all these `Promise` clean up tasks into their own PR?

thescientist13 (commented):
This is a great start, thank you so much!

My only re-occuring feedback as you'll see is that my hope was to _avoid_ `.map(async)` usage, and so "doubling" down in some cases here feels like more work / overhead / code then something more straightforward like a `for ... of` loop.  So I think in those cases if we just refactor that out like in one of my snippets, we should be good there.

Also, I can understand some of the Promise vs throw changes being `async` adjacent but I think it would probably best to isolate those to their own PR, just to help keep this one focused on just the loops and stuff.  (the nature of this refactoring has a lot of diffs, so a lot to read through)

Lastly, I think I found one last case of an async map, which should be easy enough to tweak
Screenshot 2025-08-05 at 10 22 53

----

Also, if you don't mind commenting on #823 I would like to assign it to you so it's easier to see on our board that someone is working on.

Thanks again and let me know if you have any questions / clarifications.

packages/cli/src/lifecycles/prerender.js
KaiPrince: likewise this would no longer run concurrently if we removed the promise.all.

packages/cli/src/lifecycles/bundle.js
KaiPrince: Is https://github.com/ProjectEvergreen/greenwood/pull/1546/commits/12332146b293bade5ac683ff5df2355ee9d47c4f closer to what you were thinking?

thescientist13: As I was posting this, I realized I had created such utils for `map` and `filter` already!  So maybe we can just refactor to use these?
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lib/async-utils.js
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lib/layout-utils.js#L184

thescientist13: thanks for checking!

So looking at those cases you shared, e.g.

```js
// plugin-renderer-lit
for (const script of scripts) {
  await import(script);
}
```

In the case of `await`s in a `for...of` loop, that would be OK, as even the ESLint docs acknowledge this would make sense in certain conditions
https://eslint.org/docs/latest/rules/no-await-in-loop#when-not-to-use-it
> In many cases the iterations of a loop are not actually independent of each other, and awaiting in the loop is correct

----

So basically, my main goal here was to refactor the usage of await in `map`, `forEach`, etc where `await` doesn't make sense and was more of convenience, and so those should definitely go away like you've done so far.  But for any `for...of` usages, I am fine to add an ESLint ignore line for those, which is likely what all those cases probably are.  I just want to catch and make sure we don't do this going forward for the cases we do care.

What do you think?

packages/cli/src/lifecycles/bundle.js
```diff
@@ -115,45 +116,41 @@ async function cleanUpResources(compilation) {
...
-        }
+  const pages = compilation.graph.filter(
+    (page) =>
+      !page.isSSR || (page.isSSR && page.prerender) || (page.isSSR && compilation.config.prerender),
+  );
 
-        for (const plugin of plugins) {
-          if (plugin.shouldOptimize && (await plugin.shouldOptimize(url, response.clone()))) {
-            const currentResponse = await plugin.optimize(url, response.clone());
+  await asyncMap(pages, async (page) => {
```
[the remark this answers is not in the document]
thescientist13: that makes sense, I would be open to an `asyncForEach`

packages/cli/src/lifecycles/bundle.js
```diff
@@ -202,84 +199,58 @@ async function bundleStyleResources(compilation, resourcePlugins) {
       const request = new Request(url, { headers });
       const initResponse = new Response(contents, { headers });
 
-      let response = await resourcePlugins.reduce(async (responsePromise, plugin) => {
-        const intermediateResponse = await responsePromise;
+      let response = initResponse;
+
+      for (const plugin of resourcePlugins) {
```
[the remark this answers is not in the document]
thescientist13: yeah, I would like to avoid using `reduce` unless it really makes sense, as they are harder to read / reason about as opposed to a `for...of` or similar.

thescientist13 (commented):
This coming along great!

I would be open to adding the linting rule as I think most of the violations flagged are probably just going to be in `for..of` loops and so I think those are fine to [add an ESLint ignore line for](https://github.com/ProjectEvergreen/greenwood/pull/1546#discussion_r2280084072)

In regards to your comment here
>  The downside is now lots of files depend on the custom `asyncMap` definition, but the upside is the rest of the code is slightly simpler.

I am OK with that tradeoff.  The functions are pretty simple and well named, and I think the value of having a consistent way to apply such logic is ideal for our use cases.

----

Also, there is one minor merge conflict but should be easy to resolve, just remove passing the `route` param here to `createOutputDirectory`

Screenshot 2025-08-15 at 9 21 08 PM

----

Thanks for doing the benchmarking!  💯 

Prior to the 1.0 release, I would like to revisit some of the work I've done as part of #970 in this benchmarking repo someone created for frameworks a while ago
https://github.com/thescientist13/bench-framework-markdown

packages/cli/src/lifecycles/bundle.js
thescientist13: thinking of this more, I think it would be fine if there the lint rule was somehow able to just flag the offenses we're refactoring here (e.g. `map`, `reduce`, `forEach`), otherwise I think let's use the rule to make sure we've found all usages and refactored those.  at least with all currant usages cleaned up, then I think this PR would be good to go.

packages/cli/src/lifecycles/bundle.js
```diff
@@ -300,11 +271,10 @@ async function bundleApiRoutes(compilation) {
 
   if (apiConfigs.length > 0 && apiConfigs[0].input.length !== 0) {
     console.info("bundling API routes...");
-    for (const configIndex in apiConfigs) {
-      const rollupConfig = apiConfigs[configIndex];
+    await asyncForEach(apiConfigs, async (rollupConfig) => {
```
KaiPrince, line 274: I double-checked, and `apiConfigs` is an array anyways, so converting from `for...in` to `for...of` works fine. No special handling was needed for accessing by index.

eslint.config.js
```diff
@@ -47,6 +47,7 @@ export default [
       "no-only-tests/no-only-tests": "error",
       "import/no-unresolved": "off",
       "import/enforce-node-protocol-usage": ["error", "always"],
+      "no-await-in-loop": "warn",
```
thescientist13, line 50: my concern with adding this rule with no "constraint" on this is that now we get a _ton_ of warnings
<details>
  <pre>
    /Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/commands/eject.js
  11:5  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/config/rollup.config.js
   78:42  warning  Unexpected `await` inside a loop  no-await-in-loop
   79:32  warning  Unexpected `await` inside a loop  no-await-in-loop
   90:38  warning  Unexpected `await` inside a loop  no-await-in-loop
   91:24  warning  Unexpected `await` inside a loop  no-await-in-loop
   98:14  warning  Unexpected `await` inside a loop  no-await-in-loop
  100:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  107:14  warning  Unexpected `await` inside a loop  no-await-in-loop
  109:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  114:41  warning  Unexpected `await` inside a loop  no-await-in-loop
  115:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  145:38  warning  Unexpected `await` inside a loop  no-await-in-loop
  287:40  warning  Unexpected `await` inside a loop  no-await-in-loop
  288:30  warning  Unexpected `await` inside a loop  no-await-in-loop
  293:38  warning  Unexpected `await` inside a loop  no-await-in-loop
  294:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  302:14  warning  Unexpected `await` inside a loop  no-await-in-loop
  304:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  312:14  warning  Unexpected `await` inside a loop  no-await-in-loop
  314:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  357:31  warning  Unexpected `await` inside a loop  no-await-in-loop
  363:38  warning  Unexpected `await` inside a loop  no-await-in-loop
  364:30  warning  Unexpected `await` inside a loop  no-await-in-loop
  637:22  warning  Unexpected `await` inside a loop  no-await-in-loop
  639:32  warning  Unexpected `await` inside a loop  no-await-in-loop
  646:22  warning  Unexpected `await` inside a loop  no-await-in-loop
  648:32  warning  Unexpected `await` inside a loop  no-await-in-loop
  655:22  warning  Unexpected `await` inside a loop  no-await-in-loop
  657:32  warning  Unexpected `await` inside a loop  no-await-in-loop
  663:19  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lib/async-utils.js
   6:21  warning  Unexpected `await` inside a loop  no-await-in-loop
  21:22  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lib/layout-utils.js
   23:11  warning  Unexpected `await` inside a loop  no-await-in-loop
  165:40  warning  Unexpected `await` inside a loop  no-await-in-loop
  198:40  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lib/resource-utils.js
  114:9  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lib/walker-package-ranger.js
  235:15  warning  Unexpected `await` inside a loop  no-await-in-loop
  255:11  warning  Unexpected `await` inside a loop  no-await-in-loop
  300:14  warning  Unexpected `await` inside a loop  no-await-in-loop
  304:11  warning  Unexpected `await` inside a loop  no-await-in-loop
  309:13  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lifecycles/bundle.js
   26:39  warning  Unexpected `await` inside a loop  no-await-in-loop
   27:18  warning  Unexpected `await` inside a loop  no-await-in-loop
   30:36  warning  Unexpected `await` inside a loop  no-await-in-loop
   31:18  warning  Unexpected `await` inside a loop  no-await-in-loop
   44:35  warning  Unexpected `await` inside a loop  no-await-in-loop
   45:50  warning  Unexpected `await` inside a loop  no-await-in-loop
  110:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  144:41  warning  Unexpected `await` inside a loop  no-await-in-loop
  145:37  warning  Unexpected `await` inside a loop  no-await-in-loop
  193:13  warning  Unexpected `await` inside a loop  no-await-in-loop
  194:9   warning  Unexpected `await` inside a loop  no-await-in-loop
  205:22  warning  Unexpected `await` inside a loop  no-await-in-loop
  224:18  warning  Unexpected `await` inside a loop  no-await-in-loop
  249:18  warning  Unexpected `await` inside a loop  no-await-in-loop
  274:18  warning  Unexpected `await` inside a loop  no-await-in-loop
  284:31  warning  Unexpected `await` inside a loop  no-await-in-loop
  292:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  305:22  warning  Unexpected `await` inside a loop  no-await-in-loop
  306:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  333:20  warning  Unexpected `await` inside a loop  no-await-in-loop
  334:20  warning  Unexpected `await` inside a loop  no-await-in-loop
  335:20  warning  Unexpected `await` inside a loop  no-await-in-loop
  337:20  warning  Unexpected `await` inside a loop  no-await-in-loop
  338:9   warning  Unexpected `await` inside a loop  no-await-in-loop
  346:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  372:20  warning  Unexpected `await` inside a loop  no-await-in-loop
  373:9   warning  Unexpected `await` inside a loop  no-await-in-loop
  377:13  warning  Unexpected `await` inside a loop  no-await-in-loop
  378:9   warning  Unexpected `await` inside a loop  no-await-in-loop
  387:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  433:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  434:9   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lifecycles/copy.js
  46:30  warning  Unexpected `await` inside a loop  no-await-in-loop
  48:30  warning  Unexpected `await` inside a loop  no-await-in-loop
  49:11  warning  Unexpected `await` inside a loop  no-await-in-loop
  53:11  warning  Unexpected `await` inside a loop  no-await-in-loop
  67:23  warning  Unexpected `await` inside a loop  no-await-in-loop
  73:9   warning  Unexpected `await` inside a loop  no-await-in-loop
  75:9   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lifecycles/graph.js
   73:10  warning  Unexpected `await` inside a loop  no-await-in-loop
   74:10  warning  Unexpected `await` inside a loop  no-await-in-loop
   77:27  warning  Unexpected `await` inside a loop  no-await-in-loop
   91:38  warning  Unexpected `await` inside a loop  no-await-in-loop
  110:33  warning  Unexpected `await` inside a loop  no-await-in-loop
  162:28  warning  Unexpected `await` inside a loop  no-await-in-loop
  175:29  warning  Unexpected `await` inside a loop  no-await-in-loop
  178:13  warning  Unexpected `await` inside a loop  no-await-in-loop
  347:20  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lifecycles/prerender.js
   23:32  warning  Unexpected `await` inside a loop  no-await-in-loop
   24:18  warning  Unexpected `await` inside a loop  no-await-in-loop
   40:8   warning  Unexpected `await` inside a loop  no-await-in-loop
   42:42  warning  Unexpected `await` inside a loop  no-await-in-loop
   45:36  warning  Unexpected `await` inside a loop  no-await-in-loop
   46:42  warning  Unexpected `await` inside a loop  no-await-in-loop
   92:16  warning  Unexpected `await` inside a loop  no-await-in-loop
   92:23  warning  Unexpected `await` inside a loop  no-await-in-loop
   93:12  warning  Unexpected `await` inside a loop  no-await-in-loop
   93:19  warning  Unexpected `await` inside a loop  no-await-in-loop
  112:23  warning  Unexpected `await` inside a loop  no-await-in-loop
  117:12  warning  Unexpected `await` inside a loop  no-await-in-loop
  142:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  143:5   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/lifecycles/serve.js
  70:36  warning  Unexpected `await` inside a loop  no-await-in-loop
  71:27  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/loader.js
  36:8   warning  Unexpected `await` inside a loop  no-await-in-loop
  41:24  warning  Unexpected `await` inside a loop  no-await-in-loop
  47:32  warning  Unexpected `await` inside a loop  no-await-in-loop
  51:44  warning  Unexpected `await` inside a loop  no-await-in-loop
  59:8   warning  Unexpected `await` inside a loop  no-await-in-loop
  66:11  warning  Unexpected `await` inside a loop  no-await-in-loop
  71:36  warning  Unexpected `await` inside a loop  no-await-in-loop
  75:44  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/src/plugins/resource/plugin-standard-html.js
  74:31  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/cli/test/cases/build.plugins.adapter/generic-adapter.js
  36:21  warning  Unexpected `await` inside a loop  no-await-in-loop
  40:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  43:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  53:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  58:7   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-adapter-aws/src/index.js
   84:21  warning  Unexpected `await` inside a loop  no-await-in-loop
   88:5   warning  Unexpected `await` inside a loop  no-await-in-loop
   91:5   warning  Unexpected `await` inside a loop  no-await-in-loop
   99:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  111:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  113:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  118:7   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-adapter-netlify/src/index.js
  107:21  warning  Unexpected `await` inside a loop  no-await-in-loop
  112:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  115:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  123:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  128:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  149:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  151:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  156:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  161:5   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-adapter-vercel/src/index.js
  100:21  warning  Unexpected `await` inside a loop  no-await-in-loop
  104:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  107:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  115:7   warning  Unexpected `await` inside a loop  no-await-in-loop
  127:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  129:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  134:7   warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-adapter-vercel/test/cases/build.config.base-path/build.config.base-path.spec.js
  119:28  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-adapter-vercel/test/cases/build.default.options-runtime/build.default.options-runtime.spec.js
  119:28  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-adapter-vercel/test/cases/build.default/build.default.spec.js
  135:28  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-css-modules/src/index.js
   51:32  warning  Unexpected `await` inside a loop  no-await-in-loop
   52:18  warning  Unexpected `await` inside a loop  no-await-in-loop
   59:8   warning  Unexpected `await` inside a loop  no-await-in-loop
   61:18  warning  Unexpected `await` inside a loop  no-await-in-loop
   68:8   warning  Unexpected `await` inside a loop  no-await-in-loop
   70:18  warning  Unexpected `await` inside a loop  no-await-in-loop
  168:5   warning  Unexpected `await` inside a loop  no-await-in-loop
  222:11  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-graphql/src/schema/schema.js
  43:53  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-include-html/src/index.js
  29:33  warning  Unexpected `await` inside a loop  no-await-in-loop
  50:42  warning  Unexpected `await` inside a loop  no-await-in-loop
  51:33  warning  Unexpected `await` inside a loop  no-await-in-loop
  51:51  warning  Unexpected `await` inside a loop  no-await-in-loop

/Users/owenbuckley/Workspace/project-evergreen/greenwood/packages/plugin-renderer-lit/src/execute-route-module.js
  26:7  warning  Unexpected `await` inside a loop  no-await-in-loop
  </pre>
</details>

What I was trying to callout in [my comment](https://github.com/ProjectEvergreen/greenwood/pull/1546#pullrequestreview-3125484365) was that using `await` in a `for ... of` is fine.  So unless we can limit the linting to _just_ `map`, `filter`, etc then I worry about all the extra noise this adds.  The only other option is to add an eslint ignore to all those locations, which seemed pretty tedious.

thescientist13: hmm, I realize i kind of left an unfinished sentence here, my apologies 😞 
> I would be open to adding the linting rule as I think most of the violations flagged are probably just going to be in `for..of` loops and so I think those are fine to [add an ESLint ignore line for](https://github.com/ProjectEvergreen/greenwood/pull/1546#discussion_r2280084072)

So yeah, my above comment was what I trying to say, so sorry about that mixup

KaiPrince: Yeah it's a lot. That's what I was trying to say in [my other comment](https://github.com/ProjectEvergreen/greenwood/pull/1546#discussion_r2268049875). We'd have to add `eslint-disable` comments to a lot of places, or just leave it as a warn (which as you say can get pretty noisy). 

I thought you confirmed you wanted it for [keeping us "honest"](https://github.com/ProjectEvergreen/greenwood/pull/1546#discussion_r2264168381), but no problem I'll just revert this config change for now. If we decide later we want to add all those eslint-ignore comments, I can open a separate PR.

thescientist13: Yes, apologies I didn't finish my full thought in that most recent comment, but I was trying to say if added, it should be done with the ability to carve out an exception for the use cases we do want it for (e.g. `for...of`, `for..in`)

So the options would be
1. Enable with `error` and add `// eslint-disable-line no-await-in-loop` everywhere its allowed (or see if we can just have it error on `map`, `filter`, etc
1. Just leave things as they are

So I was leaning towards option #2 depending if the carve out was feasible, or if we wanted to sprinkle in all those ignore lines.

Thoughts?

Like I said I am good with things right now, I've personally broken the habit of the async maps, filters etc so I am at least trying to keep myself honest now 😇

KaiPrince: I started adding eslint-ignore comments initially, but I think I hit a snag in the rollup config file. Let's do option 2 for now, and I'll start a follow-up PR to see where option 1 takes us.

PR has been updated from master 👍

packages/cli/src/lifecycles/bundle.js
```diff
@@ -101,65 +102,61 @@ async function emitResources(compilation) {
 async function cleanUpResources(compilation) {
   const { outputDir } = compilation.context;
 
-  for (const resource of compilation.resources.values()) {
+  await asyncForEach(compilation.resources.values(), async (resource) => {
```
thescientist13, line 105: yeah, for example, why create our custom utility here when we already have the language construct built-in?  I think anywhere where we were actually using `for...of` / `for...in` was fine as is, that's the end state we wanted the whole time, but obviously map and filter have special behaviors, hence the custom utilities.

KaiPrince: Building on my explanation in the other comment, this is a case where we want to take advantage of parallelism. As an example, assume we need to delete ("unlink") 100 files.

In the original implementation, each `fs.unlink` call doesn't start until the previous one finishes. i.e. the first runs, then the second, and so on. The call to delete the 100th file is only started after the other 99 files are deleted.

In the new implementation, each iteration is wrapped in a `Promise.all`, which means they run concurrently. All 100 `fs.unlink` calls are started, _and then_ we wait for them all to finish before returning from this function. Since the filesystem calls are performed by the OS, they run outside of the main JS thread and are actually run in parallel. This results in performance improvements since we get to take advantage of the OS using multiple threads.

packages/cli/src/lib/async-utils.js
```diff
@@ -1,27 +1,33 @@
...
   for (const item of items) {
-    mappedItems.push(await mapper(item));
+    promises.push(mapper(item));
   }
 
-  return mappedItems;
+  return Promise.all(promises);
+}
+
+async function asyncForEach(items, callback) {
```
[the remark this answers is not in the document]
thescientist13: That video link was great btw, Fireship is the best!  🔥

PR #1457, opened 2025-03-19
feature/issue 1247 standalone markdown plugin

thescientist13:
## Related Issue

resolves #1247 

## Documentation 

1. [ ] Needs documentation for the following
    - move to plugins page (w/ netlify redirect)
    - `preServe` lifecycle
    - tocHeadings - https://greenwoodjs.dev/docs/content-as-data/pages-data/#table-of-contents

## Summary of Changes

1. Created standalone **unified** based markdown plugin
1. Moved over all deps from the CLI

## TODO
1. [ ] how to support GraphQL and ToC data?
    - can it just resolve to the content server instead?
1. [ ] Confirm live reload 
1. [ ] Port Greenwood test cases
1. [ ] custom elements wrapper logic
1. [ ] integrate `shouldPreServe` / `preServe` in all places (e.g. _loader.js_)
1. [ ] Custom plugin options - https://github.com/ProjectEvergreen/greenwood/issues/1247#issuecomment-2721881358
    - types (plugins config, exports, etc)
    - tests
1. [ ] with pre lifecycles, do we even need the distinction of greenwood vs non-greenwood plugins at the top of _serve.js_?
1. [ ] Documentation
    - ~~README~~
    - options (plugins, table of content headings)
1. [ ] (nice to have) - standalone usage - https://github.com/ProjectEvergreen/greenwood/issues/1247

## Thoughts / Questions
1. [ ] does this help with some hot reloading of content as data?  https://github.com/ProjectEvergreen/greenwood/issues/1278
1. [ ] Refine plugins API discussion
    - pre option like in Vite
    - early return instead of `shouldServe` _and_ `serve`

PR #1565, opened 2025-09-01
fix response clone bug

KaiPrince:
Fix the .mp3 tests.

Failing test runs:
https://github.com/KaiPrince/greenwood/actions/runs/17111099878/job/48532445170

https://github.com/ProjectEvergreen/greenwood/pull/1457#issuecomment-3194741722

Fix: https://github.com/KaiPrince/greenwood/pull/2/commits/b133a73abee080cb62426e403624d13223d03cec

### DIFF
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -59,8 +59,13 @@ async function getDevServer(compilation) {
       let response = new Response(null, { status });
 
       for (const plugin of resourcePlugins) {
-        if (!plugin.servePage && plugin.shouldServe && (await plugin.shouldServe(url, request))) {
-          const current = await plugin.serve(url, request);
+        // ignore plugins that serve pages, as those will be handled by Greenwood's standard HTML plugin
+        if (
+          !plugin.servePage &&
+          plugin.shouldServe &&
+          (await plugin.shouldServe(url, request, response.clone()))
+        ) {
+          const current = await plugin.serve(url, request, response.clone());
           const merged = mergeResponse(response.clone(), current.clone());
 
           response = merged.clone();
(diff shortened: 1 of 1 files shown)


thescientist13:
Took a little while with all the testing, but with this change in my branch
https://github.com/ProjectEvergreen/greenwood/pull/1457/commits/23fefcab2d85837fbeaf841061aeeb912ff47cec

it has now been pretty stable again across Linux, Mac, and Windows.

KaiPrince:
The hardest part about debugging this is it's so flaky. I thought I had fixed it previously, but turns out it can just pass for no reason. I'm hoping by the time I've figured this out, I've also managed to discover the cause of the flakiness and make it more deterministic.

image

thescientist13:
Just a heads up I got my last little stretch item into #1457 so will probably merge by end of week, and builds still seem stable.  (also I rebased against mainline so your history got a little messed up since it looks like you re-pointed to that branch)

That said, I wouldn't be opposed to continuing this effort if there are some race conditions / inconsistencies happening that we want to keep stabalizing.  I know there could be some more consistency applied all around, especially with `mergeResponse`
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/bundle.js#L21
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/loader.js#L22
- https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/lifecycles/prerender.js#L33

Appreciate you looking into this!

KaiPrince:
Oh man, it's driving me crazy but I'm stumped. I have no explanation for why extracting a local const one line above fixes it in certain cases. I'm going to drop this and move on for now. If I'm hit with inspiration I'll revisit this.

PR #1612, opened 2025-11-05
bug(cli): #1611 handle view transition psuedo selectors

### LINKED ISSUE #1611: View transition pseudo-selectors compile incorrectly
### What happened?

The following CSS:
```css
::view-transition-old(content) {
  animation: none;
}
::view-transition-new(content) {
  animation: slide-from-right 0.2s ease-out;
}
```
gets compiled to the invalid CSS:
```css
::view-transition-oldcontent {
  animation: none;
}
::view-transition-newcontent {
  animation: slide-from-right 0.2s ease-out;
}
```

### Steps to reproduce

1. Copy the above into a CSS file used in a Greenwood app
2. Compile the app
3. Observe the compiled version of the file

### Environment

Greenwood 0.33.3
Tested on Chrome

### Additional Context

My initial thought was that it could be a PostCSS issue, but it still occurs with the plugin disabled.

This does not occur with `optimization: "none"`.

EskiMojo14:
## Related Issue

Resolves #1611

## Documentation 

n/a

## Summary of Changes

Added view transition pseudo selectors to CSS bundler to fix compilation

### DIFF
--- a/packages/cli/test/cases/build.config.optimization-default/fixtures/expected.css
+++ b/packages/cli/test/cases/build.config.optimization-default/fixtures/expected.css
@@ -83,3 +83,15 @@ tabbed-custom-element::part(tab){color:#0c0dcc;border-bottom:transparent solid 2
 h2{& span{color:red}}
 
 @media  (max-width:768px){span{flex-direction:column;text-align:center;}}
+
+@view-transition {navigation:auto}
+
+::view-transition-group(root){position:absolute}
+
+::view-transition-image-pair(root){position:absolute}
+
+@keyframes slide-from-right{from{transform:translateX(100%)}to{transform:translateX(0%)}}
+
+::view-transition-old(root){animation:none}
+
+::view-transition-new(root){animation:slide-from-right 0.3s ease-out}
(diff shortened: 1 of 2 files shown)


thescientist13 (commented):
This is great, thank you!

One thought would be if we could get a quick test for this added too?  There is a test suite for testing CSS bundling, and would basically just be a two step process:
1. Add an example of the CSS here to the end of this file - https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.optimization-default/src/styles/main.css
1. Add the expected output CSS here (basically removing whitespace and last `;`s - https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.optimization-default/fixtures/expected.css

Then just run the test suite (`yarn test`).  To save a little time, just change the top level `describe` to `describe.only`
https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/test/cases/build.config.optimization-default/build.config-optimization-default.spec.js

----

Will get this out in the next alpha release (happening this weekend, or next at the latest).

thescientist13 (approved):
Awesome, thank you so much!

PR #1699, opened 2026-06-28
fix(cli): #1622 fix resource linking for pages when using get static paths

### LINKED ISSUE #1622: Support pre-rendering (SSG) for dynamic routes
### Motivation

Now that Greenwood has support for [dynamic routing](https://github.com/ProjectEvergreen/greenwood/issues/882), it should be possible to build out the capability to generating static pages at build time using this convention

```js
// src/pages/blog/[slug].js

  return [{
    params: { slug: 'my-first-post' },
  }, {
    params: { slug: 'my-second-post' },
  }]
}

export default class BlogPostPage extends HTMLElement {
  #slug: string;

  constructor({ params }) {
    super();
    this.#slug = params?.slug;
  }

  connectedCallback() {
    this.innerHTML = `
      <body>
        <h1>${this.#slug}</h1>
      </body>
    `;
  }
}
```

Would generate two pages
```
public/
  blog/
    my-first-post.html
    my-second-post.html
```

### Technical Design

We would also want to support something like `getStaticProps` too, to support the ability to configure props and paths in one-shot, e.g.

```ts
// src/pages/artists/[name].ts
import { getArtists, getArtistById } from '#services/artists.ts';
import type { Artist } from '#services/artists.ts';

interface StaticPaths {
  params: {
    name: string;
    id: number;
  }
}

interface PageProps {
  params: {
    artist: Artist;
  }
}

export async function getStaticPaths(): Promise<StaticPaths[]> {
  const artists = await getArtists();

  return artists.map(artist => {
    return {
      params: {
        name: artist.name,
        id: artist.id,
      }
    }
  });
}

export async function getStaticProps({ params }: StaticPaths): Promise<Artist> {
  const artist = await getArtistById(params.id);

  return artist;
}


export default class ArtistDetailsPage extends HTMLElement {
  #artist: Artist;

  constructor({ params }: PageProps) {
    super();
    this.#artist = params?.artist;
  }

  async connectedCallback() {
    this.innerHTML = `
      <body>
        <h1>${this.#artist.name}</h1>
      </body>
    `;
  }
}
```
----

Very much inspired by these implementations
- https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-props
- https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-paths
- https://docs.astro.build/en/guides/routing/#static-ssg-mode

Would also be good to include some types for this

### Additional Context

It would be interesting to see if we still want to support the source plugin after this, or if we should deprecate it?
https://greenwoodjs.dev/docs/reference/plugins-api/#source

thescientist13:
## Related Issue

related to #1622 

Noticed in [this project](https://github.com/AnalogStudiosRI/www.analogstudios.net/pull/112#pullrequestreview-4585963456) that page specific resources were missing when building
image
image

## Documentation 

N / A

## Summary of Changes

1. Refactor page matching in `getMatchingPageByRoute` to have more explicit matching by "specificity"
1. Update `trackResourceForRoute` to leverage `getMatchingPageByRoute`
1. Misc refactoring
1. Update test case to reproduce / validate

### DIFF
--- a/packages/cli/src/lib/graph-utils.js
+++ b/packages/cli/src/lib/graph-utils.js
@@ -28,28 +28,42 @@ function getStaticPages(compilation) {
 }
 
 // get a page by route; including getStaticPaths or dynamic SSR pages
+// not sure if there's a better way to filter through all the possible matches in one-shot?
 function getMatchingPageByRoute(compilation, route) {
   const { graph, config } = compilation;
 
-  return graph.find((page) => {
-    return (
-      // exact match
-      page.route === route ||
-      // dynamic route
-      (page.segment &&
-        new URLPattern({ pathname: `${config.basePath}${page.segment.pathname}` }).test(
-          `https://example.com${route}`,
-        )) ||
-      // getStaticPaths
-      (page.hasStaticParams &&
-        page.staticPaths.find((path) => {
-          const { segment } = page;
-          const staticRoute = route.replace(`[${segment.key}]`, path.params[segment.key]);
+  const exactMatch = graph.find((page) => page.route === route);
 
-          return `${config.basePath}${staticRoute}` === route;
-        }))
-    );
-  });
+  if (exactMatch) {
+    return exactMatch;
+  }
+
+  const dynamicMatch = graph.find(
+    (page) =>
+      page.segment &&
+      new URLPattern({ pathname: `${config.basePath}${page.segment.pathname}` }).test(
+        `https://example.com${route}`,
+      ),
+  );
+
+  if (dynamicMatch) {
+    return dynamicMatch;
+  }
+
+  const staticMatch = graph.find(
+    (page) =>
+      page.hasStaticParams &&
+      page.staticPaths.find((path) => {
+        const { segment } = page;
+        const staticRoute = route.replace(`[${segment.key}]`, path.params[segment.key]);
+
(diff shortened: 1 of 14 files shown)


packages/cli/src/lib/graph-utils.js
```diff
@@ -28,28 +28,42 @@ function getStaticPages(compilation) {
...
+      new URLPattern({ pathname: `${config.basePath}${page.segment.pathname}` }).test(
+        `https://example.com${route}`,
+      ),
+  );
+
+  if (dynamicMatch) {
+    return dynamicMatch;
+  }
+
+  const staticMatch = graph.find(
```
thescientist13: Should this come first since it's a more specific check?

packages/cli/src/lib/resource-utils.js
```diff
@@ -208,9 +209,9 @@ async function trackResourcesForRoute(html, compilation, route) {
     });
   });
 
-  compilation.graph.find((page) => page.route === route).resources = resources.map(
-    (resource) => resource.sourcePathURL.pathname,
-  );
+  const matchingRoute = getMatchingPageByRoute(compilation, route);
```
thescientist13, line 212: Are there other places this matching might be happening we should check for?

PR #1708, opened 2026-07-18
fix(cli): #1707 skip over unsupported api files instead of aborting the build

### LINKED ISSUE #1707: Unsupported (non-js/ts) file in src/pages/api/ aborts the whole build with a TypeError
### What happened?

A single unsupported (non-`.js`/`.ts`) file placed anywhere under `src/pages/api/` aborts the entire Greenwood build and produces zero output.

When `walkDirectoryForPages` (in `packages/cli/src/lifecycles/graph.js`) encounters an API-route file whose extension is not supported, it logs the expected warning:

```
<file> is not a supported API file extension, skipping...
```

but then executes a bare `return;` instead of `continue;`. Returning `undefined` out of the recursive walk causes the caller (`pages = nextPages.pages;`) to throw:

```
TypeError: Cannot read properties of undefined (reading 'pages')
    at walkDirectoryForPages (.../packages/cli/src/lifecycles/graph.js:84:27)
    at async generateGraph (.../packages/cli/src/lifecycles/graph.js:344:9)
    at async generateCompilation (.../packages/cli/src/lifecycles/compile.js:113:19)
    at async run (.../packages/cli/src/index.js:12:25)
```

The warning implies the file is merely skipped, but in practice the whole build crashes. A stray `.txt`, `.md`, or editor scratch file left in the api directory is enough to brick the build.

### Steps to reproduce

1. Create a minimal Greenwood project:
   ```
   src/
     pages/
       index.html
       api/
         greeting.js      # valid API route
         notes.txt        # any non-js/ts file
   ```
2. Run `greenwood build`.
3. Observe the build throws `TypeError: Cannot read properties of undefined (reading 'pages')` and exits with no output, rather than skipping `notes.txt` and building the rest of the site.

Expected: the unsupported file is skipped (with the warning) and the build completes, emitting `index.html` and the `api/greeting.js` route.

### Environment

- Greenwood version: v0.34.0
- OS: Linux
- Node version: 22.20
- Browser: N/A (build-time issue)

### Additional Context

Root cause is a single control-flow line in `packages/cli/src/lifecycles/graph.js`: the unsupported-extension branch inside the `for (const filename of files)` loop uses `return;` where `continue;` was intended.

jstockdi:
## Related Issue

Resolves #1707

## Documentation

N / A — no user-facing documentation changes; this is a bug fix to build-time behavior.

## Summary of Changes

1. [x] Fixed `walkDirectoryForPages` in `packages/cli/src/lifecycles/graph.js` to `continue;` (skip the current file) instead of `return;` when an API-route file has an unsupported extension. The bare `return;` returned `undefined` out of the recursive page walk, causing the caller to throw `TypeError: Cannot read properties of undefined (reading 'pages')` and aborting the whole build.
2. [x] Added a regression test case `packages/cli/test/cases/build.default.api-ignored-files/` with a fixture containing a stray `src/pages/api/notes.txt` alongside a valid `src/pages/api/greeting.js` and `src/pages/index.html`. The spec asserts the build succeeds and emits both the index page and the valid API route while skipping the unsupported file.
3. [x] Verified the new spec fails on unpatched source (the `TypeError`) and passes with the fix.

No breaking changes.

### DIFF
--- a/packages/cli/src/lifecycles/graph.js
+++ b/packages/cli/src/lifecycles/graph.js
@@ -105,7 +105,9 @@ const generateGraph = async (compilation) => {
         if (isApiRoute) {
           if (extension !== "js" && extension !== "ts" && !isCustom) {
             console.warn(`${filenameUrl} is not a supported API file extension, skipping...`);
-            return;
+            // skip this file but keep walking the directory instead of aborting the whole build
+            // https://github.com/ProjectEvergreen/greenwood/issues/1707
+            continue;
           }
 
           // is there a better way to detect for isolation export without having to actually import() the module?
--- a/packages/cli/test/cases/build.default.api-ignored-files/build.default.api-ignored-files.spec.js
+++ b/packages/cli/test/cases/build.default.api-ignored-files/build.default.api-ignored-files.spec.js
@@ -1,0 +1,77 @@
+/*
+ * Use Case
+ * Run Greenwood build command with no config and an unsupported (non-js/ts) file in the api directory.
+ *
+ * User Result
+ * Should generate a bare bones Greenwood build, skipping the unsupported api file while still
+ * emitting the valid API route and pages (instead of aborting the whole build).
+ *
+ * User Command
+ * greenwood build
+ *
+ * User Config
+ * None (Greenwood Default)
+ *
+ * User Workspace
+ * src/
+ *   pages/
+ *     api/
+ *       greeting.js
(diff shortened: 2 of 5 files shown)


thescientist13 (approved):
Nice, this looks good and thanks for adding a clear test case 🏆 

Will queue this up for the next minor release line (v0.35.0) 👍

PR #1712, opened 2026-07-18
fix(cli): #1711 strip only leading frontmatter block from html page bodies

### LINKED ISSUE #1711: Frontmatter strip on HTML pages greedily deletes body content up to the last dash fence
### What happened?

When an HTML page begins with YAML frontmatter, Greenwood strips the frontmatter block before rendering the page. The strip uses a greedy regular expression (`body.replace(/---(.*)---/s, "")`) that, with the `s` (dotAll) flag, matches from the frontmatter's opening `---` all the way to the **last** `---` anywhere in the document — not to the frontmatter's closing fence.

As a result, any body content that itself contains a `---` sequence (an ASCII divider, em-dash typography written as `---`, an inline code sample, etc.) causes everything between the frontmatter and that later `---` to be silently deleted from the output. There is no error; the build succeeds and produces a page that is missing content. Confusingly, `graph.json` still shows the correct frontmatter, so the loss is hard to attribute.

This runs on the standard HTML resource path, so both `greenwood build` and `greenwood develop` produce the same truncated output.

### Steps to reproduce

1. Create a Greenwood project with a single page `src/pages/index.html`:
   ```html
   ---
   title: "Dashes Page"
   ---

   <body>
     <h1>Before the dashes</h1>
     <p>some --- inline --- dashes</p>
     <h2>After the dashes</h2>
   </body>
   ```
2. Run `greenwood build`.
3. Open `public/index.html` and inspect the rendered `<body>`.

**Expected:** only the leading frontmatter block is removed; the `<h1>`, the full `<p>` text (`some --- inline --- dashes`), and the `<h2>` all survive.

**Actual:** everything from the frontmatter up to the last `---` in the `<p>` is deleted. The `<h1>` and the first part of the paragraph are gone; the rendered body begins with ` dashes` (the tail of the paragraph).

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/plugins/resource/plugin-standard-html.js:60` — `body = body.replace(/---(.*)---/s, "")`. The greedy `.*` under the `/s` flag extends the match to the final `---` in the document instead of the frontmatter's closing fence.

jstockdi:
## Related Issue

Resolves #1711

## Documentation

N/A — no user-facing documentation changes (bug fix only; the documented behavior, "the leading frontmatter block is removed", is unchanged and now correct).

## Summary of Changes

1. [x] Replaced the greedy frontmatter strip in `packages/cli/src/plugins/resource/plugin-standard-html.js` — `body.replace(/---(.*)---/s, "")` — which matched from the frontmatter's opening `---` to the **last** `---` anywhere in the document, silently deleting body content.
2. [x] The strip now reuses the `front-matter` (`fm`) parser's computed body (`fm(body).body`) — the same package/parser already used to read the frontmatter attributes in `lifecycles/graph.js`, so the removed block and the parsed attributes always agree and only the leading block is removed.
3. [x] Added regression test `packages/cli/test/cases/build.default.frontmatter-body-dashes/` — an HTML page whose body contains inline `---` sequences; asserts the build succeeds and the `<h1>`, full `<p>` text, and `<h2>` all survive while the frontmatter block is gone. (Fails on unpatched source; passes with the fix.)
4. [x] No breaking changes.

Note: the new spec uses only the `"public"` group of the shared smoke tests. The shared `"index"` group asserts the rendered body contains no `---`, which is deliberately false for this fixture (the whole point is that inline `---` in the body must be preserved).

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-standard-html.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-html.js
@@ -5,6 +5,7 @@
  *
  */
 import fs from "node:fs/promises";
+import fm from "front-matter";
 import { getPageLayout, getAppLayout, getGreenwoodScripts } from "../../lib/layout-utils.js";
 import { requestAsObject } from "../../lib/resource-utils.js";
 import { getMatchingDynamicSsrRoute, getParamsFromSegment } from "../../lib/url-utils.js";
@@ -57,7 +58,10 @@ class StandardHtmlResource {
       // purge frontmatter data from HTML files that only have a `<body>` tag
       // frontmatter outside of an <html> tag will get ignored by node-html-parser
       if (body.trim().startsWith("---")) {
-        body = body.replace(/---(.*)---/s, "");
+        // strip only the leading frontmatter block via the same parser that reads its
+        // attributes, so a later `---` in the body isn't greedily matched and deleted
+        // https://github.com/ProjectEvergreen/greenwood/issues/1711
+        body = fm(body).body;
       }
     } else if (matchingRoute.servePage === "static") {
       const customStaticPageFormatPlugins = config.plugins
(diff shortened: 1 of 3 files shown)


thescientist13 (approved):
Thank you!  😊

PR #1714, opened 2026-07-18
fix(cli): #1713 decode url-encoded dynamic route params for get static paths

### LINKED ISSUE #1713: Dynamic route params are delivered percent-encoded to getBody and getStaticParams
### What happened?

Dynamic route params from `getStaticPaths` reach user code (`getBody`, `getStaticParams`) percent-encoded instead of decoded. When a slug contains any non-ASCII character or a space, the value returned from `getStaticPaths` (e.g. `café`, `hello world`) is delivered to the page as `caf%C3%A9` / `hello%20world`.

`lib/url-utils.js` `getParamsFromSegment` returns the raw `URLPattern` `pathname.groups`, which URL parsing has already percent-encoded, and never calls `decodeURIComponent`. As a result:

- The rendered page shows the encoded value rather than the original slug — an i18n blog slug renders the wrong text.
- With a `getStaticParams` export the build crashes outright: `execute-route-module.js` matches the decoded `getStaticPaths` value (`café`) against the encoded `URLPattern` value (`caf%C3%A9`), finds no match, spreads `undefined`, and the user's `getStaticParams({ params })` destructure throws `TypeError: Cannot read properties of undefined`. The stack trace points into the user's own page code, not Greenwood.

ASCII-only slugs work, so the failure only appears once real-world (i18n / spaced) values are used.

### Steps to reproduce

1. Create `src/pages/[slug].js`:
   ```js
     return [{ params: { slug: "café" } }];
   }

     return `<h2>${params.slug}</h2>`;
   }
   ```
2. Run `greenwood build`.
3. Open the generated `public/café/index.html`.

Observed: the page contains `<h2>caf%C3%A9</h2>` (percent-encoded).
Expected: the page contains `<h2>café</h2>` (the original, decoded slug).

(If the page also exports `getStaticParams` that looks the slug up in a data list, the build fails instead with `TypeError: Cannot read properties of undefined` thrown from the user's page.)

### Environment

- Greenwood: v0.34.0
- Node: 22.20.0
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/lib/url-utils.js` `getParamsFromSegment` returns raw (percent-encoded) `URLPattern` groups without `decodeURIComponent`. The same decode is already applied to page ids/labels/titles in `packages/cli/src/lifecycles/graph.js`. The output directory name is derived separately from the already-decoded `getStaticPaths` params, so it is unaffected — only the params handed to `getBody` / `getStaticParams` are wrong.

jstockdi:
## Related Issue

Resolves #1713

## Documentation

N/A — no user-facing documentation change. This restores the already-documented behavior that dynamic route params round-trip losslessly from `getStaticPaths` to `getBody` / `getStaticParams`.

## Summary of Changes

1. [x] `getParamsFromSegment` (`packages/cli/src/lib/url-utils.js`) now `decodeURIComponent`s each `URLPattern` group value, mirroring the decode already applied to ids/labels/titles in `lifecycles/graph.js`. Decoding is guarded with try/catch so a malformed value falls back to its raw form, and `undefined` catch-all group values are passed through untouched.
2. [x] This makes non-ASCII / spaced slugs (e.g. `café`, `hello world`) reach `getBody` decoded, and fixes the `execute-route-module.js` param comparison (both sides are now decoded), so `getStaticParams` pages no longer crash the build.
3. [x] Added regression test `packages/cli/test/cases/build.default.dynamic-params-encoded/`: a `[slug].js` whose `getStaticPaths` returns `café` and whose `getBody` echoes `params.slug`; asserts the built page contains the decoded `café`, not `caf%C3%A9`. Verified failing on unpatched source and passing with the fix.
4. [x] No breaking changes. Output file/directory paths are unaffected (they derive from the already-decoded `getStaticPaths` params). Existing dynamic-routing / getStaticPaths / SSR / API-route specs continue to pass.

### DIFF
--- a/packages/cli/src/lifecycles/bundle.js
+++ b/packages/cli/src/lifecycles/bundle.js
@@ -18,7 +18,7 @@ import { rollup } from "rollup";
 import { pruneGraph } from "../lib/content-utils.js";
 import { asyncForEach } from "../lib/async-utils.js";
 import { getDynamicPages, getStaticPages } from "../lib/graph-utils.js";
-import { getStaticRouteFromDynamicRoute } from "../lib/url-utils.js";
+import { getStaticRouteFromDynamicRoute, getOutputHrefForStaticPath } from "../lib/url-utils.js";
 
 async function interceptPage(url, request, plugins, body) {
   let response = new Response(body, {
@@ -127,14 +127,12 @@ async function optimizeStaticPages(compilation, plugins) {
       for (const staticPath of staticPaths) {
         const staticRoute = getStaticRouteFromDynamicRoute(staticPath, segment, route);
         const outputDirUrl = new URL(
-          outputHref
-            .replace(`[${segment.key}]`, staticPath.params[segment.key])
-            .replace("index.html", ""),
+          getOutputHrefForStaticPath(staticPath, segment, outputHref).replace("index.html", ""),
         );
         const url = new URL(`http://localhost:${compilation.config.port}${staticRoute}`);
         const contents = await fs.readFile(
           new URL(
-            `./${outputHref.replace(outputDir.href, "").replace(`[${segment.key}]`, staticPath.params[segment.key])}`,
+            `./${getOutputHrefForStaticPath(staticPath, segment, outputHref).replace(outputDir.href, "")}`,
             scratchDir,
           ),
           "utf-8",
@@ -160,7 +158,7 @@ async function optimizeStaticPages(compilation, plugins) {
         const body = (await response.text()).replace(/data-gwd-opt=".*?[a-z]"/g, "");
 
         await fs.writeFile(
-          new URL(outputHref.replace(`[${segment.key}]`, staticPath.params[segment.key])),
+          new URL(getOutputHrefForStaticPath(staticPath, segment, outputHref)),
           body,
         );
       }
--- a/packages/cli/src/lib/url-utils.js
+++ b/packages/cli/src/lib/url-utils.js
@@ -39,9 +39,26 @@ function getMatchingDynamicSsrRoute(compilation, route) {
 
 // get params for dynamic routes from URLPattern based segment extraction
 function getParamsFromSegment(compilation, segment, route) {
-  return new URLPattern({ pathname: `${compilation.config.basePath}${segment.pathname}` }).exec(
-    `https://example.com${route}`,
-  )?.pathname?.groups;
+  const groups = new URLPattern({
+    pathname: `${compilation.config.basePath}${segment.pathname}`,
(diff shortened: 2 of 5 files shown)


thescientist13 (approved):
💯

jstockdi:
Heads up post-approval: pushed 87957b3e in response to [Owen's question on #1710](https://github.com/ProjectEvergreen/greenwood/pull/1710#issuecomment-5132284725) — a `getStaticPaths` param containing a literal `%` (e.g. `"100%"`) crashed the build with `URIError: URI malformed` when the raw value was spliced into the output `file://` URL (`fs.mkdir` in prerender.js, then `fs.readFile` in bundle.js). The commit adds a shared `getOutputHrefForStaticPath` helper that percent-encodes the param at those splice sites, plus a test case for a `100%` slug. The adapter, develop/serve dynamic-routing, and base-path suites all still pass. Happy to move it to a separate PR if you'd rather keep this one as approved.

packages/cli/src/lifecycles/bundle.js
```diff
@@ -127,14 +127,12 @@ async function optimizeStaticPages(compilation, plugins) {
...
-          outputHref
-            .replace(`[${segment.key}]`, staticPath.params[segment.key])
-            .replace("index.html", ""),
+          getOutputHrefForStaticPath(staticPath, segment, outputHref).replace("index.html", ""),
         );
         const url = new URL(`http://localhost:${compilation.config.port}${staticRoute}`);
         const contents = await fs.readFile(
           new URL(
-            `./${outputHref.replace(outputDir.href, "").replace(`[${segment.key}]`, staticPath.params[segment.key])}`,
+            `./${getOutputHrefForStaticPath(staticPath, segment, outputHref).replace(outputDir.href, "")}`,
```
thescientist13, line 135: nice, I think I had tried to do this the first time around but couldn't get it working, so this is definitely a nice refactor

thescientist13 (commented):
Also confirmed this with `greenwood develop` too 👍 

Image

thescientist13:
@jstockdi 
There's a merge conflict here now, fyi

jstockdi:
Rebased onto `37e68db3` and force-pushed as `e90bb723` — conflict resolved, PR is mergeable again.

The conflict was additive only: master gained `safeDecodeURIComponent` (#1709) in `url-utils.js` while this branch adds `getOutputHrefForStaticPath` (#1713), and both landed in the same spot at the end of the module. Kept both, no behaviour change on either side.

`build.default.dynamic-params-encoded` still passes 5/5 after the rebase.

Issue #1249, opened 2024-06-21
support `prerender` during development


thescientist13:
## Summary

While Greenwood supports [prerendering pages at build time](https://www.greenwoodjs.io/docs/configuration/#prerender), for example, having a "client" side component like this
```html
<! -- src/pages/index.html -->
<html>
  <head>
    <script type="module" src="./components/header.js"></script>
  </head>
  <body>
    <app-header></app-header>
</html>
```

This actually doesn't take place during development, so it is very common to see layout shifts and FOUC, as well as not having the same behavior during development and production!  🫨 

## Details

The main trick is what to do about the `<script>` tag in the head, because if you have a Light DOM component, e.g.
```js
// src/components/header.js
export default class Header extends HTMLElement {
  connectedCallback() {
    this.innerHTML = '<h1>Welcome to my website!</h1>'
  }
}

customElements.define('x-header', Header);
```

if you don't "guard" it like you would with declarative shadow DOM, e.g.
<details>
 ```js
export default class Header extends HTMLElement {
  connectedCallback() {
    if(this.innerHTML) return;

    this.innerHTML = '<h1>Welcome to my website!</h1>'
  }
}

customElements.define('x-header', Header);
```

You'll see duplicate rendering.  

// TODO get a screenshot!  @thescientist13 

So maybe could we strip the `<script>` tag for them if they have added `data-gwd-opt="static"`, or maybe we can do something clever like this?

```js
'use static'

export default class Header extends HTMLElement {
  connectedCallback() {
    if(this.innerHTML) return;

    this.innerHTML = '<h1>Welcome to my website!</h1>'
  }
}

customElements.define('x-header', Header);
----

Coming out of #1212, we should see if any of these observations at the time can be handled as part of this work (without scope creeping too much 😅 )
- revisit `<!-- greenwood-ssr-start -->` implementation
- In `preRenderCompilationWorker` could we consolidate into ONE SSR call? (instead serve + execute-route-module)
    - would this tie into the Workers refactor discussion?

PR #1716, opened 2026-07-18
fix(cli): #1715 split active content keys on first hyphen so hyphenated collections resolve

### LINKED ISSUE #1715: getContentByCollection returns empty for hyphenated collection names
### What happened?

Active content resolves content keys (e.g. `collection-my-posts`, `route-/my-blog`) by
splitting the key on **every** hyphen and taking the second piece. As a result any collection
whose name contains a hyphen silently resolves to nothing.

For `X-CONTENT-KEY: collection-my-posts` the key is split into
`["collection", "my", "posts"]` and only `"my"` is used as the collection name, so
`filterContentByCollection(graph, "my")` matches no pages and returns `[]`. The same defect
exists client-side in `getContentByCollection()` (via `key.split("-").pop()`), and route keys
like `route-/my-blog` are truncated to `/my`, which over-matches unrelated routes.

Kebab-case is the most common collection naming convention, so `getContentByCollection('my-posts')`
returns an empty array with no error to debug.

### Steps to reproduce

1. Create a Greenwood project with `activeContent: true`.
2. Add two pages that declare a hyphenated collection, e.g. frontmatter `collection: my-posts`.
3. Run `greenwood develop`.
4. Request the collection:
   `curl -s http://localhost:1984/___graph.json -H "X-CONTENT-KEY: collection-my-posts"`
5. Observe the response is `[]`. A hyphen-free collection (e.g. `collection-nav`) returns its
   pages correctly, confirming only hyphenated names are affected. The same empty result is
   returned by `getContentByCollection('my-posts')` in a non-prerendered build.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause:
- `packages/cli/src/plugins/resource/plugin-active-content.js:35` — `contentKey.split("-")` then
  `keyPieces[1]` is used as the collection/route value.
- `packages/cli/src/data/client.js:23` — `key.split("-").pop()` used for the same purpose when
  `prerender` is off.

jstockdi:
## Related Issue

Resolves #1715

## Documentation

N/A — no user-facing documentation changes (bug fix only; the `collection-<name>` /
`route-<route>` content key format is unchanged).

## Summary of Changes

1. [x] Split active content content keys on only the **first** hyphen so the value keeps any
   remaining hyphens, in `packages/cli/src/plugins/resource/plugin-active-content.js`
   (server `/___graph.json` handler) and `packages/cli/src/data/client.js` (client-side
   `getContentByCollection` / `getContentByRoute` when not prerendering).
2. [x] Hyphenated collection names (e.g. `my-posts`) now resolve to their pages instead of `[]`;
   hyphenated routes (e.g. `/my-blog`) keep their full value instead of being truncated to a
   prefix that over-matches.
3. [x] Behavior for hyphen-free keys (`graph`, `collection-nav`, `route-/blog`) is unchanged.
4. [x] Added regression test
   `packages/cli/test/cases/develop.config.active-content-collection-hyphen/` that serves a
   collection named `my-posts` across two pages and asserts
   `X-CONTENT-KEY: collection-my-posts` returns those pages (fails with `[]` on unpatched source).
5. [x] No breaking changes.

### DIFF
--- a/packages/cli/src/data/client.js
+++ b/packages/cli/src/data/client.js
@@ -18,7 +18,10 @@ async function getContentAsData(key = "") {
   } else if (CONTENT_STATE && !PRERENDER) {
     // if user is not prerendering, just fetch the entire graph but apply the same filtering
     const graph = await fetch("/graph.json").then((resp) => resp.json());
-    const value = key.split("-").pop();
+    // split on only the first "-" so hyphenated collection names / routes keep their hyphens
+    // https://github.com/ProjectEvergreen/greenwood/issues/1715
+    const delimiterIndex = key.indexOf("-");
+    const value = delimiterIndex === -1 ? "" : key.slice(delimiterIndex + 1);
 
     if (key === "graph") {
       return graph;
--- a/packages/cli/src/plugins/resource/plugin-active-content.js
+++ b/packages/cli/src/plugins/resource/plugin-active-content.js
@@ -32,7 +32,11 @@ class ContentAsDataResource {
   async serve(url, request) {
     const { graph } = this.compilation;
     const contentKey = request.headers.get("x-content-key") ?? "";
-    const keyPieces = contentKey.split("-");
+    // split on only the first "-" so hyphenated collection names / routes keep their hyphens
+    // https://github.com/ProjectEvergreen/greenwood/issues/1715
+    const delimiterIndex = contentKey.indexOf("-");
+    const type = delimiterIndex === -1 ? contentKey : contentKey.slice(0, delimiterIndex);
+    const value = delimiterIndex === -1 ? "" : contentKey.slice(delimiterIndex + 1);
     let status;
     let body;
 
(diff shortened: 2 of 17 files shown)


thescientist13 (commented):
Not a blocker per se, but I do try and avoid testing against implementation details.  I know it might be a little more work but for example in other content collection test cases, I usually do something through a component (often SSR'd) to validate the output that way.

You can see an example here
https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/build.config.prerender-collections

Thoughts?

jstockdi:
Makes sense — added `test/cases/build.config.prerender-collections-hyphen/` in b2215d81, modeled directly on `build.config.prerender-collections`: a prerendered build with SSR'd components calling `getContentByCollection("my-posts")` and `getContentByRoute("/my-blog")`, asserting the rendered links and the emitted `data-*.json` query files rather than the wire format.

Verified it pins the regression: against the unfixed key parsing, the collection component renders zero links (`collection-my-posts` resolved to the collection `"my"`). One nuance the run surfaced: the route-based case passes even unfixed, because the old split's `"/my"` still `startsWith`-matches `/my-blog/` — so the collection assertions are the real guard, and the route ones are there for symmetry.

I kept the original develop-mode test alongside since it exercises the dev-server `___graph.json` path the browser client hits when not prerendering — happy to drop it if you'd rather rely on the SSR case alone.

thescientist13:
> I kept the original develop-mode test alongside since it exercises the dev-server ___graph.json path the browser client hits when not prerendering — happy to drop it if you'd rather rely on the SSR case alone.

Ah yeah, that makes sense at this time.  Let me track this as an item in #1249 to circle back

thescientist13 (approved):
Thank you, will get this queued up for the next release line!

Issue #1334, opened 2024-12-01
content as data breaks when prerendering with a custom dev server port


thescientist13:
## Type of Change

Bug

## Summary

If you change the dev server port and run a build with `prerender: true`,  e.g.
```js
// greenwood.config.js
export default {
  prerender: true,
  devServer: {
    port: 8181
  }
}
```

You will get an error running `greenwood build`
```
TypeError [Error]: fetch failed
    at node:internal/deps/undici/undici:12618:11
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async getContentAsData (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/@greenwood/cli/src/data/client.js:9:7)
    at async getContentByCollection (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/@greenwood/cli/src/data/client.js:18:10)
    at async Header.connectedCallback (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/src/components/header/header.js:12:23)
    at async initializeCustomElement (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/wc-compiler/src/wcc.js:233:5)
    at async renderComponentRoots (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/wc-compiler/src/wcc.js:54:33)
    at async renderComponentRoots (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/wc-compiler/src/wcc.js:78:7)
    at async renderComponentRoots (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/wc-compiler/src/wcc.js:78:7)
    at async renderFromHTML (file:///Users/owenbuckley/Workspace/project-evergreen/www.greenwoodjs.dev/node_modules/wc-compiler/src/wcc.js:278:21) {
  cause: Error: connect ECONNREFUSED ::1:1984
      at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1555:16) {
    errno: -61,
    code: 'ECONNREFUSED',
    syscall: 'connect',
    address: '::1',
    port: 1984
  }
}
```

## Details

[...]

PR #1718, opened 2026-07-18
fix(cli): #1717 keep dev server alive when live reload port is in use (and harden livereload extension filter)

### LINKED ISSUE #1717: Second concurrent 'greenwood develop' crashes on hardcoded livereload port 35729
### What happened?

Running a second `greenwood develop` on the same machine (or leaving a stray dev server running) crashes the whole dev server with a raw, unhandled `EADDRINUSE` on port `35729` — a port the developer never configured.

The live reload server is created without a `port` option in `packages/cli/src/plugins/server/plugin-livereload.js`, so it always binds the hardcoded default `35729`, and no `error` handler is attached. When `35729` is already taken, the bind failure surfaces as an unhandled `'error'` event and the process exits with code 1 — even when a unique `devServer.port` is configured. The first line of output is misleading: the server prints `Started local development server at http://localhost:<port>` and only then dies.

```
Started local development server at http://localhost:1988
node:events:497
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE: address already in use :::35729
    ...
    at LiveReloadServer.start (file:///.../src/plugins/server/plugin-livereload.js:49:41)
Emitted 'error' event on Server instance at:
    ...
{ code: 'EADDRINUSE', errno: -98, syscall: 'listen', address: '::', port: 35729 }
```

There is also a latent bug in the same file: the "basic filter for false positives" on the collected extensions is `.filter((ext) => ext !== "*" || ext !== "")`, which is always `true` (`||` should be `&&`), so it filters nothing.

### Steps to reproduce

1. Create a minimal Greenwood project (a single `src/pages/index.html`) with `greenwood.config.js` setting a unique dev port, e.g. `export default { devServer: { port: 1988 } };`.
2. Occupy port `35729` before starting Greenwood (e.g. another `greenwood develop`, or `require("net").createServer().listen(35729)`).
3. Run `greenwood develop`.
4. Observe: the server prints `Started local development server at http://localhost:1988`, then immediately crashes with the unhandled `EADDRINUSE` dump above and exits code 1. The page is never served.

Expected: the dev server keeps serving pages and prints a single clear warning that live reload is disabled for this session.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/plugins/server/plugin-livereload.js:49` — `livereload.createServer(...)` is called with no `port` (always binds `35729`) and no `'error'` handler, so a busy port is an unhandled `'error'` event that kills the process. Bonus always-true filter at `plugin-livereload.js:46` (`||` should be `&&`).

jstockdi:
## Related Issue

Resolves #1717

## Documentation

N/A — no user-facing documentation changes. Behavior change is a graceful-degradation warning instead of a crash.

## Summary of Changes

1. [x] Attach an `'error'` handler to the live reload server in `packages/cli/src/plugins/server/plugin-livereload.js` so that an `EADDRINUSE` on the hardcoded live reload port (`35729`) no longer crashes the dev server. On conflict it now logs `live reload port 35729 is in use — live reload disabled for this session.` and keeps serving; any other error is re-thrown (unchanged behavior).
2. [x] Fix the always-true extension filter in the same file: `.filter((ext) => ext !== "*" || ext !== "")` → `&& ` (the `||` made the "basic filter for false positives" a no-op).
3. [x] Add regression test `packages/cli/test/cases/develop.default.livereload-port-in-use/` that occupies port `35729` before starting `develop` on a unique `devServer.port`, and asserts the dev server still serves the page (HTTP 200) instead of crashing.
4. [x] No breaking changes. When the live reload port is free, behavior is identical to before (existing `develop.*` specs unaffected: `develop.default` 115/115, `develop.default.hud-disabled` 26/26).

Note: this is the minimal graceful-degradation fix. A fully configurable live reload port (also templating the chosen port into the injected `<script src="http://localhost:35729/livereload.js">` snippet) is a larger enhancement left out of scope.

### DIFF
--- a/packages/cli/src/plugins/server/plugin-livereload.js
+++ b/packages/cli/src/plugins/server/plugin-livereload.js
@@ -42,7 +42,8 @@ class LiveReloadServer {
       ...customPluginsExtensions,
       ...this.compilation.config.devServer.extensions,
     ]
-      .filter((ext) => ext !== "*" || ext !== "") // basic filter for false positives
+      // https://github.com/ProjectEvergreen/greenwood/issues/1717
+      .filter((ext) => ext !== "*" && ext !== "") // basic filter for false positives
       .filter((ext, idx, array) => array.indexOf(ext) === idx) // dedupe
       .map((ext) => (ext.startsWith(".") ? ext.replace(".", "") : ext)); // trim . from all entries
 
@@ -62,6 +63,17 @@ class LiveReloadServer {
         );
       },
     );
+
+    // don't crash the whole dev server if the live reload port (35729) is already in use,
+    // e.g. a second concurrent `greenwood develop`; degrade gracefully and keep serving
+    // https://github.com/ProjectEvergreen/greenwood/issues/1717
+    liveReloadServer.on("error", (error) => {
+      if (error.code === "EADDRINUSE") {
+        console.warn("live reload port 35729 is in use — live reload disabled for this session.");
+      } else {
+        throw error;
+      }
+    });
 
     liveReloadServer.watch(userWorkspace.pathname);
   }
--- a/packages/cli/src/plugins/server/plugin-livereload.js
+++ b/packages/cli/src/plugins/server/plugin-livereload.js
@@ -42,7 +42,6 @@ class LiveReloadServer {
       ...customPluginsExtensions,
       ...this.compilation.config.devServer.extensions,
     ]
-      // https://github.com/ProjectEvergreen/greenwood/issues/1717
       .filter((ext) => ext !== "*" && ext !== "") // basic filter for false positives
       .filter((ext, idx, array) => array.indexOf(ext) === idx) // dedupe
       .map((ext) => (ext.startsWith(".") ? ext.replace(".", "") : ext)); // trim . from all entries
--- a/packages/cli/test/cases/develop.default.livereload-port-in-use/develop.default.livereload-port-in-use.spec.js
+++ b/packages/cli/test/cases/develop.default.livereload-port-in-use/develop.default.livereload-port-in-use.spec.js
@@ -1,0 +1,110 @@
+/*
+ * Use Case
+ * Run Greenwood develop command when the live reload port (35729) is already in use
+ * by another process (e.g. a second concurrent `greenwood develop`).
+ *
+ * User Result
+ * Should start the development server on the configured devServer.port and keep serving
+ * pages, gracefully disabling live reload with a warning instead of crashing the process.
+ *
+ * User Command
+ * greenwood develop
+ *
+ * User Config
+ * devServer: {
+ *   port: 1988,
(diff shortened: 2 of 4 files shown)


packages/cli/src/plugins/server/plugin-livereload.js
```diff
@@ -42,7 +42,8 @@ class LiveReloadServer {
       ...customPluginsExtensions,
       ...this.compilation.config.devServer.extensions,
     ]
-      .filter((ext) => ext !== "*" || ext !== "") // basic filter for false positives
+      // https://github.com/ProjectEvergreen/greenwood/issues/1717
```
thescientist13: maybe I'm missing it, but what changed here?

thescientist13 (commented):
I wonder if we should make a new issue feature task for auto-detecting / selecting a new port?  Something along the lines of [these issues](https://github.com/ProjectEvergreen/greenwood/issues/1717#issuecomment-5040819949)?

packages/cli/src/plugins/server/plugin-livereload.js
jstockdi: The operator: `ext !== "*" || ext !== ""` is always true (any value differs from at least one of the two), so the filter never removed anything — `*` and empty entries flowed through to the livereload extension list. `&&` makes the filter do what the comment says. Happy to split it into its own PR if you'd rather keep this one to the crash fix.

jstockdi:
Agreed — filed #1734 for auto-selecting a free live reload port, referencing #71 and #1334 per your links. The wrinkle specific to live reload is that the injected client snippet hardcodes `http://localhost:35729/livereload.js`, so port selection has to flow into the snippet too; noted that in the issue.

This PR stays scoped to not crashing: the EADDRINUSE guard becomes the last-resort fallback once auto-selection lands.

packages/cli/src/plugins/server/plugin-livereload.js
thescientist13: oh right, sorry I totally missed the switch from `||` -> `&&` and yeah, that totally makes sense 🤦‍♂️ 

I think it's OK as part of this PR, let me tweak the PR title

thescientist13 (approved):

thescientist13:
thanks, will get this in the next release line 👍

packages/cli/src/plugins/server/plugin-livereload.js
thescientist13: not a big deal, but we probably don't need a ticket comment here

jstockdi: Good call — dropped in `26446cee`. The `&&` operator fix itself is unchanged, only the ticket comment went.

Issue #1623, opened 2025-12-14
support dynamic directory routing capabilities for SSR pages and API routes


thescientist13:
### Motivation

Coming out adding support for [dynamic routing](https://github.com/ProjectEvergreen/greenwood/issues/882), it would be good to add support for being able to have dynamic segments for directories in file based routing paths, e.g.

```js
src/
  pages/
    [org]/
      [team]/
        [board].js
```

Would match / yield the following params for this kind of route; `/projectevergreen/greenwood/team1/milestone1-0/`
```js
{
  org: 'projectevergreen',
  team: 'greenwood',
  board: 'milestone1-0'
}
```

### Technical Design

Inspired by these implementations
- https://docs.astro.build/en/guides/routing/#rest-parameters
- https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#catch-all-segments

### Additional Context

Should also make sure to account for adapter plugins

thescientist13:
I could see this being a 1.x feature...

jstockdi:
One case to account for when this lands, spun out of PR #1720 / issue #1719: today a page like `src/pages/[a]/[b].js` isn't rejected up front — the build emits a literal `public/[a]/` directory and then dies in the optimize pass with `TypeError: Cannot read properties of undefined (reading 'resources')`, which points nowhere near the cause. When multi-segment support is implemented here, `getDynamicSegmentsFromRoute` (`packages/cli/src/lib/url-utils.js`) will also need to return one key per bracket pair — it currently derives a single `segmentKey` from the first `[param]` in the route.

PR #1720, opened 2026-07-18
fix(cli): #1719 derive dynamic route segment key from the route bracket

### LINKED ISSUE #1719: Dynamic route param name containing the file extension is mis-parsed; nested routes crash unhelpfully
### What happened?

Two related defects in dynamic route segment parsing (`getDynamicSegmentsFromRoute`, `packages/cli/src/lib/url-utils.js`):

**1. Param name containing the file extension substring silently produces wrong output.**
The segment key is computed with an unanchored `relativePagePath...replace(extension, "")`. When the param name contains the extension as a substring, the key is mangled — e.g. `[json].js` yields the key `onjs` instead of `json` (any TypeScript param containing `ts`, such as `[posts].ts`, is affected too). The URL pattern (`/:json/`) and the key then disagree, so `getStaticRouteFromDynamicRoute` substitutes a key that isn't in the route. The build SUCCEEDS (exit 0) but emits a literal `public/[json]/index.html` with garbage params (`params={"json":"[json]"}`) instead of `public/alpha/index.html` with `params={"json":"alpha"}`. No warning of any kind.

**2. A nested/multi-segment dynamic route crashes with an unrelated `TypeError`.**
`route.replace("[", ":").replace("]", "")` only converts the first bracket pair, so `src/pages/[a]/[b].js` becomes the URL pattern `/:a/[b]/` (the second segment is left as a literal `[b]`). A literal `public/[a]/` directory is written and the build then dies in the optimize pass with `TypeError: Cannot read properties of undefined (reading 'resources')` (`plugins/resource/plugin-standard-html.js`), which gives no hint that the nested dynamic segment is the cause.

### Steps to reproduce

Defect 1 (silent corruption):

1. Create `src/pages/[json].js`:
   ```js
     return [{ params: { json: "alpha" } }];
   }
     return `<h2>params=${JSON.stringify(params)}</h2>`;
   }
   ```
2. Run `greenwood build`.
3. Observe: build succeeds but emits `public/[json]/index.html` (with `params={"json":"[json]"}`); `public/alpha/index.html` is never generated.

Defect 2 (build crash):

1. Create `src/pages/[a]/[b].js` with a `getStaticPaths` returning `{ params: { a: "x", b: "1" } }`.
2. Run `greenwood build`.
3. Observe: a literal `public/[a]/` directory is written, then the build aborts (exit 1) with `TypeError: Cannot read properties of undefined (reading 'resources')`.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/lib/url-utils.js` `getDynamicSegmentsFromRoute` — line 3 (`route.replace("[", ":").replace("]", "")`, first-occurrence only) and lines 4-9 (unanchored `.replace(extension, "")` for the segment key). Full multi-segment (nested `[a]/[b]`) support is a larger enhancement spanning the graph, prerender, bundle, execute-route-module, static-router and adapters; this report/fix makes single-segment routes correct and turns the multi-segment case into a clear, actionable error rather than a confusing downstream crash.

jstockdi:
## Related Issue

Resolves #1719

## Documentation

## Summary of Changes

1. [x] `getDynamicSegmentsFromRoute` (`packages/cli/src/lib/url-utils.js`): derive the segment key from the bracket in the route (`/\[([^\]]+)\]/`) instead of an unanchored `.replace(extension, "")`, so a param name that contains the extension substring (e.g. `[json].js`, `[posts].ts`) no longer produces a mangled key. `[json].js` now builds correctly to `/alpha/index.html` with `params={"json":"alpha"}`.
2. [x] Trim the now-unused arguments passed to `getDynamicSegmentsFromRoute` at both call sites in `packages/cli/src/lifecycles/graph.js`.
3. [x] Regression coverage folded into the existing dynamic-routing test cases (a `[json].js` fixture page with `getStaticPaths` in each):
   - `develop.dynamic-routing-static-paths` — asserts the dev server resolves `/beta/` with its own params.
   - `serve.dynamic-routing` — asserts the build emits a static path per `getStaticPaths` entry with the correct params, does NOT emit a literal `/[json]/` directory, and that the prod server serves each static path with its own resolved params.

### Scope / follow-up

Nested/multi-segment dynamic routes (`[category]/[slug]`) remain unsupported and are intentionally untouched here — that work is tracked in #1623. This PR only makes single-segment param names parse correctly regardless of what substring they contain.

No breaking changes. Existing dynamic-routing / `getStaticPaths` / SSR behavior is unchanged (verified against the existing dynamic-routing, getStaticPaths, and SSR specs for build, serve, and develop).


thescientist13 (commented):
Could you help me navigate this issue / PR a bit?

For the second defect at least, I think it's an issue with using segments + dynamic routing for for directories?  If so, that's still a feature that [needs to be implemented](https://github.com/ProjectEvergreen/greenwood/issues/1623), so not a defect, per se.

I'm not sure I quite understand the first defect though.  Could you elaborate a bit more on it?

Thanks!

jstockdi:
Sure — the two defects are separate, and you're right that the second borders #1623.

**Defect 1 is a bug in already-supported single-segment routes.** The segment key is computed with an unanchored `relativePagePath.replace(extension, "")` in `getDynamicSegmentsFromRoute` (`packages/cli/src/lib/url-utils.js`), which strips the extension substring from *anywhere* in the path — including inside the param name. So `src/pages/[json].js` yields the key `onjs` instead of `json` (same for any TS param containing `ts`, e.g. `[posts].ts`). The URL pattern (`/:json/`) and the key then disagree, and the build silently succeeds with a literal `public/[json]/index.html` and `params={"json":"[json]"}` instead of `public/alpha/index.html`. Exit 0, no warning. Repro is in #1719; the regression test is `test/cases/build.default.dynamic-routing-param-name-ext/`. This is the core fix of the PR: derive the key from the bracket in the route itself.

**Defect 2 — agreed, multi-segment support is the #1623 feature, and this PR doesn't implement it.** What it changes is only the failure mode: today `src/pages/[a]/[b].js` writes a literal `public/[a]/` directory and then dies in the optimize pass with `TypeError: Cannot read properties of undefined (reading 'resources')`, which points nowhere near the cause. The PR detects the unsupported shape at graph time and throws `Multiple dynamic route segments are not currently supported (found in ...)`. When #1623 lands, that guard is the natural place to remove.

Happy to split the fail-fast into its own PR (or drop it and note the limitation on #1623) if you'd rather keep this one to the key-parsing fix — say the word.

jstockdi:
Gentle bump on this one — curious whether the breakdown above answers your questions on the two defects. Happy to split the multi-segment fail-fast into its own PR (or drop it in favor of a note on #1623) if that makes the key-parsing fix easier to take.

thescientist13:
> So src/pages/[json].js yields the key onjs instead of json 

Oooooh, I get it now.  The extension substitution _js_ is applying the dynamic segment `[json]`.  Makes total sense now, good catch!

thescientist13 (changes requested):
Thanks for explaining the first issue, I totally get it now.  😅 

I would like to add for a couple feedback items
1. Let's keep the dynamic directory handling out of this changeset, since that's not really a support feature at all, but will be
1. Let's put the `[json]` test into the [existing dynamic routing test case](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/develop.dynamic-routing-static-paths), so everything can be found / tested / validate in one place

For the dynamic directory item, if you want to leave a comment in #1623 so we make sure to account for that case, that would be great! 🙏

jstockdi:
Both done in 13c908f8:

1. Dropped the multi-segment fail-fast and its test case — the changeset is now just the segment-key parsing fix in `getDynamicSegmentsFromRoute`. Left a note on #1623 (https://github.com/ProjectEvergreen/greenwood/issues/1623#issuecomment-5151489757) so the dynamic-directory case gets accounted for there.
2. Moved the `[json]` test into `develop.dynamic-routing-static-paths`. To make the wrong key observable in develop mode the fixture uses two static paths plus `getStaticParams` — with the old `onjs` key, the static-path lookup in `execute-route-module.js` matches the first entry regardless of the requested value, so `/beta/` renders the `alpha` entry's content. Verified the new block fails against master (`expected 'Alpha' to equal 'Beta'`) and passes with the fix.

packages/cli/src/lifecycles/graph.js
```diff
@@ -111,12 +111,7 @@ const generateGraph = async (compilation) => {
           // is there a better way to detect for isolation export without having to actually import() the module?
           const contents = await fs.readFile(filenameUrl, "utf8");
           const isolation = contents.indexOf("export const isolation = true;") >= 0;
-          const { segmentKey, dynamicRoute } = getDynamicSegmentsFromRoute({
-            route,
```
thescientist13, line 117: wow, pretty clean ✨

thescientist13 (commented):
Thanks for the clean up, but looks like we lost the build test case in the process?  I think it would be good to test on both paths here just to ensure expected behavior for `greenwood develop` and `greenwood build` 👍

jstockdi:
Restored in 6a01d0c7 — the build test case is back under its original name, with the fixture upgraded to mirror the develop case on this branch (alpha/beta static paths from `getStaticPaths`), so `greenwood develop` and `greenwood build` now cover the identical scenario. Assertions check the emitted static paths, that no literal `[json]/index.html` is written, and that each page renders its own params. Develop + build specs both passing; the multi-segment fail-fast stays removed.

packages/cli/test/cases/build.default.dynamic-routing-param-name-ext/build.default.dynamic-routing-param-name-ext.spec.js
```diff
@@ -0,0 +1,86 @@
...
+import { expect } from "chai";
+import fs from "node:fs/promises";
+import { JSDOM } from "jsdom";
+import path from "node:path";
+import { getOutputTeardownFiles } from "../../../../../test/utils.js";
+import { Runner } from "gallinago";
+import { fileURLToPath } from "node:url";
+
+// https://github.com/ProjectEvergreen/greenwood/issues/1719
+describe("Build Greenwood With: ", function () {
```
thescientist13: Maybe this test could get folded in the [server test case too](https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/serve.dynamic-routing), just like we did for develop?  Then we can cover not only the build output, but also the response from the prod server, all in one place 💯

thescientist13:
I think we should also update the PR title to better reflect the contents / scope of the PR

jstockdi:
Folded the build case into `serve.dynamic-routing` as suggested — the `[json].js` fixture now lives there, with the build-output assertions (a static path emitted per `getStaticPaths` entry, no literal `/[json]/` directory) plus response assertions against the prod server for both paths, mirroring the develop setup. The standalone `build.default.dynamic-routing-param-name-ext` case is removed.

Re-verified the folded assertions catch the bug: with `url-utils.js`/`graph.js` reverted to master they fail exactly as expected (literal `/[json]/index.html` emitted, `/alpha/`–`/beta/` missing, serve responds 500); with the fix all 6 specs in the case pass.

Also retitled the PR and refreshed the description to match the reduced scope — the multi-segment fail-fast is out of this changeset, with nested/multi-segment support tracked in #1623.

thescientist13 (approved):
Verified on my end, and looks good!

Issue #1133, opened 2023-08-03
implement a more idiomatic (HTML) templating solution


thescientist13:
## Type of Change
Enhancement

## Summary
Coming out of #1126 , it's clear that how Greenwood handles various aspects of its templating, primarily for HTML, is a little too naive / brittle.

For example, Greenwood uses placeholders for injecting content into an HTML template
```html
<html>
  <body>
    <content-outlet><content-outlet>
  </body>
</html>
```

Would get handled like this in the CLI
```sh
body = body.replace(/\<content-outlet>(.*)<\/content-outlet>/s, processedMarkdown.contents);
```

This issue is that [`$1` is used for matching](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement) when used in conjunction with the String `.replace` function.

So we have to keep doing things like
```js
body = body.replace(/\<content-outlet>(.*)<\/content-outlet>/s, processedMarkdown.contents.replace(/\$/g, '$$$'));
```

## Details
So rather than keep suffering through these issues...
- #1126 
- #743
- #657

Since we are already using an [HTML parser](https://www.npmjs.com/package/node-html-parser), I see it supports methods like [`set_content`](https://github.com/taoqf/node-html-parser#set_contentcontent-string--node--node) which could a much more practical way to approach this problem.

So basically any [usage of `.replace`](https://github.com/search?q=repo%3AProjectEvergreen%2Fgreenwood%20.replace&type=code), for HTML at minimum, would be a candidate for refactoring.

thescientist13:
Seeing similar issues as per #1241

PR #1722, opened 2026-07-18
fix(cli): #1721 rewrite import attributes polyfill with ast-range splicing

### LINKED ISSUE #1721: Import-attributes polyfill corrupts source via whole-file string replacement (comments, string literals, withX bindings, dynamic imports)
### What happened?

With the import attributes polyfill enabled:

```js
// greenwood.config.js
export default {
  polyfills: {
    importAttributes: ["css", "json"],
  },
};
```

the standard JavaScript resource plugin (`packages/cli/src/plugins/resource/plugin-standard-javascript.js`, `preIntercept()`) rewrites `import ... with { type: "css" | "json" }` into Greenwood's `?polyfill=type-<attr>` form using whole-file string operations (`String.prototype.replace(line, ...)`, `String.prototype.replace(value, ...)` and `line.split("with")[0]`). Because those operations search the entire file for the first textual occurrence and split on the literal word `with`, they corrupt code that merely *looks* like an import, miss real imports, and crash on syntax they do not understand. Eight distinct failures result:

1. A commented-out copy of the import line is rewritten instead of the real import, and a `;\n` is injected into the comment, splitting it into live code → `SyntaxError`.
2. `with { type: ... }` appearing inside a string literal is rewritten, injecting `;\n` mid-string → unterminated string constant.
3. An import whose binding name contains `with` (e.g. `withTheme`) is destroyed by `line.split("with")[0]`, collapsing the statement to `import ;` → `SyntaxError`.
4. Two imports of the same specifier: the first gets the `?polyfill=...` query appended twice and the second is left un-polyfilled (first-occurrence replace).
5. A comment that mentions the specifier path receives the `?polyfill` query while the real import is left un-polyfilled (silent — no error, but the module is not polyfilled).
6. A multi-line `with {\n  type: "css"\n}` clause is never matched (the guard strips only spaces, not newlines) → silent no-op, import not polyfilled.
7. Dynamic `import(specifier, { with: { type } })` is never handled (only `ImportDeclaration` nodes are visited).
8. A file using the legacy `assert { type }` keyword makes `acorn.parse` throw with no `try`/`catch`, so the whole file 500s / fails to build.

### Steps to reproduce

1. Create a Greenwood project with `polyfills: { importAttributes: ["css", "json"] }` in `greenwood.config.js`.
2. Add a served JS module that contains any of the tricky-but-valid inputs, e.g.:
   ```js
   // import shadow from "./theme.css" with { type: "css" }; older example
   import withTheme from "./theme.css" with { type: "css" };
   import a from "./shared.css" with { type: "css" };
   import b from "./shared.css" with { type: "css" };
   import config from "./data.json"
     with {
       type: "json"
     };
   const dynamic = await import("./dynamic.css", { with: { type: "css" } });
   ```
3. Run `greenwood develop` (or `greenwood build`) and request `/main.js`.
4. Observe: the commented-out line is mutated and a `;` is injected into the comment; the `withTheme` import collapses to `import ;`; only one of the two `./shared.css` imports is polyfilled (and it is doubled); the multi-line `./data.json` import is not polyfilled; the dynamic import is not polyfilled. Adding a separate file with `assert { type: "json" }` returns a 500 (the parse throws).

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/plugins/resource/plugin-standard-javascript.js`, `preIntercept()` (v0.34.0, ~lines 44-61) — the rewrite uses whole-file `String.prototype.replace(line, ...)` / `String.prototype.replace(value, ...)` (first-occurrence over the entire file) and `line.split("with")[0]` (splits on the first literal `with`, including inside binding names), only visits `ImportDeclaration` nodes, matches the attributes clause by whitespace-normalized substring (spaces only), and calls `acorn.parse` without a `try`/`catch`.

jstockdi:
## Related Issue

Resolves #1721

## Documentation

N/A — no user-facing configuration or behavior changes; this fixes the existing
`polyfills.importAttributes` feature so it behaves as already documented.

## Summary of Changes

Rewrites the import attributes polyfill in `packages/cli/src/plugins/resource/plugin-standard-javascript.js` (`preIntercept()`) to splice by AST character offsets instead of whole-file string operations.

1. [x] Rewrite each specifier at its own `node.source.start`/`node.source.end` range (append `?polyfill=type-<attr>` before the closing quote) and remove the `with { ... }` / `assert { ... }` clause by its parsed AST range, applying edits by descending offset so earlier splices do not shift later ones. No more whole-file `String.replace` and no more `line.split("with")`.
2. [x] Detect the attribute from the parsed AST (`node.attributes` with `key.name === "type"` and a `value.value` in the configured list) rather than by whitespace-normalized substring matching. Fixes: commented-out import copies being rewritten, `with { type }` inside string literals being rewritten, binding names containing `with` being truncated, comments that mention the specifier path being mutated, and multi-line attributes clauses being silently skipped.
3. [x] Handle each import independently by offset, so two imports of the same specifier each get exactly one `?polyfill` query (no double query on the first, none skipped on the second).
4. [x] Add an `ImportExpression` visitor that rewrites the dynamic `import(specifier, { with: { type } })` form and drops its options argument by AST range.
5. [x] Wrap `acorn.parse` in `try`/`catch` and pass the source through unmodified on parse failure, so syntax the polyfill cannot parse (e.g. the legacy `assert` keyword, which acorn's `ecmaVersion: "latest"` rejects) can no longer 500 / brick the file.
6. [x] Adds regression test `packages/cli/test/cases/develop.config.polyfills-import-attributes-edge-cases/` — a `greenwood develop` case whose served `main.js` exercises a commented-out import copy, a `withTheme` binding, two imports of the same specifier, a multi-line attributes clause, a dynamic import, and a specifier (`./banner.css`) that appears verbatim in a string literal before its own real import with no shadowing comment. Each of these is asserted to survive/transform correctly, and every assertion except the two content-type/200 sanity checks is verified to FAIL against genuine `origin/master` source and PASS with the fix (patched: 11 passing; pre-fix: 3 passing / 8 failing). The string-literal case is a true regression assertion: the old whole-file `replace(value/line, …)` mangles the `./banner.css` string literal, so `should not mangle a string literal that duplicates a real import's specifier` fails pre-fix and passes post-fix. The commented-out-import assertion likewise proves the comment-mutation case (#5). A separate `legacy.js` using the removed `assert` keyword asserts graceful pass-through (200, unmodified) where the old code returned a 500.

**No breaking changes.** Ordinary imports (no attributes) are untouched, and the existing `develop.config.polyfills-import-attributes` and `loaders-*` import-attributes specs continue to pass unchanged.

**Known limitation (documented):** a file containing the *removed* `assert` import-attributes keyword is passed through unmodified as a safety net (acorn `ecmaVersion: "latest"` cannot parse `assert`, and there is no acorn option to re-enable it). That means any *other* imports in the same file are also not polyfilled — but the file no longer crashes the dev server / build. The modern `with` keyword (including single-quoted `type: 'json'`) is fully handled.

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-standard-javascript.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-javascript.js
@@ -46,19 +46,84 @@ class StandardJavaScriptResource {
     const body = await response.clone().text();
     let polyfilled = body;
 
-    walk.simple(acorn.parse(body, ACORN_OPTIONS), {
-      async ImportDeclaration(node) {
-        const line = body.slice(node.start, node.end);
-        const { value } = node.source;
+    // rewrite `import ... with { type: "css" | "json" }` (static and dynamic) into
+    // Greenwood's `?polyfill=type-<attr>` query form, using each node's own AST character
+    // offsets so only the matched specifier and its attributes clause are edited, never
+    // comments, string literals, or same-named bindings, and each import is handled
+    // independently; unparseable sources (e.g. legacy `assert`) pass through untouched
+    // https://github.com/ProjectEvergreen/greenwood/issues/1721
+    const edits = [];
+    const keyName = (key) => key.name ?? key.value;
+    const polyfillAttribute = (type) =>
+      typeof type === "string" && polyfills.importAttributes.includes(type) ? type : null;
+    const rewriteSpecifier = (source, attribute) => {
+      const raw = body.slice(source.start, source.end);
+      const quote = raw.slice(-1);
 
-        polyfills.importAttributes.forEach((attribute) => {
-          if (line.replace(/ /g, "").replace(/"/g, "'").includes(`with{type:'${attribute}'}`)) {
-            polyfilled = polyfilled.replace(line, `${line.split("with")[0]};\n`);
-            polyfilled = polyfilled.replace(value, `${value}?polyfill=type-${attribute}`);
+      edits.push({
+        start: source.start,
+        end: source.end,
+        replacement: `${raw.slice(0, -1)}?polyfill=type-${attribute}${quote}`,
+      });
+    };
+
+    try {
+      walk.simple(acorn.parse(body, ACORN_OPTIONS), {
+        ImportDeclaration(node) {
+          const attributes = node.attributes ?? [];
+          const typeAttribute = attributes.find((attribute) => keyName(attribute.key) === "type");
+          const attribute = typeAttribute && polyfillAttribute(typeAttribute.value.value);
+
+          if (!attribute) {
+            return;
           }
-        });
-      },
-    });
+
+          rewriteSpecifier(node.source, attribute);
+          // drop the trailing `with { ... }` / `assert { ... }` clause by its AST range
+          const closeBrace = body.indexOf("}", attributes[attributes.length - 1].end);
+          edits.push({ start: node.source.end, end: closeBrace + 1, replacement: "" });
+        },
+        ImportExpression(node) {
+          if (node.options?.type !== "ObjectExpression") {
+            return;
+          }
+
+          const clause = node.options.properties.find((property) =>
+            ["with", "assert"].includes(keyName(property.key)),
+          );
+
+          if (clause?.value.type !== "ObjectExpression") {
+            return;
+          }
+
+          const typeProperty = clause.value.properties.find(
+            (property) => keyName(property.key) === "type",
+          );
+          const attribute = typeProperty && polyfillAttribute(typeProperty.value.value);
+
+          if (!attribute) {
+            return;
+          }
+
+          rewriteSpecifier(node.source, attribute);
+          // drop the trailing `, { with: { type } }` options argument by its AST range
+          edits.push({ start: node.source.end, end: node.options.end, replacement: "" });
+        },
+      });
+
+      // apply edits by descending offset so earlier splices do not shift later ones
+      polyfilled = edits
+        .sort((a, b) => b.start - a.start)
+        .reduce(
+          (source, { start, end, replacement }) =>
+            source.slice(0, start) + replacement + source.slice(end),
+          body,
+        );
+    } catch (e) {
+      // never brick the build on syntax the polyfill cannot parse; serve it unchanged
+      console.warn(`Unable to parse ${url.pathname} for import attributes, skipping polyfill.`, e);
+      polyfilled = body;
+    }
 
     return new Response(polyfilled, {
       headers: {
--- a/packages/cli/src/plugins/resource/plugin-standard-javascript.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-javascript.js
@@ -79,7 +79,7 @@ class StandardJavaScriptResource {
           }
 
           rewriteSpecifier(node.source, attribute);
-          // drop the trailing `with { ... }` / `assert { ... }` clause by its AST range
+          // drop the trailing `with { ... }` clause by its AST range
           const closeBrace = body.indexOf("}", attributes[attributes.length - 1].end);
           edits.push({ start: node.source.end, end: closeBrace + 1, replacement: "" });
         },
@@ -88,8 +88,9 @@ class StandardJavaScriptResource {
             return;
           }
 
-          const clause = node.options.properties.find((property) =>
-            ["with", "assert"].includes(keyName(property.key)),
+          // only the standard `with` key is supported; legacy `assert` is not
+          const clause = node.options.properties.find(
+            (property) => keyName(property.key) === "with",
           );
 
           if (clause?.value.type !== "ObjectExpression") {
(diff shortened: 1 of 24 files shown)


packages/cli/src/plugins/resource/plugin-standard-javascript.js
```diff
@@ -46,19 +46,84 @@ class StandardJavaScriptResource {
...
+          const closeBrace = body.indexOf("}", attributes[attributes.length - 1].end);
+          edits.push({ start: node.source.end, end: closeBrace + 1, replacement: "" });
+        },
+        ImportExpression(node) {
+          if (node.options?.type !== "ObjectExpression") {
+            return;
+          }
+
+          const clause = node.options.properties.find((property) =>
+            ["with", "assert"].includes(keyName(property.key)),
```
thescientist13: assert shouldn't be valid / supported, only `with` is valid now from a standards perspective

thescientist13 (commented):
So is the main issue here just having a commented out import attribute line?

packages/cli/src/plugins/resource/plugin-standard-javascript.js
jstockdi: Agreed — dropped in 05d479b0. The dynamic-import matcher now accepts only a `with` key.

Static `assert` never reached the rewrite path: acorn rejects it at parse, so the file serves unchanged. The edge-cases spec pins that pass-through behavior (`src/legacy.js`), which I've kept — it verifies `assert` is unsupported rather than supported.

jstockdi:
No — the commented-out import is just the trigger reported in #1721. The root defect is that the old rewrite ran two whole-file `String.replace` calls keyed on the import's source text, so *any* other occurrence of that text got edited too. The edge-cases fixture (`test/cases/develop.config.polyfills-import-attributes-edge-cases/src/main.js`) pins each failure mode:

- a commented-out copy of an import — the replace hit the comment instead of the real import
- the import line quoted inside a string literal — the string got a `?polyfill` query and a stray `;` injected
- two imports of the same specifier — first-occurrence replace left the second unrewritten
- a multi-line `with { ... }` clause — the single-line string match never fired
- a dynamic `import("./a.css", { with: { type: "css" } })` — not handled at all

The new implementation splices by each AST node's own character range, so only the matched specifier and its attributes clause are edited. Per your inline note, `assert` matching is now removed as well (05d479b0).

thescientist13 (commented):
Ah yeah, the current implementation is indeed quite naive / brittle.  😥 

(I've been wanting to do more refactoring for this for [HTML too](https://github.com/ProjectEvergreen/greenwood/issues/1133) now that I've gotten more familiar with ASTs)

----

I do think we should create a version of this that generates a build, so in addition to all this, so we can validate that all these edge cases still work when bundling is involved, where the paths would get a hash included in their output.  So would want to make sure those come out as expected too.  👍

jstockdi:
Added in dd8ee33e — `build.config.polyfills-import-attributes-edge-cases`, same fixture as the develop case but run through `greenwood build`, asserting on the hashed bundle output:

- CSS and JSON land inlined in the hashed `main.<hash>.js` bundle exactly once each — including the twice-imported `shared.css`, which rollup dedupes to a single constructable stylesheet
- the two string literals that quote import syntax come through the minifier verbatim (the count-exact `with { type: ... }` assertion pins that those are the only occurrences left)
- the dynamic `import("./dynamic.css", { with: ... })` becomes `import("./dynamic.<hash>.js")` pointing at its own hashed chunk, which exports the stylesheet
- the page references the hashed bundle

One deliberate omission: `legacy.js` (the `assert` pass-through) stays develop-only — rollup's parser rejects the removed `assert` keyword outright, so it can't go through a bundle. For a sanity check I also ran this new case against the current master implementation: the build itself fails (`SyntaxError` out of the `greenwood-import-meta-url` transform on the mangled source), so the case discriminates cleanly.

packages/cli/test/cases/develop.config.polyfills-import-attributes-edge-cases/src/legacy.js
```diff
@@ -0,0 +1,4 @@
+import data from "./data.json" with { type: "json" };
```
thescientist13: i noticed you [calling this one out](https://github.com/ProjectEvergreen/greenwood/pull/1722#issuecomment-5151502437), and I not sure it's valuable to have a legacy example, especially as we can't even run it through a build.

I would much rather guide developers to use the current standard of `with` anyway, IMO.

jstockdi: Dropped in c088bb9d — removed `legacy.js` and its `config.json` (referenced only by the legacy fixture) along with the spec's legacy `assert` block. Both the develop and build edge-case suites still pass (9/9 each), and all `with`-syntax coverage stays as-is.

thescientist13 (approved):
This looks good!

What do you think about renaming the test case files a bit, just to be a little clearer what the test case is for?  Edge cases feels a little too broad I suppose, and it's only really clear if you go into the description of the case itself.

PR #1724, opened 2026-07-18
fix(cli): #1723 gate import.meta.url bundling on js content-type so unsupported resource skip acorn parsing

### LINKED ISSUE #1723: Direct .wasm import crashes build with a cryptic acorn SyntaxError from greenwood-import-meta-url
### What happened?

Referencing a `.wasm` file in a browser script with a direct ES import — e.g.:

```js
import * as wasm from "./add.wasm";
```

crashes `greenwood build` with a cryptic acorn `SyntaxError` that points at Greenwood's own internal plugin code, giving the user no hint of what went wrong:

```
SyntaxError: Unexpected character ' ' (1:0)
    at Object.transform (.../packages/cli/src/config/rollup.config.js:327:25)
  code: 'PLUGIN_ERROR',
  plugin: 'greenwood-import-meta-url',
  hook: 'transform',
  id: '.../src/scripts/add.wasm'
```

Root cause: the `greenwood-import-meta-url` Rollup transform runs `acorn.parse(await response.text(), ...)` on every module some resource plugin can serve. `canTransform` is set `true` whenever a plugin's `shouldServe` matches the resource — **regardless of the served content-type** — and the transform then parses the served body without checking its type. So even though `.wasm` is ultimately served as `application/wasm`, its **binary** body (UTF-8-lossy-decoded) is still fed to the JavaScript parser and acorn throws.

The same crash class affects any binary resource matched by a standard resource plugin (`.png`, `.woff2`, etc.).

Note the asymmetry: `greenwood develop` serves the same project fine (the dev server returns the wasm with `application/wasm`); only `build` hard-crashes, and it does so with a parser error that appears to come from Greenwood internals rather than the user's code.

There is also a latent operator-precedence bug on the neighboring content-type check (`rollup.config.js:366`): `response?.headers?.get("content-type") || "".indexOf("text/javascript") >= 0` — `||` binds before `.indexOf`, so the expression is truthy for any content-type.

### Steps to reproduce

1. Create a Greenwood project with a browser script that directly imports a `.wasm` module:
   - `src/pages/index.html` with `<script type="module" src="../scripts/main.js"></script>`
   - `src/scripts/main.js` containing `import * as wasm from "./add.wasm";`
   - `src/scripts/add.wasm` (any valid WebAssembly module)
2. Run `greenwood build`.
3. The build aborts (exit 1, no `public/` output) with the acorn `SyntaxError` shown above, attributed to the `greenwood-import-meta-url` plugin.

The supported pattern `new URL("./add.wasm", import.meta.url)` (with `fetch` + `WebAssembly.instantiateStreaming`) builds correctly and emits a content-hashed `.wasm` asset — the direct-import syntax is what crashes.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

- `packages/cli/src/config/rollup.config.js:327` — `acorn.parse(await response.text(), ...)` runs whenever `canTransform` is true.
- `packages/cli/src/config/rollup.config.js:295-300` — `canTransform` is set `true` on serve regardless of content-type.
- `packages/cli/src/plugins/resource/plugin-node-modules.js:43-57` — serves any existing `file:` resource as `text/javascript`.
- `packages/cli/src/plugins/resource/plugin-standard-wasm.js:16-18` — serves `.wasm` as `application/wasm`.
- `packages/cli/src/config/rollup.config.js:366` — operator-precedence bug on the content-type check.

jstockdi:
## Related Issue

Resolves #1723

## Documentation

N/A — no user-facing documentation changes. (The supported wasm pattern, `new URL("./x.wasm", import.meta.url)`, is unchanged; this PR only fixes the crash for the unsupported direct-import syntax.)

## Summary of Changes

1. [x] Gate the `greenwood-import-meta-url` Rollup `transform` so it only parses a module with acorn when the **final served response content-type is JavaScript**. Binary resources (`.wasm`, and by extension `.png` / `.woff2` / etc.) are no longer fed to the JS parser, so `greenwood build` no longer crashes with a cryptic acorn `SyntaxError` attributed to internal plugin code.
   - The previous `canTransform` flag was set `true` whenever *any* resource plugin served the module; because the node-modules resource plugin serves any existing file as `text/javascript`, wasm binaries slipped through. The fix keeps `canTransform` and additionally requires `response.headers.get("content-type")` to contain `text/javascript` before parsing.
2. [x] Fix an operator-precedence bug on the neighboring content-type check (`rollup.config.js:366`): `response?.headers?.get("content-type") || "".indexOf("text/javascript") >= 0` was truthy for any content-type; parenthesized to `(response?.headers?.get("content-type") || "").indexOf("text/javascript") >= 0` so it actually tests that the content-type contains `text/javascript`.
3. [x] Add regression tests under `packages/cli/test/cases/`:
   - `build.default.wasm-import-meta-url` — the **supported** `new URL("./add.wasm", import.meta.url)` pattern: asserts the build succeeds, emits exactly one content-hashed `.wasm` asset that is **byte-identical** to the source, and that the bundled script references the hashed filename.
   - `build.default.wasm-import-invalid` — an **unsupported** direct `import ... from "./add.wasm"`: asserts the build still fails (this syntax is not supported), but **no longer** crashes with the acorn `SyntaxError` from the `greenwood-import-meta-url` plugin. Post-fix the direct import surfaces a clear, Rollup-native `PARSE_ERROR` instead of a stack pointing at Greenwood internals.

### Resulting behavior for a direct `.wasm` import

The direct `import x from "./x.wasm"` syntax remains unsupported and the build fails, but the failure is now a Rollup-native `PARSE_ERROR` (Rollup parsing the served wasm bytes as a module) rather than the misleading acorn crash inside `greenwood-import-meta-url`. Users should use the supported `new URL("./x.wasm", import.meta.url)` + `fetch` / `WebAssembly.instantiateStreaming` pattern, which content-hashes the asset correctly.

### Breaking changes

None. The supported `new URL(...)` wasm path is unchanged (verified byte-for-byte), and legitimate JavaScript modules still transform normally (their served content-type is `text/javascript`).


packages/cli/test/cases/build.default.wasm-import-invalid/build.default.wasm-import-invalid.spec.js
```diff
@@ -0,0 +1,67 @@
+/*
```
thescientist13: nit: I wonder if it might make more sense to combine both of these test cases?  Seems like the are pretty related and we could just all the same related behavior under one suite.  Just thinking out loud.

packages/cli/src/config/rollup.config.js
```diff
@@ -363,7 +368,9 @@ function greenwoodImportMetaUrl(compilation) {
           if (plugin.shouldServe && (await plugin.shouldServe(url, request))) {
             const response = await plugin.serve(url, request);
 
-            if (response?.headers?.get("content-type") || "".indexOf("text/javascript") >= 0) {
+            // parenthesize so we test the content-type string contains "text/javascript"
+            // https://github.com/ProjectEvergreen/greenwood/issues/1723
+            if ((response?.headers?.get("content-type") || "").indexOf("text/javascript") >= 0) {
```
thescientist13, line 368: oh wow, good catch!

packages/cli/src/config/rollup.config.js
```diff
@@ -320,7 +320,12 @@ function greenwoodImportMetaUrl(compilation) {
         }
       }
 
-      if (!canTransform) {
+      // only parse resources that were actually served as JavaScript, otherwise binary resources
+      // (e.g. .wasm / .png / .woff2) get fed to acorn and crash the build with a cryptic SyntaxError
+      // https://github.com/ProjectEvergreen/greenwood/issues/1723
+      const contentType = response.headers.get("content-type") || "";
+
+      if (!canTransform || !contentType.includes("text/javascript")) {
```
thescientist13, line 325: I see yeah, this makes sense to me.

thescientist13 (approved):
Nice, just left one item for feedback, but not a big deal.  Will get this queued up for the next release candidate minor release 👍

packages/cli/test/cases/build.default.wasm-import-invalid/build.default.wasm-import-invalid.spec.js
jstockdi: Done in 4516e42e — merged both into a single `build.default.wasm-import` suite: the supported `new URL(..., import.meta.url)` scenario builds and asserts the emitted asset, then a second describe swaps `main.js` for the direct-import variant, asserts the build fails without the acorn/`greenwood-import-meta-url` error, and restores the fixture. One workspace, one wasm binary, 8 passing.

packages/cli/src/config/rollup.config.js
```diff
@@ -363,7 +368,9 @@ function greenwoodImportMetaUrl(compilation) {
           if (plugin.shouldServe && (await plugin.shouldServe(url, request))) {
             const response = await plugin.serve(url, request);
 
-            if (response?.headers?.get("content-type") || "".indexOf("text/javascript") >= 0) {
+            // parenthesize so we test the content-type string contains "text/javascript"
```
thescientist13: not sure we need a comment here, this was just a basic logic bug, pretty self-evident enough, IMO

jstockdi: Agreed — removed in `8bd768c3`. The parenthesization itself is untouched. I left the comment above the content-type gate in `greenwoodImportMetaUrl`, since why a non-JS resource must never reach acorn is less self-evident there, but happy to drop that one too if you'd rather.

packages/cli/test/cases/build.default.wasm-import/build.default.wasm-import.spec.js
```diff
@@ -0,0 +1,136 @@
...
+ * Run Greenwood build command with .wasm modules, covering both the supported
+ * `new URL("./add.wasm", import.meta.url)` pattern and an unsupported direct
+ * `import ... from "./add.wasm"`.
+ *
+ * User Result
+ * For the supported pattern, should generate a Greenwood build that emits the .wasm as a
+ * content-hashed, byte-identical asset, without crashing the JavaScript parser on the binary.
+ * For the unsupported direct import, the build should fail, but NOT with the misleading acorn
+ * `SyntaxError` thrown from inside the `greenwood-import-meta-url` Rollup plugin (which pointed
+ * users at internal plugin code when the wasm binary was fed to the JavaScript parser). A clear,
```
thescientist13: > A clear, Rollup-native parse error is acceptable.

I think we can probably leave the LLM editorializing out of the comments, let's just try and have the comment focus on the issue reproduction / outcome 👍 

That said, in the issue it was mentioned that this was a risk for any content type, since we weren't checking it for `application/javascript` specifically, so wondering if we should rename this test to be a bit more generic to better capture the scenario, since it's not only for WASM (pedantically)

Maybe something _build.default.binary-import-bundling_ or similar.  WDYT?

thescientist13 (commented):
Actually, could you provide a standalone reproduction repo for this?  I was trying to reproduce locally using the setup of the test case and was not able to.  I was trying in an existing project so maybe it was combination of plugins or settings in my greenwood config that may be influencing things?

I think it at least would help us better understand if there's a narrower set of cases when this happens, or it's as trivial / vanilla to repro as presented in the issue.

jstockdi:
Reproduced it standalone, and I think I know why it did not reproduce for you.

**The discriminator is the import form, not the project setup.**

A *direct* import crashes:
```js
import * as wasm from "./add.wasm";
```

The `new URL` pattern does **not**:
```js
const wasmUrl = new URL("./add.wasm", import.meta.url);
```

That second form is what `src/scripts/main.js` ships with in the test fixture — the direct-import variant only gets written in by the second `before` hook in the spec. So anyone setting up from the fixture by hand gets the `new URL` form, a clean build, and concludes it does not reproduce. That is almost certainly what happened.

Verified both ways against unfixed `37e68db3` on a default, config-less project — no `greenwood.config.js`, no plugins:

| `src/scripts/main.js` | unfixed master |
|---|---|
| `import * as wasm from "./add.wasm"` | crashes |
| `new URL("./add.wasm", import.meta.url)` | builds fine |

The crash:
```
SyntaxError: Unexpected character ' ' (1:0)
  plugin: 'greenwood-import-meta-url',
  hook: 'transform',
  id: '.../src/scripts/add.wasm'
```

With the fix, that same direct import still fails the build — a direct wasm import is genuinely unsupported — but as a Rollup-native `PARSE_ERROR` with no `greenwood-import-meta-url` attribution, which is the outcome the test asserts.

**Standalone repro:** https://github.com/Battle-Creek-LLC/greenwood/tree/repro/wasm-import-content-type — branched off `37e68db3` with one commit adding a `repro/` project (4 files) and a README covering the above. So it is as vanilla as the issue presented; it just needs the direct-import form.

**Renamed** to `build.default.binary-import-bundling` as you suggested, and trimmed the editorializing from the spec header and the explanatory comment from `rollup.config.js`. Pushed as `e15d96aa`, 8/8 passing. I kept wasm as the concrete binary rather than adding a second type — say the word if you would rather the case actually exercise a `.png` or `.woff2` alongside it, now that the name claims the general scenario.

**One thing worth a second look**, since it is in the diff and is not the content-type gate. The second hunk fixes an operator-precedence bug:

```js
// before — "".indexOf() is always -1, so this is (contentType || false)
if (response?.headers?.get("content-type") || "".indexOf("text/javascript") >= 0) {
// after
if ((response?.headers?.get("content-type") || "").indexOf("text/javascript") >= 0) {
```

The old form added a plugin's extensions to `bundleExtensions` whenever the plugin returned *any* content-type. It only bites projects with a custom resource plugin, which is why it is not part of the vanilla repro above. Happy to split it into its own issue and PR if you would rather keep this one to the content-type gate.

thescientist13 (commented):
Now I'm curious if this is even "allowed"?  I recreated your reproduction and couldn't even get it load in the browser in development mode, let alone when trying to bundle 🤔 
https://github.com/thescientist13/greenwood-import-wasm

Image

I based Greenwood's support / test cases on what I found on [MDN](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/Module/imports_static) 
- https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/JavaScript_interface/Module/imports_static
- https://developer.mozilla.org/en-US/docs/WebAssembly/Guides/Loading_and_running
- https://wasmbyexample.dev/examples/hello-world/hello-world.assemblyscript.en-us
- https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/develop.default
- https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/serve.default

Do you have an example of this working on the web as you've authored it?  Maybe like a CodePen or something?  I guess I'm not sure if this is an "expectations" issue, or a Greenwood issue atm.

jstockdi:
You're right, and I don't have an example — there isn't one to have. Evaluation-phase ESM integration (`import * as wasm from "./add.wasm"`) hasn't shipped in any browser. What has standardized is the source phase import, `import source mod from "./add.wasm"`, which hands you a `WebAssembly.Module` and is what Chrome and Deno have implemented. So your repro behaving that way is the expected outcome, not a Greenwood bug.

That splits into two things, and I think only the second is a Greenwood concern:

1. The import in my reproduction isn't supported on the web. That's an expectations issue, and I don't think it's worth supporting.
2. Separately from wasm, `greenwoodImportMetaUrl` called `acorn.parse()` on every intercepted resource without checking the content type, so any resource plugin serving non-JS bytes reaches the JS parser. A `.wasm` import is just the cheapest way to get there. With the gate in place the build still fails — it fails with the Rollup error instead of an acorn `SyntaxError` pointing at a binary.

That's why I renamed the case to `build.default.binary-import-bundling` in e15d96aa, following your earlier note — it's about the content-type gate rather than wasm specifically.

If you'd rather Greenwood say something explicit for an unsupported binary import instead of skipping the transform and letting Rollup raise it, happy to open a separate issue for that.

packages/cli/test/cases/build.default.binary-import-bundling/build.default.binary-import-bundling.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    // restore the workspace fixture no matter how the build run above ends
+    after(async function () {
+      await fs.writeFile(mainScriptUrl, originalMainScript);
+    });
+
+    it("should fail the build", function () {
+      expect(buildError).to.not.equal(undefined);
+    });
+
+    // https://github.com/ProjectEvergreen/greenwood/issues/1723
```
thescientist13, line 125: I think linking this issue above already covers us on having to do it again :)

packages/cli/test/cases/build.default.binary-import-bundling/build.default.binary-import-bundling.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    const mainScriptUrl = new URL("./src/scripts/main.js", import.meta.url);
+    let originalMainScript;
+    let buildError;
+
+    before(async function () {
+      originalMainScript = await fs.readFile(mainScriptUrl, "utf-8");
+      await fs.writeFile(
+        mainScriptUrl,
+        [
+          'import * as wasm from "./add.wasm";',
```
thescientist13, line 101: Yeah, I guess based on your last comment, I'm not sure if it's worth having a test case for this, as it seems that we confirmed the root cause was on rollup bundling + binary reference, which I think the `describe` block above already covers, where as previously it seems like that would have failed, right?

If so, then I that's the behavior worth covering IMO to validate the breakage reported.  As we seemed to came to the same conclusion on, the rest about the incorrect syntax just added more noise.  So leave that's noise out of the test cases as well.  👍

PR #1726, opened 2026-07-18
fix(cli): #1725 reconcile wildcard import keys against condition not filename match

### LINKED ISSUE #1725: import map generation for wildcards is not matching against the export condition
### What happened?

Observed in [this project](https://github.com/thescientist13/greenwood-foldkit/pull/1) that import generation was not matching correctly, favoring the matching wildcard of the matched file path when generating keys

For example, given this exports map and condition
https://app.unpkg.com/effect@4.0.0-beta.70/files/package.json#L51
```
"./*": "./dist/*.js",
```

The key includes dist
```
"effect/dist/Cause.js": "/~/Users/owenbuckley/Workspace/github/greenwood-foldkit/node_modules/effect/dist/Cause.js",
```

It should be this (no dist, no .js)
```
"effect/Cause": "/~/Users/owenbuckley/Workspace/github/greenwood-foldkit/node_modules/effect/dist/Cause.js",
```

### Steps to reproduce

1. Install **effect** as a dependency
1. Import it in a script

Image

### Environment

Latest Greenwood version

### Additional Context

This seems to make sense in looking at the code, the `subKey` is only derived from the filename path match
https://github.com/ProjectEvergreen/greenwood/blob/v0.34.0/packages/cli/src/lib/walker-package-ranger.js#L103

thescientist13:
## Related Issue

resolves #1725 

## Documentation 

N / A

## Summary of Changes

1. reconcile wildcard import keys against condition not filename match (fixed a lot of import maps entries!)

> Note: I also tested in all the key demo projects here - https://github.com/thescientist13/greenwood-lit-ssr/pulls

## TODO
1. [x] fix windows test cases
1. [ ] more wildcard exports support? - https://github.com/ProjectEvergreen/greenwood/pull/1726#issuecomment-5148610871


thescientist13:
is there actually more to do here, or is effect "right?  In it's package.json, it still doesn't have wildcards
https://app.unpkg.com/effect@4.0.0-beta.102/files/package.json#L29

Do we need to more (assumed) wildcard support / expansion somehow? 🤔

PR #1740, opened 2026-07-30
fix(cli): #1739 copy both favicon.ico and favicon.svg when both exist

### LINKED ISSUE #1739: favicon.svg silently dropped from build output when favicon.ico also exists
### What happened?

When a project workspace contains both `src/favicon.ico` and `src/favicon.svg`, the favicon copy plugin copies only the `.ico` to the output directory — the `.svg` never reaches `public/`. `greenwood develop` serves `/favicon.svg` fine (200 from the workspace), but after `greenwood build` the file is missing, so production 404s it while the page's `<link rel="icon" href="/favicon.svg">` still references it. Another works-in-dev / breaks-in-prod trap.

Shipping both favicons is the standard modern pattern (`.svg` for scalability/dark-mode with `.ico` as the legacy fallback), so silently dropping one of them is surprising.

### Steps to reproduce

1. Create a project with `src/pages/index.html` that links both icons (`<link rel="icon" href="/favicon.svg" type="image/svg+xml">` and `<link rel="icon" href="/favicon.ico" sizes="any">`), plus both `src/favicon.ico` and `src/favicon.svg`.
2. `greenwood develop` → both `/favicon.ico` and `/favicon.svg` return 200.
3. `greenwood build` → the build log shows only `favicon.ico` copied; `public/favicon.svg` does not exist.
4. `greenwood serve` → `curl -sI http://localhost:8080/favicon.svg` → **404**.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/plugins/copy/plugin-copy-favicon.js` — the plugin is an `if (ico exists) { copy ico } else if (svg exists) { copy svg }`, making the two mutually exclusive with `.ico` winning. Two independent `if` blocks would copy both when both exist.

jstockdi:
## Related Issue

Resolves #1739

## Documentation

N/A

## Summary of Changes

1. [x] copy `favicon.svg` independently of `favicon.ico` in `plugin-copy-favicon.js` — the previous `else if` made the two mutually exclusive (ico wins), so a workspace shipping both (the standard modern pattern: svg + ico legacy fallback) silently lost the svg from `public/`, producing a prod 404 that `develop` masks
2. [x] new regression case `build.default.favicon-both` — fixture with both favicons and a page linking both; asserts both files exist in the output directory and are byte-equal to their sources (fails on unpatched master with `ENOENT` for `favicon.svg`)

No breaking changes.


thescientist13 (commented):
left some comments in https://github.com/ProjectEvergreen/greenwood/issues/1739#issuecomment-5152446430 to better understand the use case here

thescientist13 (approved):

PR #1741, opened 2026-07-30
fix(cli): #1738 prefix base path on active content client requests

### LINKED ISSUE #1738: activeContent client fetches /graph.json without basePath prefix in production builds
### What happened?

On a production (non-prerendered) `activeContent` site with `basePath` configured, the shipped data client fetches a hard-coded `/graph.json` instead of `${basePath}/graph.json`, so every content query fails at runtime with a 404 once the site is hosted under its subpath.

Local `greenwood serve` masks the bug — the static server also serves output files at unprefixed URLs — so the site works locally and breaks only on a real subpath host (e.g. nginx routing only `/my-app/*`, GitHub Pages project sites). The sibling code paths are already basePath-aware: the `data-*.json` fetch prefixes `BASE_PATH`, and the develop-mode `___graph.json` fetch prefixes it too. Only the production `graph.json` branch is missing the prefix.

### Steps to reproduce

1. Create a project with `greenwood.config.js` setting `activeContent: true` and `basePath: "/my-app"`, and a component that calls `getContent()` (or any content API) at runtime.
2. `greenwood build`
3. Inspect the bundled data client in the output: the content fetch targets `/graph.json`, not `/my-app/graph.json` (while `data-*.json` and develop-mode `___graph.json` fetches are prefixed).
4. Host `public/` under the `/my-app/` subpath (any static server that only exposes the subpath). Load the site → the client fetch to `/graph.json` 404s and content never renders.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/data/client.js:20` — the `CONTENT_STATE && !PRERENDER` branch fetches `"/graph.json"` while line 16 (`data-*.json`) and line 32 (`___graph.json`) both prefix `BASE_PATH`. `graph.json` is emitted at the output root by `lifecycles/bundle.js`, so it is reachable at `${basePath}/graph.json` — the fix is to prefix this fetch like its siblings.

jstockdi:
## Related Issue

Resolves #1738

## Documentation

N/A

## Summary of Changes

1. [x] prefix `BASE_PATH` on the `graph.json` fetch in the production (`CONTENT_STATE && !PRERENDER`) branch of `data/client.js`, mirroring the sibling branches that already prefix it for `data-*.json` and develop-mode `___graph.json`
2. [x] new regression case `build.config.base-path-data-client` — an `activeContent: true` + `basePath: "/my-path"` fixture whose component imports the data client; asserts the emitted bundle resolves the base path at runtime (`__GWD_BASE_PATH__`) and fetches `` `${basePath}/graph.json` `` rather than the hard-coded `fetch("/graph.json")`

A note on the test approach: local `greenwood serve` masks this bug (its static server answers unprefixed URLs too — the 404 only manifests on a real subpath host like nginx or GitHub Pages routing only `/my-app/*`), so the spec asserts on the emitted bundle source instead of an end-to-end fetch. The two bug assertions fail on unpatched master and pass with the fix.

One adjacent observation for maintainers (not addressed here): optimized production HTML only injects `__CONTENT_AS_DATA_STATE__`, never `__CONTENT_OPTIONS__`, so production always takes the `!PRERENDER` branch even for prerendered sites — which widens the blast radius of this bug. Happy to file that separately if useful.

No breaking changes.


packages/cli/test/cases/build.config.base-path-active-content/build.config.base-path-active-content.spec.js
```diff
@@ -0,0 +1,119 @@
+/*
```
thescientist13, line 1: For these test cases, I would consider renaming this to _build.config.base-path-active-content_ (and the filename too) and we should make a development version too.


This should then co-locate nicely along with our other two base path test cases
- https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/develop.config.base-path
- https://github.com/ProjectEvergreen/greenwood/tree/master/packages/cli/test/cases/serve.config.base-path

jstockdi: Done in 66e40f6c:

- Renamed to `build.config.base-path-active-content` (directory + spec filename) so it co-locates with the other base path cases.
- Added `develop.config.base-path-active-content` with the same fixture.

Writing the develop version surfaced a real gap: in develop mode the data client requests `${BASE_PATH}/___graph.json`, but the active content plugin's `shouldServe` only matched a bare `/___graph.json` — so with a base path configured, active content silently returned an empty response during `greenwood develop`. Included the one-line fix here (`shouldServe` now matches `${basePath}/___graph.json`), and the new develop spec pins it: its `___graph.json` cases fail without the fix and pass with it. The default no-base-path behavior is unchanged — `develop.config.active-content` still passes.

thescientist13:
> One adjacent observation for maintainers (not addressed here): optimized production HTML only injects __CONTENT_AS_DATA_STATE__, never __CONTENT_OPTIONS__, so production always takes the !PRERENDER branch even for prerendered sites — which widens the blast radius of this bug. Happy to file that separately if useful.

Yeah, might be good to open an issue for this

packages/cli/test/cases/build.config.base-path-active-content/build.config.base-path-active-content.spec.js
thescientist13: nice, good catch!

thescientist13 (approved):

jstockdi:
Filed as #1757 👍

packages/cli/src/plugins/resource/plugin-active-content.js
```diff
@@ -23,10 +23,12 @@ class ContentAsDataResource {
   }
 
   async shouldServe(url) {
-    const { activeContent } = this.compilation.config;
+    // the develop mode data client requests this prefixed with the base path
```
thescientist13: same, I don't think we really need comments here, just an oversight bug

packages/cli/src/data/client.js
```diff
@@ -17,7 +17,8 @@ async function getContentAsData(key = "") {
     ).then((resp) => resp.json());
   } else if (CONTENT_STATE && !PRERENDER) {
     // if user is not prerendering, just fetch the entire graph but apply the same filtering
-    const graph = await fetch("/graph.json").then((resp) => resp.json());
+    // https://github.com/ProjectEvergreen/greenwood/issues/1738
```
thescientist13: don't think we need a comment here, just an oversight bug.  we have it pretty well covered in the test case files

jstockdi:
Both comments dropped and the branch rebased onto `37e68db3` — force-pushed as `1811e595`, conflict resolved.

Worth flagging how the conflict resolved, since it touched more than whitespace. This branch was cut before #1715 landed, so `client.js` still had the old `key.split("-").pop()` split. Master's first-delimiter version is the correct one, so I kept master's logic verbatim and took only the `${BASE_PATH}` prefix from this branch:

```js
const graph = await fetch(`${BASE_PATH}/graph.json`).then((resp) => resp.json());
// split on only the first "-" so hyphenated collection names / routes keep their hyphens
// https://github.com/ProjectEvergreen/greenwood/issues/1715
const delimiterIndex = key.indexOf("-");
const value = delimiterIndex === -1 ? "" : key.slice(delimiterIndex + 1);
```

So this PR is now purely the two `basePath` prefixes and nothing else. Verified both sides still hold: `build.config.base-path-active-content`, `develop.config.base-path-active-content` and `build.config.base-path-data-client` pass 37/37, and your #1715 cases `build.config.prerender-collections-hyphen` and `develop.config.active-content-collection-hyphen` pass 33/33 against the resolution.

thescientist13:
Thanks, will get this into the next release!

Issue #1614, opened 2025-11-11
update to minimum Node version of >= 24.12.0


thescientist13:
### Motivation

As Node 24 moves into Active LTS, Node 22 will go out of maintenance by this time next years
https://nodejs.org/en/about/previous-releases
<img width="963" height="573" alt="Image" src="https://github.com/user-attachments/assets/6085e056-efe0-4531-a125-1c31fc73b01c" />

However, in particular for #1563 , we adopted a polyfill for now to support the transition from Node 22 -> 24 since `URLPattern` had not been backported at this time, and [serverless hosting providers](https://github.com/ProjectEvergreen/greenwood/discussions/1533#discussioncomment-14908430) are still working on support for Node24, so this didn't feel like the right time to force the cutover.

### Technical Design
1. Should bump to [>= 24.12.0](https://nodejs.org/en/blog/release/v24.12.0) since that is the first LTS with stable TypeScript type-stripping support
1. Bump all engines (CLI, init)
1. Remove the `URLPattern` polyfill
1. Update GitHub Actions (shift to 24, 26)
1. Update _.nvmrc_
1. Update any relevant website docs
1. Update Vercel adapter for Node versions
    - set new default to match - https://vercel.com/docs/functions/runtimes/node-js/node-js-versions
1. Update Adapter plugins to use `URLPattern` for extracting props per https://github.com/ProjectEvergreen/greenwood/pull/1563
1. SST is waiting on Node 24 support - released! - https://github.com/anomalyco/sst/releases/tag/4.0.2
    - https://sst.dev/docs/migrate-from-v3/
    - https://github.com/anomalyco/sst/pull/6243
    - https://github.com/anomalyco/sst/pull/5944
1. remove resolution - https://github.com/ProjectEvergreen/greenwood/pull/1686/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R37

### Additional Context

We should at least make sure to do this prior to the 1.0 release since this would be a breaking change.

Also, for Netlify (at least right now) an environment variable is required for Node 24 support - https://github.com/ProjectEvergreen/greenwood/discussions/1533#discussioncomment-15249252

PR #1742, opened 2026-07-30
fix(cli): #1737 return 404 for dynamic route values not returned by get static paths

### LINKED ISSUE #1737: greenwood serve returns 500 instead of 404 for dynamic route values not returned by getStaticPaths
### What happened?

Under `greenwood serve`, requesting a dynamic-route value that `getStaticPaths` never returned yields **HTTP 500** with an `ENOENT` stack trace dumped to the server console, instead of a clean **404**. The response body text even says "Not Found" while the status is 500.

Worse, a page whose `getStaticPaths` returns `[]` turns *every* single-segment URL under that route into a 500, because the route matching only checks that `page.staticPaths` is truthy (an empty array is truthy) and that the URLPattern (`/:slug/`) matches — the actual list of enumerated paths is never consulted. The server then blindly `fs.readFile`s `public/<value>/index.html`, and the `ENOENT` is caught by the top-level handler which logs the raw error and sets `ctx.status = 500`.

For a statically-exported site this is a correctness and operations problem: bots and stale links generate 5xxs (alarming monitoring and hurting SEO) for what is semantically a plain "this page does not exist".

### Steps to reproduce

1. Create a project with a dynamic route `src/pages/[slug].js` whose `getStaticPaths` returns a finite list, e.g. `[{ params: { slug: "alpha" } }]`, and a `getBody` rendering the param.
2. `greenwood build && greenwood serve`
3. `curl -i http://localhost:8080/alpha/` → 200 as expected.
4. `curl -i http://localhost:8080/unknown/` → **500**, body "Not Found", and an `ENOENT ... public/unknown/index.html` stack in the server console. Expected: **404**.
5. Change `getStaticPaths` to return `[]`, rebuild, and observe that *any* URL like `/anything/` now 500s.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause: `packages/cli/src/lifecycles/serve.js` (~lines 318–345). The matching uses `page.staticPaths && getParamsFromSegment(...)`, i.e. any single-segment path matches the URLPattern; `staticPaths` content is never checked, and the unguarded `fs.readFile` of the derived output path throws `ENOENT` into the catch-all 500 handler. Next.js semantics for this case (`fallback: false`) are a 404 for values not enumerated by `getStaticPaths`.

jstockdi:
## Related Issue

Resolves #1737

## Documentation

N/A

## Summary of Changes

1. [x] in the static server's route matching (`serve.js`), a dynamic route now only matches when the requested pathname is one of the concrete routes enumerated from `getStaticPaths` (Next.js `fallback: false` semantics) — unknown values produce no matching route, so the middleware never reaches the `fs.readFile` and the response is a clean 404 instead of a 500 with an `ENOENT` stack in the server console
2. [x] the same check fixes `getStaticPaths` returning `[]` (previously *every* single-segment URL under the route 500'd, since the empty array is truthy and the URLPattern matches any value)
3. [x] new regression case `serve.default.ssr-static-paths-unknown` — one fixture with a finite `getStaticPaths` and one with an empty one; asserts `/alpha/` → 200, `/nope/` → 404, `/empty/anything/` → 404 (the two 404 assertions fail on unpatched master with 500)

Implementation notes:

- The pathname comparison checks both the raw and `encodeURI`-encoded forms of each enumerated route, since `URL` percent-encodes `url.pathname`. `encodeURI` was chosen deliberately over decoding the request path — `decodeURIComponent` throws on malformed percent-sequences, which would have re-introduced a 500 via the surrounding catch. This composes with, but does not depend on, #1714.
- The hybrid server needs no change: its `isDynamicRoute` check already excludes `staticPaths` routes, so unknown values inherit this 404.

Behavior change to be aware of (arguably the point of the fix): requests for non-enumerated dynamic values now return 404 where they previously returned 500. There is currently no fallback-rendering option for `serve`, so nothing functional is lost — but calling it out in case a fallback mode is on the roadmap, in which case this guard is where it would slot in.

No breaking changes.

### DIFF
--- a/packages/cli/src/lib/url-utils.js
+++ b/packages/cli/src/lib/url-utils.js
@@ -38,10 +38,8 @@ function getMatchingDynamicSsrRoute(compilation, route) {
       return false;
     }
 
-    // for getStaticPaths routes, only match values enumerated at build time (Next.js
-    // `fallback: false` semantics) so unknown values 404 in develop and serve alike,
-    // instead of live rendering in develop and throwing an ENOENT / 500 in serve
-    // https://github.com/ProjectEvergreen/greenwood/issues/1737
+    // for getStaticPaths routes, only match values enumerated at build time so unknown
+    // values 404 in develop and serve alike
     return (
       !node.staticPaths ||
       node.staticPaths.some((staticPath) => {
--- a/packages/cli/src/lib/url-utils.js
+++ b/packages/cli/src/lib/url-utils.js
@@ -27,12 +27,29 @@ function getMatchingDynamicSsrRoute(compilation, route) {
   const { graph, config } = compilation;
 
   return graph.find((node) => {
-    return (
+    const matchesSegment =
       route !== "/404/" &&
       node.segment &&
       new URLPattern({ pathname: `${config.basePath}${node.segment.pathname}` }).test(
         `https://example.com${route}`,
-      )
+      );
+
+    if (!matchesSegment) {
+      return false;
+    }
+
+    // for getStaticPaths routes, only match values enumerated at build time (Next.js
+    // `fallback: false` semantics) so unknown values 404 in develop and serve alike,
+    // instead of live rendering in develop and throwing an ENOENT / 500 in serve
+    // https://github.com/ProjectEvergreen/greenwood/issues/1737
+    return (
+      !node.staticPaths ||
+      node.staticPaths.some((staticPath) => {
+        const staticRoute = getStaticRouteFromDynamicRoute(staticPath, node.segment, node.route);
+
+        // compare the raw and encoded forms since URL percent-encodes the pathname
+        return staticRoute === route || encodeURI(staticRoute) === route;
+      })
     );
   });
 }
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -13,6 +13,7 @@ import {
   getMatchingDynamicApiRoute,
   getMatchingDynamicSsrRoute,
   getParamsFromSegment,
+  getStaticRouteFromDynamicRoute,
 } from "../lib/url-utils.js";
 import { Readable } from "node:stream";
 import { Worker } from "node:worker_threads";
@@ -320,7 +321,21 @@ async function getStaticServer(compilation, composable) {
       const url = new URL(`http://localhost:${port}${ctx.url}`);
       const matchingRoute = compilation.graph.find(
         (page) =>
-          (page.staticPaths && getParamsFromSegment(compilation, page.segment, url.pathname)) ||
+          // for dynamic routes, only match values enumerated from getStaticPaths, otherwise
+          // unknown values should fall through to a 404 instead of an ENOENT / 500
+          // https://github.com/ProjectEvergreen/greenwood/issues/1737
+          (page.staticPaths &&
+            getParamsFromSegment(compilation, page.segment, url.pathname) &&
+            page.staticPaths.some((staticPath) => {
+              const staticRoute = getStaticRouteFromDynamicRoute(
+                staticPath,
+                page.segment,
+                page.route,
+              );
+
+              // compare the raw and encoded forms since URL percent-encodes the pathname
+              return staticRoute === url.pathname || encodeURI(staticRoute) === url.pathname;
+            })) ||
           page.route === url.pathname,
       );
       const isSPA = compilation.graph.find((page) => page.isSPA);
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -318,9 +318,6 @@ async function getStaticServer(compilation, composable) {
   app.use(async (ctx, next) => {
     try {
       const url = new URL(`http://localhost:${port}${ctx.url}`);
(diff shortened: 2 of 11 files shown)


thescientist13 (commented):
Good catch!

Do we know what the behavior is in the following places, and could we add test cases for them if possible?
1. `greenwood develop`
1. Adatper plugins

> Note: if there's something needed in adapter plugins, do note that there is some small refactoring happening there per #1614 to be able to use `URLPattern`

jstockdi:
**TL;DR: neither surface shares the 500 — and develop now 404s unknown values too, so dev and prod agree. Adapters need no change.**

- **`greenwood develop`:** didn't 500 (renders on demand, no prebuilt-file read) but live rendered *any* dynamic value with a 200 — which would have diverged from serve's new 404. Both develop and serve match through `getMatchingDynamicSsrRoute`, so I moved the enumerated-values guard into that helper: develop now 404s non-enumerated values (`fallback: false` semantics, same as Next.js dev), and the static server's inline check collapsed into the shared helper. Flag this commit if you'd rather keep dev permissive — it's isolated and easy to drop.
- **New test case:** `develop.default.ssr-static-paths-unknown` — 200 for enumerated, 404 for unknown, 404 when `getStaticPaths` returns `[]`.
- **Adapter plugins:** nothing needed. All three build functions from `getDynamicPages()`, which filters on `page.isSSR && !page.staticPaths` — so `getStaticPaths` pages get no function and no route/redirect rule; they're prerendered static and unknown values fall through to the platform's native 404. Confirmed with a real Vercel adapter build of this fixture: `static/alpha/index.html`, no `functions/` directory, `config.json` of `{"version":3}`. Local `serve` was the odd one out; this PR aligns it with deployments.
- **Adapter tests:** the existing `build.dynamic-routing` cases (Vercel/Netlify/AWS) already had an `artists/[name].js` fixture but only implicit coverage via function counts — added explicit assertions: no function for the route, no route/redirect rule for the segment, static HTML per enumerated path.
- **#1614:** no interaction — `getStaticPaths` pages never enter the adapter routing code being refactored.

thescientist13 (approved):

jstockdi:
Pushed `e5ebe8b6` on an already-approved PR, so a note on why.

The comment I had left above the `getMatchingDynamicSsrRoute` call in `serve.js` was inaccurate. It said the function "only matches values enumerated from getStaticPaths" without the qualifier that this applies only to pages which actually define `getStaticPaths` — the return is `!node.staticPaths || node.staticPaths.some(...)`, so a dynamic SSR page without it still matches any value and renders per request. I misread my own comment while working on something unrelated today and drew the wrong conclusion from it, which seemed like reason enough to remove it.

Deleted rather than corrected: the accurate explanation already sits in `lib/url-utils.js` where the logic is, and it does carry the qualifier. `url-utils.js` is untouched.

Comment-only, 3 deletions, no code change. `serve.default.ssr-static-paths-unknown` (4), `develop.default.ssr-static-paths-unknown` (4) and `serve.dynamic-routing` (3) all pass; lint and format clean. GitHub kept the approval, so nothing needed from you unless you would rather I left it alone.

packages/cli/src/lifecycles/serve.js
```diff
@@ -315,11 +315,9 @@ async function getStaticServer(compilation, composable) {
...
     try {
       const url = new URL(`http://localhost:${port}${ctx.url}`);
-      const matchingRoute = compilation.graph.find(
-        (page) =>
-          (page.staticPaths && getParamsFromSegment(compilation, page.segment, url.pathname)) ||
-          page.route === url.pathname,
-      );
+      const matchingRoute =
+        compilation.graph.find((page) => page.route === url.pathname) ||
+        getMatchingDynamicSsrRoute(compilation, url.pathname);
```
thescientist13, line 323: good refactor on using `getMatchingDynamicSsrRoute`, I think I meant to do that as part of the initial implementation but must have forgot 😅

packages/cli/src/lib/url-utils.js
```diff
@@ -27,12 +27,29 @@ function getMatchingDynamicSsrRoute(compilation, route) {
...
       new URLPattern({ pathname: `${config.basePath}${node.segment.pathname}` }).test(
         `https://example.com${route}`,
-      )
+      );
+
+    if (!matchesSegment) {
+      return false;
+    }
+
+    // for getStaticPaths routes, only match values enumerated at build time (Next.js
```
thescientist13: thanks for catching this for the develop too 🙏 

this seems like a simple enough bug, so I think just having a comment saying this is sufficient
> for getStaticPaths routes, only match values enumerated at build time so unknown values 404 in develop and serve alike

we don't need to document the entire thing here since we have the test case which catches a lot of this info 👍

jstockdi:
Trimmed to your wording in `77f51b94` — the comment above the `getStaticPaths` check in `url-utils.js` is now:

```js
// for getStaticPaths routes, only match values enumerated at build time so unknown
// values 404 in develop and serve alike
```

Dropped the `fallback: false` aside and the issue link, since the test case carries that context. All 8 specs across `serve.default.ssr-static-paths-unknown` and `develop.default.ssr-static-paths-unknown` still pass.

thescientist13:
Thanks for applying the clean up items 👍

PR #1744, opened 2026-08-01
fix(cli): #1743 avoid interpolating active frontmatter values that contain replacement patterns

### LINKED ISSUE #1743: Active frontmatter values containing $ replacement patterns corrupt interpolated page output
### What happened?

With `activeContent: true`, a frontmatter value containing a `$` replacement
pattern (`$&`, `` $` ``, `$'`, `$$`, `$1`…) is corrupted when interpolated into
the page, because it is passed as the *replacement string* to
`String.prototype.replace`, where those sequences have special meaning.

- `$&` re-inserts the search needle, so `pay $& now` renders as
  `pay ${globalThis.page.data.money} now`.
- `` $` `` re-inserts the **entire preceding document**, silently duplicating the
  page inside itself — head, scripts and all — producing bizarre, broken HTML.

Perfectly ordinary values trigger it (prices like `save $1 now`, shell or regex
snippets in a blog post). The corruption is silent — the build succeeds. It fires
in both `develop` and `build`.

A secondary defect in the same loops: the *search* pattern is string-concatenated
from unescaped data key / collection names, so a key containing a regex
metacharacter builds a wrong `RegExp` (or throws).

### Steps to reproduce

1. Create a project with `greenwood.config.js` containing `activeContent: true`.
2. Create `src/pages/index.html`:
   ```
   ---
   title: costs $& per month
   money: "pay $& now $` later and $1 tomorrow"
   ---

   <html>
     <body>
       <h1>${globalThis.page.title}</h1>
       <p>${globalThis.page.data.money}</p>
     </body>
   </html>
   ```
3. Run `greenwood build`.
4. Inspect `public/index.html` — the `<p>` contains the search needle and a full
   copy of the preceding document instead of the literal frontmatter value, and
   the `<h1>` / `<title>` contain a mangled, self-repeating title.

Expected: the values render verbatim — `pay $& now $` later and $1 tomorrow` and
`costs $& per month`.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause — raw values passed as `replace` replacement strings:

- `packages/cli/src/plugins/resource/plugin-active-content.js:111` — active
  frontmatter keys loop (`${globalThis.page.*}`)
- `packages/cli/src/plugins/resource/plugin-active-content.js:122` — custom
  frontmatter data loop (`${globalThis.page.data.*}`)
- `packages/cli/src/plugins/resource/plugin-active-content.js:130-131` —
  collections loop (`${globalThis.collection.*}`)
- `packages/cli/src/lib/layout-utils.js:113` — the same defect for the page
  `<title>` during layout merging (this is why the head title corrupts too)

The unescaped-key search patterns are built at
`plugin-active-content.js:107/116/127`. MDN documents the `$` special patterns
and recommends a function replacement when the substitution must be literal.

jstockdi:
## Related Issue

Resolves #1743

## Documentation

N/A — no user-facing documentation changes (bug fix only).

## Summary of Changes

1. [x] Used the function form of `String.prototype.replace` in all three active
   content interpolation loops in
   `packages/cli/src/plugins/resource/plugin-active-content.js` (active
   frontmatter keys, custom `data.*` frontmatter, collections), so a frontmatter
   value containing `$&`, `` $` ``, `$'`, `$$` or `$n` is inserted verbatim
   instead of being expanded as a replacement pattern (which could silently
   duplicate the whole page inside itself).
1. [x] Applied the same function-form fix to the page `<title>` interpolation in
   `packages/cli/src/lib/layout-utils.js`, which had the identical defect on the
   layout-merge path.
1. [x] Added a small `escapeRegExpCharacters` helper (local, since
   `RegExp.escape` requires Node >= 24 and CI also runs Node 22) and used it on
   the interpolated key / collection names when building the search pattern, so
   a data key or collection name containing a regex metacharacter no longer
   builds a wrong `RegExp`.
1. [x] Added a regression test case
   `packages/cli/test/cases/build.config.active-frontmatter-dollar/` with a
   fixture page whose frontmatter `title` and custom `money` values contain
   `$&`, `` $` `` and `$1`, asserting they render verbatim in the built HTML and
   that the page is not duplicated inside itself. Verified the spec fails on the
   unpatched source and passes with the fix.
1. [x] No breaking changes — values without `$` patterns and keys without regex
   metacharacters interpolate exactly as before (the existing
   `build.config.active-frontmatter` case still passes).

### DIFF
--- a/packages/cli/test/cases/build.config.active-frontmatter-dollar/build.config.active-frontmatter-dollar.spec.js
+++ b/packages/cli/test/cases/build.config.active-frontmatter-dollar/build.config.active-frontmatter-dollar.spec.js
@@ -1,0 +1,96 @@
+/*
+ * Use Case
+ * Run Greenwood with activeContent configuration enabled for validating active frontmatter
+ * values that contain regex replacement patterns ($&, $`, $1).
+ *
+ * User Result
+ * Should generate a bare bones Greenwood build with frontmatter values interpolated verbatim,
+ * with no replacement pattern expansion or page duplication.
+ *
+ * User Command
+ * greenwood build
+ *
+ * User Config
+ * {
+ *   activeContent: true
+ * }
+ *
+ * User Workspace
+ * Greenwood default
+ *  src/
+ *   pages/
+ *     index.html
+ */
+// https://github.com/ProjectEvergreen/greenwood/issues/1743
+import { JSDOM } from "jsdom";
+import path from "node:path";
+import fs from "node:fs/promises";
+import { expect } from "chai";
+import { getOutputTeardownFiles } from "../../../../../test/utils.js";
+import { runSmokeTest } from "../../../../../test/smoke-test.js";
+import { Runner } from "gallinago";
+import { fileURLToPath } from "node:url";
+
+describe("Build Greenwood With: ", function () {
+  const LABEL = "Active Frontmatter with Dollar Sign Replacement Patterns";
+  const cliPath = path.join(process.cwd(), "packages/cli/src/bin.js");
+  const outputPath = fileURLToPath(new URL(".", import.meta.url));
+  let runner;
+
+  before(function () {
+    this.context = {
+      publicDir: path.join(outputPath, "public"),
+    };
+    runner = new Runner();
+  });
+
+  describe(LABEL, function () {
+    before(async function () {
+      await runner.setup(outputPath);
+      await runner.runCommand(cliPath, "build");
+    });
+
+    runSmokeTest(["public", "index"], LABEL);
+
+    describe("Frontmatter values containing replacement patterns for the home page", function () {
+      let dom;
+      let html;
+
+      before(async function () {
+        const htmlPath = path.resolve(this.context.publicDir, "./index.html");
+
+        dom = await JSDOM.fromFile(htmlPath);
+        html = await fs.readFile(htmlPath, "utf-8");
+      });
+
+      it("should interpolate a custom data frontmatter value with $&, $` and $1 verbatim", function () {
+        const money = dom.window.document.querySelector("body p.money").textContent;
+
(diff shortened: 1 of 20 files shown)


packages/cli/test/cases/build.config.active-frontmatter-dollar/build.config.active-frontmatter-dollar.spec.js
```diff
@@ -0,0 +1,96 @@
+/*
```
thescientist13, line 1: A couple oppourtunities for test maintenance here
- We have a develop version of active-content specs, maybe good to add this test case there?
- We don't have a serve version of this test case, maybe we could adapt the develop suite for a production build, including this test cast there?

WDYT?

jstockdi: Agreed on both — done in 539e16aa.

- **develop.config.active-content** — added a `pricing.html` page carrying the replacement-pattern frontmatter and a describe block asserting the dev server output interpolates it verbatim (graph count assertion bumped 6 → 7 for the new page).
- **serve.config.active-content** (new) — the develop suite adapted for a production build with `prerender: true`: same workspace plus `pricing.html`, running `build` + `serve` and asserting the prerendered nav / blog-post links from the content as data APIs, the `content-as-data-state` / `data-client-options` setup scripts, the cached `data-*.json` files, and the replacement-pattern page.

Both new blocks fail on `master` without this fix — the serve title path even compounds through the layout title merge, rendering `costs costs costs ${globalThis.page.title} per month` — and pass with it. I kept this build case as-is since it covers the third mode (static build without prerender).

thescientist13 (approved):

PR #1747, opened 2026-08-02
feat(cli): #1232 dynamic sitemap generation from a sitemap.xml.(js|ts) handler

### LINKED ISSUE #1232: Sitemap Generation

jstockdi:
Fresh implementation of dynamic sitemap generation, superseding #1260 and picking up the direction from the review there: the sitemap is a **route that returns a `Response`**, not a copied asset.

## Related Issue

Resolves #1232

## Summary of Changes

The user authors `src/sitemap.xml.js` (or `.ts`) exporting an API-route-style handler:

```js
export async function handler(request, { compilation }) {
  const urls = compilation.graph
    .map((page) => `  <url><loc>https://www.example.com${page.route}</loc></url>`)
    .join("\n");

  return new Response(`<urlset>...</urlset>`, {
    headers: { "Content-Type": "text/xml" },
  });
}
```

Same `handler(request, context)` / `Response` contract as API routes (SvelteKit-style, per the review feedback on #1260), with `compilation` provided in the context so the page graph is available without any user-side imports.

1. [x] `lifecycles/graph.js` — tracks the module as a dedicated `compilation.manifest.sitemap` entry (`pageHref` + `route`). A static `src/sitemap.xml` takes precedence with a console warning, so existing projects are unaffected.
2. [x] `plugins/resource/plugin-sitemap.js` + `lib/sitemap-route-worker.js` — `greenwood develop` serves `/sitemap.xml` by executing the handler in a fresh worker per request (mirroring `api-route-worker.js`, so edits are picked up live), passing the JSON-serialized compilation the same way SSR page workers receive it.
3. [x] `config/rollup.config.js` + `lifecycles/bundle.js` — `greenwood build` bundles the module like an API route (`sitemap.xml.route.js`, node export conditions, chunks + `outputHref` synced into the manifest), then executes the **bundled** module once and writes the result to `public/sitemap.xml` as a static snapshot. Because the bundle is what gets executed, user `import`s (a CMS SDK, a database client) are proven to survive bundling — and the bundled handler is sitting in the output with its manifest entry for adapter plugins to mount on demand as a follow-up.
4. [x] TypeScript support falls out of the standard resource pipeline (same path `.ts` API routes take through the Rollup resource loader) — covered by a dedicated test case.
5. [x] Fails the build loudly if the module doesn't export a `handler` function or the handler returns a non-`Response`.

Test cases: `build.default.sitemap-dynamic`, `build.default.sitemap-dynamic-ts`, `build.default.sitemap-static-and-dynamic` (static-wins precedence), `develop.default.sitemap-dynamic`, `develop.default.sitemap-static-and-dynamic` (static precedence served with `text/xml` in develop).

## Documentation

Happy to PR a docs section to www.greenwoodjs.dev for `sitemap.xml.(js|ts)` / the `handler` contract alongside this once the shape is signed off.

## TODO / Open Questions

- Kept the `sitemap.xml.(js|ts)` filename (the `.xml` makes precedence vs. a literal `sitemap.xml` self-evident) — happy to switch to `sitemap.(js|ts)` Next.js-style if you'd prefer.
- Adapter plugins don't consume `manifest.sitemap` yet; they keep working today via the static snapshot. Wiring the bundled handler into the adapters could be a follow-up issue.

## Known caveats (shared with existing API route / SSR behavior)

- `greenwood serve` doesn't route `/sitemap.xml` through the handler — it relies on the built `public/sitemap.xml` snapshot. Static-file serving of `.xml` under `greenwood serve` is a pre-existing gap (same for a copied static `sitemap.xml` or `robots.txt`); happy to take that as a follow-up.
- In develop, `.ts` modules are imported with Node's native type stripping (same as `.ts` API routes), while build goes through the resource pipeline — so `useTsc`-only syntax works at build but not in develop. Inherited from the API route path.
- The build-time snapshot executes the handler with a synthetic `Request` (`localhost` origin), so handlers should not derive their site origin from `request.url` — worth a line in the docs when they're written.


thescientist13:
I think there's overlap around technical design as being discussed in https://github.com/ProjectEvergreen/greenwood/issues/1752#issuecomment-5199650075 so let's try and align there if we can

PR #1749, opened 2026-08-02
fix(cli): #1748 reject prod server requests that resolve outside the output directory

### LINKED ISSUE #1748: Path traversal / arbitrary file read in `greenwood serve` static file handler
### Summary

The production server (`greenwood serve`) builds a filesystem path directly from the request URL with no containment check, so a request path containing `../` escapes the output directory and reads arbitrary files off the host. For files whose extension is one a standard static-resource plugin serves (`.js`, `.css`, `.json`, `.map`, `.svg`, `.png`/`.jpg`/`.gif`/`.webp`/`.avif`/`.ico`, `.woff`/`.woff2`/`.ttf`/`.eot`, `.wasm`, `.mp3`, `.mp4`), the file contents are returned in the response body. For any other extension the file is still `fs.access`-ed, which leaks its existence.

### Affected code

`packages/cli/src/lifecycles/serve.js`, the static-asset middleware in `getStaticServer` (v0.34.0):

```js
const url = new URL(`.${ctx.url.replace(basePath, "")}`, outputDir.href);

if (await checkResourceExists(url)) {
  // ...serve the bytes via the matching standard resource plugin
}
```

`ctx.url` is the raw request target. Node's HTTP server and Koa both pass `../` through untouched (neither normalizes `req.url`/`ctx.url`). Because `outputDir.href` is a `file://` URL, it has no origin for the URL parser to clamp against, so `new URL('./../../../../etc/passwd', 'file:///app/public/')` resolves to `file:///etc/passwd` — outside the web root.

The two sibling middlewares in the same file that resolve pages (they build `new URL('http://localhost:' + port + ctx.url)`) are **not** affected: an `http://` base clamps `../` to the origin, so the escape is specific to the `file://`-based static handler.

### Reproduction

```sh
# any Greenwood project
greenwood build && greenwood serve

# browsers/fetch normalize `..`, so use a client that sends the raw path:
curl --path-as-is 'http://localhost:8080/../../../../../../etc/hosts'
# -> 200, body is the contents of /etc/hosts

# a source file above the web root, contents returned because `.js` is a served type:
curl --path-as-is 'http://localhost:8080/../greenwood.config.js'
```

Verified against `@greenwood/cli@0.34.0`: a raw request `GET /../<abs-path>/secret.txt` resolves to that absolute path and returns its bytes; the resolved path is confirmed outside `outputDir`.

### Impact

- Unauthenticated remote arbitrary file read for any file with a served extension (source, config, `.env.js`-style secrets, private JSON, etc.).
- Existence oracle for every other path.
- Reachable by any client that does not pre-normalize the path (`curl --path-as-is`, a raw socket, most HTTP libraries with an explicit path). Browsers and `fetch()` normalize `..`, so it is not click-through exploitable but is trivially scriptable.

Severity: high (network-reachable, unauthenticated, confidentiality).

### Suggested fix

Reject any resolved URL that falls outside `outputDir` before touching the filesystem. Because `outputDir.href` ends in `/`, a `startsWith` check is boundary-safe (it rejects both `../` escapes and sibling-prefix paths like `.../public-evil/`), and the URL parser has already collapsed encoded `..` (`%2e%2e`) by the time the check runs. PR to follow.

jstockdi:
### Related Issue

Closes #1748

### Summary of Changes

The static-asset middleware in `getStaticServer` (`packages/cli/src/lifecycles/serve.js`) built a `file://` URL directly from the request path and passed it to `checkResourceExists`/the resource plugins without checking that it stayed inside the output directory. Because the base is a `file://` URL — which has no origin for the URL parser to clamp against — a request path containing `../` resolved outside the web root, so `greenwood serve` returned the contents of arbitrary files with a served extension (`.js`, `.css`, `.json`, `.map`, images, fonts, `.wasm`, media) and an existence oracle for everything else. Node and Koa pass `../` through untouched, so any client that does not pre-normalize the path (e.g. `curl --path-as-is`, a raw socket) could reach it.

This adds a containment check: the resolved URL must fall within `outputDir` before the filesystem is touched, otherwise the request falls through to the existing 404 path. Because `outputDir.href` ends in `/`, the `startsWith` boundary also rejects sibling-prefix paths (e.g. `.../public-evil/`), and the URL parser has already collapsed percent-encoded `..` by the time the check runs.

The two sibling page-resolution middlewares in the same file are unaffected — they build `http://` URLs, which clamp `../` to the origin, so they were never able to escape.

```diff
  const url = new URL(`.${ctx.url.replace(basePath, "")}`, outputDir.href);
+ // reject paths that resolve outside the output directory (e.g. `../` traversal);
+ // outputDir.href ends in `/`, so this boundary also excludes sibling-prefix paths
+ const isWithinOutputDir = url.href.startsWith(outputDir.href);

- if (await checkResourceExists(url)) {
+ if (isWithinOutputDir && (await checkResourceExists(url))) {
```

### Testing

New case `serve.default.path-traversal` with a canary file placed one level above the output directory:

- a legitimate in-root asset (`/assets/data.json`) still returns `200` with its contents
- a raw `../` traversal (`/../secret-outside-webroot.json`) returns `404` and does not leak the file
- a percent-encoded traversal (`/%2e%2e/secret-outside-webroot.json`) returns `404` and does not leak the file

Verified the test fails against `master` (both traversal cases return `200` with the canary body) and passes with the fix. The existing `serve.default` suite still passes (49 passing together), confirming normal static serving is unaffected.

### DIFF
--- a/packages/cli/src/lifecycles/serve.js
+++ b/packages/cli/src/lifecycles/serve.js
@@ -232,8 +232,11 @@ async function getStaticServer(compilation, composable) {
   app.use(async (ctx, next) => {
     try {
       const url = new URL(`.${ctx.url.replace(basePath, "")}`, outputDir.href);
+      // reject paths that resolve outside the output directory (e.g. `../` traversal);
+      // outputDir.href ends in `/`, so this boundary also excludes sibling-prefix paths
+      const isWithinOutputDir = url.href.startsWith(outputDir.href);
 
-      if (await checkResourceExists(url)) {
+      if (isWithinOutputDir && (await checkResourceExists(url))) {
         const resourcePlugins = standardResourcePlugins
           .filter((plugin) => plugin.isStandardStaticResource)
           .map((plugin) => {
--- a/packages/cli/test/cases/serve.default.path-traversal/greenwood.config.js
+++ b/packages/cli/test/cases/serve.default.path-traversal/greenwood.config.js
@@ -1,0 +1,3 @@
+export default {
+  port: 8182,
+};
--- a/packages/cli/test/cases/serve.default.path-traversal/secret-outside-webroot.json
+++ b/packages/cli/test/cases/serve.default.path-traversal/secret-outside-webroot.json
@@ -1,0 +1,1 @@
+{ "canary": "THIS_FILE_IS_OUTSIDE_THE_WEB_ROOT" }
--- a/packages/cli/test/cases/serve.default.path-traversal/serve.default.path-traversal.spec.js
+++ b/packages/cli/test/cases/serve.default.path-traversal/serve.default.path-traversal.spec.js
@@ -1,0 +1,139 @@
+/*
+ * Use Case
+ * Run Greenwood serve command and make sure the static file handler cannot be
+ * tricked into reading files outside the output directory via `../` traversal.
+ *
+ * User Result
+ * Should return the requested in-root asset, but 404 for any path that resolves
+ * outside the output directory.
+ *
+ * User Command
+ * greenwood serve
+ *
+ * User Config
+ * {
+ *   port: 8182
+ * }
+ *
+ * User Workspace
+ * src/
+ *   pages/
+ *     index.html
+ *   assets/
+ *     data.json
+ * secret-outside-webroot.json  (sibling of the build output, must never be served)
+ *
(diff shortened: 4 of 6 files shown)


thescientist13 (approved):
Ah, great catch!

PR #1728, opened 2026-07-19
fix(cli): #1585 make css bundling static asset emit sync

thescientist13:
## Related Issue

resolves #1585 

## Documentation 

N / A

## Summary of Changes

1. make css bundling static asset emit sync

> _Could probably release this as 0.34.1_

----

Did a little splunking (with AI) with and realized the issue was coming from CSS resource plugin, which bundles static assets in CSS files, _**not**_ the copy plugin like I thought 🫨 

In this case, the operation needed to be sync since it's happening in an sync context inside the AST walker

## TODO
1. [ ] review for this case - https://github.com/ProjectEvergreen/greenwood/issues/1585#issuecomment-3393372618

PR #1750, opened 2026-08-02
fix(cli): #1585 copy css url() assets synchronously to avoid concurrent duplicate copies

### LINKED ISSUE #1585: windows builds fail very intermittently on resource locked error
### What happened?

The Greenwood monorepo CI for Windows fails intermittently every once in a while on this error
```
node:internal/fs/promises:623
  return await PromisePrototypeThen(
         ^

Error: EBUSY: resource busy or locked, copyfile 'D:\a\greenwood\greenwood\www\assets\link.png' -> 'D:\a\greenwood\greenwood\public\assets\link.1200825667.png'
    at async Object.copyFile (node:internal/fs/promises:623:10) {
  errno: -4082,
  code: 'EBUSY',
  syscall: 'copyfile',
  path: 'D:\\a\\greenwood\\greenwood\\www\\assets\\link.png',
  dest: 'D:\\a\\greenwood\\greenwood\\public\\assets\\link.1200825667.png'
}

Node.js v22.19.0
```

Was [reported in Discord](https://discord.com/channels/1015649977780936775/1334986449640226867/1334986449640226867) a while ago too

### Steps to reproduce

- https://github.com/ProjectEvergreen/greenwood/actions/runs/18062860813/job/51401574111
- https://github.com/ProjectEvergreen/greenwood/actions/runs/14424324170/job/40450764566
- https://github.com/ProjectEvergreen/greenwood/actions/runs/18363585455/job/52349315299

### Environment

- NodeJS
- (only?) Windows

### Additional Context

Would be good to at least know what causes the error, as it seems something exclusive to Windows.  maybe it's just flaky CI?  (though I think I have seen this once or twice when running in Parallels)?

Couple other issues being tracked here
1. [ ] https://github.com/ProjectEvergreen/greenwood/pull/1573#issuecomment-3393334134
1. [ ] https://github.com/ProjectEvergreen/greenwood/pull/1776
1. [ ] https://github.com/ProjectEvergreen/greenwood/actions/runs/31270673751/job/93138010596

jstockdi:
## Related Issue

Resolves #1585

Most recently observed instance: the Windows `build (22)` job on #1747's CI run ([job link](https://github.com/ProjectEvergreen/greenwood/actions/runs/30770248106/job/91556054517)), failing in `build.default.import-node-modules` on `fontawesome-webfont.eot`.

## Summary of Changes

**Root cause.** `bundleCss` fires an unawaited `fs.promises.copyFile` for every `url()` reference it walks ([plugin-standard-css.js#L137](https://github.com/ProjectEvergreen/greenwood/blob/598ec15618871580e83a37f00cc27173dcf9e2e5/packages/cli/src/plugins/resource/plugin-standard-css.js#L137)). When the same source file is referenced from more than one `url()` — e.g. font-awesome's `fontawesome-webfont.eot?v=4.7.0` plus its `?#iefix` IE fallback on the very next line, or any asset shared across stylesheets — two copies to the same content-hashed destination start in the same synchronous tick of the CSS AST walk. On Windows, `copyfile` takes an exclusive handle on the destination, so the second copy intermittently fails with `EBUSY`; because the promise is fire-and-forget, that error surfaces as an unhandled rejection that crashes the entire build. Linux and macOS tolerate concurrent writes to the same path, which is why only the Windows CI jobs ever see it.

**Fix.** Copy synchronously, skipping destinations that already exist — the destination name is content-hashed, so an existing file already has the right contents. Synchronous copies cannot interleave (eliminating the race by construction), a genuine copy failure now throws catchably instead of killing the process, and sync `fs` matches the surrounding code, which already uses `readFileSync` on the same asset for hashing. Awaiting the promises instead was considered, but `bundleCss` is synchronous and recursive, so that would have meant refactoring its whole call chain.

**Regression test.** `build.default.css-duplicate-asset-references` builds a workspace whose stylesheet references the same font from multiple `url()`s (the `?#iefix` pattern). Since the race is benign on Linux, the Windows destination-locking semantics are emulated deterministically by `windows-copyfile-emulation.js`, loaded into the spawned CLI process via `NODE_OPTIONS --import`: it rejects with `EBUSY` when a `fs.promises.copyFile` targets a destination that already has a copy in flight, exactly as win32 does. Happy to drop or rework the shim if you'd prefer a different approach.

Proven to fail on pristine `master` with the same signature as the CI flake:

```
1) Build Greenwood With:
     Default Greenwood Configuration and duplicate CSS asset references
       "before all" hook ...:
   Error: the string "...
   Error: EBUSY: resource busy or locked, copyfile '.../src/fonts/example.woff?v=1.0' -> '.../public/fonts/example.JMHWbgbO.woff?v=1.0'
       at Object.copyFile (.../windows-copyfile-emulation.js:17:19)
       at Object.enter (.../packages/cli/src/plugins/resource/plugin-standard-css.js:137:25)
       at walkNode (.../css-tree/lib/walker/create.js:169:36)
   ...
   Node.js v22.20.0
   " was thrown, throw an Error :)
```

and to pass with the fix (`5 passing`), alongside `build.default.import-node-modules` (`14 passing`) and `build.config.optimization-default` (`16 passing`). Lint, type check, and format check all clean.

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-standard-css.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-css.js
@@ -136,12 +136,7 @@ function bundleCss(body, sourceUrl, compilation, workingUrl) {
 
             const destinationUrl = new URL(`.${finalValue}`, outputDir);
 
-            // copy synchronously, skipping destinations already written, since the same source
-            // can be referenced from multiple url()s (e.g. an eot ?#iefix fallback) and
-            // concurrent copies to the same destination intermittently fail on Windows with
-            // EBUSY, crashing the build as an unhandled rejection.  the destination name is
-            // content-hashed, so an existing file can safely be kept as is
-            // https://github.com/ProjectEvergreen/greenwood/issues/1585
+            // destination names are content-hashed, so an existing file is already correct
             if (!fs.existsSync(destinationUrl)) {
               fs.copyFileSync(resolvedUrl, destinationUrl);
             }
--- a/packages/cli/src/plugins/resource/plugin-standard-css.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-css.js
@@ -134,7 +134,17 @@ function bundleCss(body, sourceUrl, compilation, workingUrl) {
               recursive: true,
             });
 
-            fs.promises.copyFile(resolvedUrl, new URL(`.${finalValue}`, outputDir));
+            const destinationUrl = new URL(`.${finalValue}`, outputDir);
+
+            // copy synchronously, skipping destinations already written, since the same source
+            // can be referenced from multiple url()s (e.g. an eot ?#iefix fallback) and
+            // concurrent copies to the same destination intermittently fail on Windows with
+            // EBUSY, crashing the build as an unhandled rejection.  the destination name is
+            // content-hashed, so an existing file can safely be kept as is
+            // https://github.com/ProjectEvergreen/greenwood/issues/1585
+            if (!fs.existsSync(destinationUrl)) {
+              fs.copyFileSync(resolvedUrl, destinationUrl);
+            }
           }
 
           optimizedCss += `url('${basePath}${finalValue}')`;
--- a/packages/cli/test/cases/build.default.css-duplicate-asset-references/build.default.css-duplicate-asset-references.spec.js
+++ b/packages/cli/test/cases/build.default.css-duplicate-asset-references/build.default.css-duplicate-asset-references.spec.js
@@ -1,0 +1,113 @@
+/*
+ * Use Case
+ * Run Greenwood build command with no config and a stylesheet that references the same
+ * asset from more than one url(), e.g. the classic eot ?#iefix @font-face fallback.
+ * Each url() fires its own copy of the same source file to the same hashed destination,
+ * which intermittently fails on Windows with EBUSY (destination locked by the concurrent
+ * copy) and crashes the build as an unhandled rejection.  The Windows locking behavior
+ * is emulated deterministically via windows-copyfile-emulation.js, loaded into the CLI
+ * process through NODE_OPTIONS --import.
+ * https://github.com/ProjectEvergreen/greenwood/issues/1585
+ *
+ * User Result
+ * Should generate a bare bones Greenwood build with the referenced font copied once.
+ *
+ * User Command
+ * greenwood build
+ *
+ * User Config
+ * None (Greenwood Default)
+ *
+ * User Workspace
+ *  src/
+ *   fonts/
+ *     example.woff
+ *   pages/
+ *     index.html
+ *   styles/
+ *     main.css
+ */
+import { expect } from "chai";
+import fs from "node:fs/promises";
+import path from "node:path";
+import { runSmokeTest } from "../../../../../test/smoke-test.js";
+import { getOutputTeardownFiles } from "../../../../../test/utils.js";
+import { Runner } from "gallinago";
+import { fileURLToPath, pathToFileURL } from "node:url";
+
+describe("Build Greenwood With: ", function () {
+  const LABEL = "Default Greenwood Configuration and duplicate CSS asset references";
+  const cliPath = path.join(process.cwd(), "packages/cli/src/bin.js");
+  const outputPath = fileURLToPath(new URL(".", import.meta.url));
+  let runner;
+
+  before(function () {
+    this.context = {
+      publicDir: path.join(outputPath, "public"),
+    };
+    runner = new Runner();
+  });
+
+  describe(LABEL, function () {
+    before(async function () {
+      const shimUrl = pathToFileURL(path.join(outputPath, "windows-copyfile-emulation.js"));
+      const previousNodeOptions = process.env.NODE_OPTIONS;
+
+      process.env.NODE_OPTIONS = `--import=${shimUrl.href}`;
+
+      try {
+        await runner.setup(outputPath);
+        await runner.runCommand(cliPath, "build");
+      } finally {
+        if (previousNodeOptions === undefined) {
+          delete process.env.NODE_OPTIONS;
+        } else {
+          process.env.NODE_OPTIONS = previousNodeOptions;
+        }
+      }
+    });
+
+    runSmokeTest(["public"], LABEL);
+
+    describe("Bundled CSS with duplicate references to the same font file", function () {
+      let styles;
+      let fonts;
+
+      before(async function () {
+        styles = await Array.fromAsync(
+          fs.glob("main.*.css", { cwd: new URL("./public/styles/", import.meta.url) }),
+        );
+        fonts = await Array.fromAsync(
+          fs.glob("example.*.woff", { cwd: new URL("./public/fonts/", import.meta.url) }),
+        );
+      });
+
+      it("should reference the hashed font file from the bundled stylesheet", async function () {
+        const stylesContents = await fs.readFile(
+          new URL(`./public/styles/${styles[0]}`, import.meta.url),
+          "utf-8",
+        );
+
+        expect(stylesContents).to.contain(`/fonts/${fonts[0]}`);
+      });
+
+      it("should copy the referenced font file once with its original contents", async function () {
+        const sourceContents = await fs.readFile(
+          new URL("./src/fonts/example.woff", import.meta.url),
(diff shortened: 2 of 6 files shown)


thescientist13 (commented):
Hah, I have an [open PR](https://github.com/ProjectEvergreen/greenwood/pull/1728/) for this too :) 

I see you definitely dug into the issue a bit more and the root cause analysis seems to make sense to me.  One thing that would be helpful to fully validate this, is can we confirm that old website build in this monorepo (www) has this issue?  Presumably there must be an asset referenced multiple times like in your example?

packages/cli/test/cases/build.default.css-duplicate-asset-references/build.default.css-duplicate-asset-references.spec.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+    runner = new Runner();
+  });
+
+  describe(LABEL, function () {
+    before(async function () {
+      const shimUrl = pathToFileURL(path.join(outputPath, "windows-copyfile-emulation.js"));
+      const previousNodeOptions = process.env.NODE_OPTIONS;
+
+      process.env.NODE_OPTIONS = `--import=${shimUrl.href}`;
```
thescientist13, line 56: can you explain this part a bit more and how it works to reproduce the issue?

jstockdi: Sure — the spec reproduces the Windows-only race deterministically on any OS by emulating Windows' destination locking:

1. `gallinago` spawns the CLI as a child `node` process, and the child inherits the parent's env — so setting `NODE_OPTIONS=--import=<shim>` around `runner.runCommand()` loads `windows-copyfile-emulation.js` into the CLI process before any CLI code runs.
2. The shim wraps `fs.promises.copyFile` and tracks in-flight destinations: if a copy starts while another copy to the same destination is still pending, it rejects with the same `EBUSY` error Windows produces (matching `code`/`errno`/`syscall`). First copy in wins, overlapping copy fails — which is exactly the destination-locking semantics behind the CI failures.
3. The fixture's stylesheet references the same font from multiple `url()`s (the classic eot `?#iefix` fallback), so on master `bundleCss` fires two fire-and-forget `copyFile`s to the same hashed destination, the second one rejects, and — because nothing awaits it — the build dies as an unhandled rejection, exactly like the CI logs in #1585. With the fix the copies never overlap and the build passes.

One deliberate choice worth calling out: the shim wraps only `fs.promises.copyFile`, not `copyFileSync`. That's not the test dodging its own trap — a sync copy can't overlap another copy in a single-threaded process, which is precisely why switching to sync is the fix; there's no overlapping-copy semantic for the shim to emulate on the sync path.

jstockdi:
Confirmed — the www workspace does have the shape for this, and it lines up with the exact failure in #1585 (`link.png` → `public/assets/link.1200825667.png`):

- `link.png` is referenced from a single `url()` in `www/src/styles/page.css`, so it's not the multiple-`url()`s-in-one-file case.
- But `page.css` is linked from **two layouts** (`blog.html` and `docs.html`), and www builds with `prerender: true`. `bundleCss` runs every time that stylesheet is served during prerender (via the plugin's `intercept`) and again in the bundle phase — and since the `copyFile` was fire-and-forget (never awaited), copies of the same asset from successive passes can still be in flight simultaneously. Same content-hashed destination → EBUSY under Windows' destination locking.
- The eot `?#iefix` double-reference (the `fontawesome-webfont.eot` failure I hit on #1747's Windows job, in `build.default.import-node-modules`) is the same race compressed into a single `bundleCss` pass — which is why the regression fixture uses that pattern.

Re #1728 — happy to reconcile however you prefer: land yours and I'll rebase this down to the regression test + the skip-existing guard on top, or fold whatever's useful from here into #1728. The skip-existing check is worth keeping either way — destinations are content-hashed, so re-copying the same asset once per referencing pass is redundant I/O even after the copies are made safe.

packages/cli/src/plugins/resource/plugin-standard-css.js
```diff
@@ -134,7 +134,17 @@ function bundleCss(body, sourceUrl, compilation, workingUrl) {
               recursive: true,
             });
 
-            fs.promises.copyFile(resolvedUrl, new URL(`.${finalValue}`, outputDir));
+            const destinationUrl = new URL(`.${finalValue}`, outputDir);
+
+            // copy synchronously, skipping destinations already written, since the same source
```
thescientist13: nit: maybe we could trim this down or even remove the comment entirely I think? it was just a small oversight on my part to do an async operation in a sync context

jstockdi: Trimmed in `c588d01f` down to one line — just why skipping an existing destination is safe (the name is content-hashed). The EBUSY/Windows background and the issue link are gone from the source; that story lives in the issue and the PR body.

thescientist13: yeah, it's also captured in the test case too, which I think is fine if wanting to capture more context / history, so I think this is good now.

thescientist13 (approved):
Thanks, I closed https://github.com/ProjectEvergreen/greenwood/pull/1728 in favor of this one, thanks for adding the test case here 💯

packages/cli/test/cases/build.default.css-duplicate-asset-references/build.default.css-duplicate-asset-references.spec.js
thescientist13: So I guess in general, this test would really just catching _anything_ being copied using `fs.promise.copyFile` yes, not just in our css bundling pipeline?

Was just thinking that since this was intermittent to begin with, how deterministic can this test case be?  And would / could it lead to false positives happening elsewhere outside the Greenwood build pipeline?

jstockdi:
Good questions — taking them in order.

**Does it catch any `copyFile`, or just the css pipeline?**

The shim patches `fs.promises.copyFile` process-wide, but it does not fail on copies as such. It only rejects when a second copy to the **same destination** starts while a first one to that destination is still in flight:

```js
if (inFlightDestinations.has(key)) { /* EBUSY */ }
```

Sequential copies to one destination, and concurrent copies to different destinations, both pass straight through. So the trigger is specifically the overlapping-duplicate-destination pattern, which is the bug.

**How deterministic is it?**

Fully — that is the reason for the shim. The Windows failure was timing-dependent; the emulation removes the timing by making the overlap *always* throw. Measured just now on `37e68db3`, same machine, six consecutive runs:

| source | result |
|---|---|
| with the fix | 5 passing, 5 passing, 5 passing |
| fix reverted | 1 failing, 1 failing, 1 failing |

All three failures are the same error at the same line — `EBUSY … copyfile 'example.woff' -> 'example.JMHWbgbO.woff'` thrown from `plugin-standard-css.js:137`. No flake in either direction.

**Could it false-positive outside the build pipeline?**

No, on two counts.

Scope: `NODE_OPTIONS` is set in this case's `before` hook and restored in a `finally`, so the shim only ever loads into the single CLI process this one case spawns. It is not global to the mocha run and nothing is installed in the repo.

More to the point, the fix switched that call to `fs.copyFileSync`, so on fixed code the shim never intercepts this path at all — it is inert. It cannot false-positive on the current implementation; it can only fire if someone reintroduces an async duplicate copy to one destination.

The one honest caveat: if some *other* part of the build started doing two concurrent copies to a single destination during this fixture's build, this case would catch that too. Nothing does today, and I would argue catching it is a feature rather than a false positive — it is the same bug class as #1764.

thescientist13:
Ah, ok, I think I get it now.

Since the shim has a `Set` in it, you're basically testing that if there is _any_ duplicated copy call, which should not ever happen now that we've also added the `fs.existsSync` check, then in theory this is more of a test of _that_ check, per se.  Would that be an accurate understanding?

thescientist13:
Only other thought on this approach is that if we implemented #71 , one hope I had was that we could start running our tests in parallel, to help speed up CI, in particular.

Do you think this test case would introduce any concern there, in terms of leaking over into parallel test cases?  I _assume_ mocha would probably spin up threads to make that possible, and so if that was the case, then the process isolation would keep this `--import` shim approach contained to the process?

Just thinking out loud.  WDYT?

edit: added a callout for this in #71 👍

PR #1754, opened 2026-08-04
feature(plugins): #357 add html minification plugin

### LINKED ISSUE #357: html minification

jstockdi:
## Related Issue

Resolves #357

Design proposal with rationale posted on the issue: https://github.com/ProjectEvergreen/greenwood/issues/357#issuecomment-5178298232

## Documentation

Package README included in this PR. If accepted as first-party, I will open a follow-up PR to [www.greenwoodjs.dev](https://github.com/ProjectEvergreen/www.greenwoodjs.dev) adding a plugin docs page.

## Summary of Changes

1. [x] new package `@greenwood/plugin-minify-html` — an opt-in resource plugin using the existing `shouldOptimize` / `optimize` lifecycle, so it only affects `greenwood build` output
1. [x] deliberately conservative transform (~50 lines): strips HTML comments and collapses whitespace runs to a single space (the same collapsing browsers perform at render time); never removes whitespace outright, so inline-element gaps and rendering semantics are untouched
1. [x] preserves comments matching `ignoreCustomComments` patterns, defaulting to `[/^!/, /^\/?lit-/]` so Lit SSR hydration markers (`<!--lit-part ...-->` / `<!--/lit-part-->`) and license style comments survive; option is user-overridable
1. [x] `<pre>`, `<textarea>`, `<script>`, `<style>`, and `<noscript>` subtrees are left byte-identical (no inline JS/CSS minification — the bundler already runs terser over scripts)
1. [x] zero new dependencies in the tree — implemented with `node-html-parser`, already a `@greenwood/cli` dependency (declared at the same `^7.0.1`)
1. [x] two fixture-based test cases (`default`, `option-ignore-custom-comments`) covering comment stripping, marker preservation, whitespace collapsing, inline gaps, and `<pre>` fidelity — all passing locally alongside the common smoke tests
1. [x] the default case fixture is added to `.prettierignore` (intentional `<pre>` formatting, following the existing fixture precedent)
1. [x] `yarn lint` (ls-lint, eslint, tsc types) and `yarn format:check` clean

One observation from fixture testing, not addressed here: comments authored inside a page's `<head>` are already dropped by core's layout merging before any resource plugin runs, so this plugin only governs body comments in practice.

No breaking changes.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -7,4 +7,5 @@ packages/cli/test/cases/develop.default.hud/src/pages/index.html
 packages/cli/test/cases/develop.config.polyfills-import-attributes/src/components/hero.js
 packages/cli/test/cases/develop.plugins.context/fixtures/my-layouts/app.html
 packages/cli/test/cases/**/wasm_hello_world.js
+packages/plugin-minify-html/test/cases/default/src/pages/index.html
 /.nx/workspace-data
--- a/packages/plugin-minify-html/README.md
+++ b/packages/plugin-minify-html/README.md
@@ -1,0 +1,75 @@
+# @greenwood/plugin-minify-html
+
+## Overview
+
+A Greenwood plugin for minifying your HTML output at build time.  It deliberately sticks to the two transforms that are safe by construction:
+
+- **Strips HTML comments**, preserving comments that frameworks rely on (Lit SSR hydration markers like `<!--lit-part ...-->`) and license style `<!--! ... -->` comments
+- **Collapses whitespace runs to a single space** — the same collapsing browsers already perform when rendering — leaving `<pre>`, `<textarea>`, `<script>`, `<style>`, and `<noscript>` content untouched
+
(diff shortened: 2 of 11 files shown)


thescientist13 (commented):
Thanks for the PR!

I opted-to leave  some comments in the issue, as I would prefer to discuss approach a bit more first before we get to any sort of implementation just yet.
https://github.com/ProjectEvergreen/greenwood/issues/357#issuecomment-5199343947

Let me put this into draft for the time being.

PR #1759, opened 2026-08-06
fix(cli): #1758 walk array pattern exports without crashing and isolate import map failures per dependency

### LINKED ISSUE #1758: import map: subpath pattern with an array target crashes the walk and drops all remaining dependencies
### What happened?

If any dependency's `package.json` has a subpath **pattern** export whose target is an **array** ("fallback" form), generating the import map in `greenwood develop` crashes the dependency walk — and because the whole dependency loop sits inside a single `try/catch`, **every dependency after the offending one is silently dropped from the import map**. The only signal is a stack trace in the develop console:

```
Error building up import map TypeError: Invalid URL
    at walkExportPatterns (walker-package-ranger.js:95)
    at walkPackageForExports (walker-package-ranger.js:190)
... { code: 'ERR_INVALID_URL', input: 'undefined' }
```

The user-visible symptom is mass `Failed to resolve module specifier` errors in the browser for packages that are themselves perfectly fine — which makes it very hard to trace back to the one third-party package with the unusual (but valid per Node) exports shape. `greenwood build` is unaffected (Rollup resolves correctly), so this presents as a confusing dev/prod divergence.

### Steps to reproduce

1. Install a dependency whose `package.json` contains e.g.:

```json
{
  "exports": {
    ".": "./index.js",
    "./*": [{ "import": "./src/*.js" }]
  }
}
```

2. Have at least one more dependency listed after it in the project's `package.json`.
3. Run `greenwood develop` and inspect the generated `<script type="importmap">`.

Observed: the crasher's `.` entry is present, its pattern subpaths are missing, and **every dependency after it has no entries at all** (the browser then throws `TypeError: Failed to resolve module specifier "..."` for those). Expected: entries for all dependencies.

### Environment

- Greenwood v0.34.0 (reproduced on current `master`)
- NodeJS v22.20
- Linux

### Additional Context

Root cause, two parts:

1. `packages/cli/src/lib/walker-package-ranger.js:190` — `walkExportPatterns(dependency, sub, exports[sub][condition], resolvedRoot)` passes 4 arguments to the 3-parameter `walkExportPatterns(dependency, condition, resolvedRoot)` (line 89). The third argument is `exports[sub][condition]` — but `exports[sub]` is the **array** here, so indexing it by a condition name yields `undefined`, which lands in the `resolvedRoot` parameter and throws at `new URL(undefined)` (line 95). The correct value is `item[condition]` from the array item already matched two lines above.
2. `walkPackageJson` (line 279) wraps the **entire** dependency loop in one `try/catch`, so any throw from one package's walk aborts all remaining packages instead of just the offending one.

Happy to submit a PR — fix for the array-pattern call plus per-dependency error isolation, with a regression test covering both the pattern subpath entry and the survival of subsequent dependencies.

(Found during a broader audit of the import-map walker; keying of pattern subpaths (LHS vs RHS), array "." exports, and legacy flat-`main` globbing have separate issues to follow.)

jstockdi:
## Related Issue

Resolves #1758

## Documentation

## Summary of Changes

1. [x] `walkPackageForExports` (`packages/cli/src/lib/walker-package-ranger.js`): for a subpath **pattern** export with an **array** target (`"./*": [{ "import": "./src/*.js" }]`), walk the pattern from the matched array item (`item[condition]`) instead of calling `walkExportPatterns` with a 4th argument and `exports[sub][condition]` — which indexes the array by a condition name, yields `undefined`, and throws `TypeError: Invalid URL` at `new URL(undefined)`.
2. [x] `walkPackageJson`: move the `try/catch` inside the per-dependency loop, so a package with an unexpected exports shape degrades to a `console.error` naming that package instead of silently dropping **every remaining dependency** from the import map. (The error message now includes the dependency name for traceability.)
3. [x] Regression test `develop.default.import-map-array-pattern-exports`: two fake packages (the array-pattern shape above plus a plain dependency listed after it), asserting the crasher's `.` entry, its pattern subpath entry, and — the headline — that the subsequent dependency still gets its entry. On unpatched `master` the last two fail (subpath missing; subsequent dependency dropped from the map):

```
  1 passing (548ms)
  2 failing
  1) ... should have an import map entry for the subpath resolved from the array pattern target:
     AssertionError: expected +0 to be above +0
  2) ... should still have an import map entry for the dependency listed after the array pattern exports package:
     AssertionError: expected undefined to not equal undefined
```

With the fix, all 3 pass; `develop.default` (import-map snapshot) and `develop.config.polyfills-import-maps` still pass (120 specs), and lint / format / ls-lint are clean.

### Test setup note

The import map walker resolves dependencies with `import.meta.resolve` from the CLI's own module location, not from the test project directory — so the spec stages its two fake packages into the monorepo's `node_modules/` in `before()` and removes them in `after()`. Flagging it since it's a new pattern for a test case; happy to restructure if there's a preferred way to get fixture packages onto the resolver's walk-up path.

### Scope / follow-up

This is deliberately just the crash + truncation fix. The walker has separate keying issues found in the same audit (pattern subpaths keyed by the RHS target path instead of the LHS subpath; array "." exports keyed by target paths; legacy flat-`main` globbing the whole package tree) — issues to follow so each can be reviewed on its own.

### DIFF
--- a/packages/cli/src/lib/walker-package-ranger.js
+++ b/packages/cli/src/lib/walker-package-ranger.js
@@ -186,8 +186,14 @@ async function walkPackageForExports(dependency, packageJson, resolvedRoot) {
               if (item[condition]) {
                 matched = true;
                 if (sub.indexOf("*") >= 0) {
-                  // basic support, not properly tested
-                  await walkExportPatterns(dependency, sub, exports[sub][condition], resolvedRoot);
+                  // walk the pattern target from the matched array item; indexing the array
+                  // itself by condition name would pass undefined as the resolved root
+                  // https://github.com/ProjectEvergreen/greenwood/issues/1758
+                  await walkExportPatterns(
+                    dependency,
+                    item[condition].default ?? item[condition],
+                    resolvedRoot,
+                  );
                 } else {
                   // could there be more sub conditions here?  Going with default for now
                   updateImportMap(
@@ -276,10 +282,14 @@ async function walkPackageForExports(dependency, packageJson, resolvedRoot) {
 
 // we recursively cache / memoize walkedPackages to account for scenarios where Greenwood can (pre)render concurrently
 async function walkPackageJson(packageJson = {}, walkedPackages = new Set()) {
-  try {
-    const dependencies = Object.keys(packageJson.dependencies || {});
+  const dependencies = Object.keys(packageJson.dependencies || {});
 
-    for (const dependency of dependencies) {
+  for (const dependency of dependencies) {
+    // isolate failures per dependency, so one package with an unexpected exports shape
+    // degrades to an error for that package instead of silently dropping every remaining
+    // dependency from the import map
+    // https://github.com/ProjectEvergreen/greenwood/issues/1758
+    try {
       const resolved = resolveBareSpecifier(dependency);
 
       if (resolved) {
@@ -311,9 +321,9 @@ async function walkPackageJson(packageJson = {}, walkedPackages = new Set()) {
           }
         }
       }
+    } catch (e) {
+      console.error(`Error building up import map for dependency => \`${dependency}\``, e);
     }
-  } catch (e) {
-    console.error("Error building up import map", e);
   }
 
   return { importMap, diagnostics };
--- a/packages/cli/test/cases/develop.default.import-map-array-pattern-exports/develop.default.import-map-array-pattern-exports.spec.js
+++ b/packages/cli/test/cases/develop.default.import-map-array-pattern-exports/develop.default.import-map-array-pattern-exports.spec.js
@@ -1,0 +1,121 @@
+/*
+ * Use Case
+ * Run Greenwood develop command with a dependency whose package.json exports a subpath
+ * pattern with an array ("fallback") target, e.g. "./*": [{ "import": "./src/*.js" }],
+ * followed by another perfectly valid dependency.
+ *
+ * User Result
+ * Should generate an import map covering both dependencies, instead of crashing the
+ * walk on the array target and silently dropping every dependency after it.
+ *
+ * User Command
+ * greenwood develop
+ *
+ * User Config
+ * None (Greenwood Default)
+ *
+ * User Workspace
+ *  src/
+ *   pages/
+ *     index.html
+ * package.json (fake-array-pattern-exports, fake-subsequent-dependency)
+ *
+ * The two fake packages live in ./fixtures and are staged into the monorepo's own
+ * node_modules for the duration of the suite, since the import map walker resolves
+ * dependencies with import.meta.resolve from the CLI's location, not the project's.
+ */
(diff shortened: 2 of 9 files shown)


jstockdi:
This one keeps a package whose `exports` use an array target from taking down the whole import map walk. Right now that throw hits the loop-wide catch, so every dependency after it silently drops out of the map and develop mode ends up missing packages that had nothing to do with the failure. Scope is deliberately just the walker; the other import map findings I have (map keying, gating, and payload bloat) are held back for separate PRs to keep this one small and reviewable. CI is green with no conflicts, it is only behind main, so I can rebase it or narrow it further if that makes it easier to take.

packages/cli/src/lib/walker-package-ranger.js
```diff
@@ -276,10 +282,14 @@ async function walkPackageForExports(dependency, packageJson, resolvedRoot) {
...
-    const dependencies = Object.keys(packageJson.dependencies || {});
-
-    for (const dependency of dependencies) {
+  const dependencies = Object.keys(packageJson.dependencies || {});
+
+  for (const dependency of dependencies) {
+    // isolate failures per dependency, so one package with an unexpected exports shape
+    // degrades to an error for that package instead of silently dropping every remaining
+    // dependency from the import map
+    // https://github.com/ProjectEvergreen/greenwood/issues/1758
```
thescientist13, line 291: i don't think we need the issue comment here, this just seems like a general purpose callout

packages/cli/src/lib/walker-package-ranger.js
```diff
@@ -186,8 +186,14 @@ async function walkPackageForExports(dependency, packageJson, resolvedRoot) {
               if (item[condition]) {
                 matched = true;
                 if (sub.indexOf("*") >= 0) {
-                  // basic support, not properly tested
-                  await walkExportPatterns(dependency, sub, exports[sub][condition], resolvedRoot);
+                  // walk the pattern target from the matched array item; indexing the array
+                  // itself by condition name would pass undefined as the resolved root
+                  // https://github.com/ProjectEvergreen/greenwood/issues/1758
```
thescientist13, line 191: do you have an example package that reproduces this?  we generally like to capture a link to unpkg for a the dependency that exhibits the use case were trying to handle, for reference

packages/cli/src/lib/walker-package-ranger.js
```diff
@@ -186,8 +186,14 @@ async function walkPackageForExports(dependency, packageJson, resolvedRoot) {
...
+                  await walkExportPatterns(
+                    dependency,
```
thescientist13, line 193: oh yeah, this was definitely being called wrong with four params 😬

thescientist13 (commented):
Left some general comments and have a follow up on testing, but first would like to understand this - https://github.com/ProjectEvergreen/greenwood/issues/1758#issuecomment-5433462507

PR #1761, opened 2026-08-06
fix(cli): #1760 dedupe shared script and link tags across app layout, page layout, and page

### LINKED ISSUE #1760: shared script/link referenced by app layout, page layout, and page is emitted (and preloaded) once per occurrence
### What happened?

A `<script src>` or `<link rel="stylesheet">` referenced by more than one of the app layout, page layout, and page is emitted once **per occurrence** in the built HTML — for the natural pattern "app layout has the shared bundle, page layout and page also reference it", the output contains the same tag three times, plus three `modulepreload` / `preload` hints for it.

Consequences beyond bloat:

- A **classic** (non-module) duplicate executes repeatedly — double event listeners, double side effects — since only module semantics dedupe by URL in the browser.
- Duplicated stylesheets apply twice, which can produce cascade-order surprises.
- Duplicated preload hints waste requests.

There is no warning; the developer only notices when inspecting the output (or debugging a double-fired listener).

### Steps to reproduce

1. Reference the same resource in all three places, e.g.:

```html
<!-- src/layouts/app.html and src/layouts/page.html and src/pages/index.html -->
<script type="module" src="/scripts/shared.js"></script>
<link rel="stylesheet" href="/styles/shared.css" />
```

2. `greenwood build`
3. Inspect `public/index.html`.

Observed: 3× the shared `<script>` tag, 3× the shared stylesheet `<link>`, 3× `modulepreload`, 3× `preload`. Expected: each once.

### Environment

- Greenwood v0.34.0 (reproduced on current `master`)
- NodeJS v22.20
- Linux

### Additional Context

Two compounding spots:

1. `packages/cli/src/lib/layout-utils.js` — `mergeContentIntoLayout` builds `mergedLinks` (line 143) and `mergedScripts` (line 186) by concatenating parent + child head tags with no dedupe by `src`/`href`, so each layout-merge pass preserves duplicates.
2. `packages/cli/src/lib/resource-utils.js:214` — `matchingRoute.resources` is built with `resources.map(...)`, preserving duplicates (only the `compilation.resources` Map is deduped), and the standard HTML plugin injects one preload hint per occurrence while rewriting each tag.

Happy to submit a PR: dedupe merged `<script src>`/`<link href>` tags in the layout merge (keeping the first occurrence, so app layout → page layout → page order is preserved), plus a `Set` on `matchingRoute.resources`, with regression assertions folded into the existing `build.default.workspace-layouts-page-and-app` test case.

(Found during the same audit pass as #1707/#1709/#1711 etc.)

jstockdi:
## Related Issue

Resolves #1760

## Documentation

## Summary of Changes

1. [x] `mergeContentIntoLayout` (`packages/cli/src/lib/layout-utils.js`): dedupe merged `<script>` tags by `src` and `<link>` tags by `href` when combining parent and child `<head>` contents (new `dedupeMergedResourceTags` helper). First occurrence wins, so the existing app layout → page layout → page merge order is preserved; tags without a `src`/`href` (inline scripts, inline styles) are never touched. Since each layout merge pass dedupes, a resource referenced by app layout + page layout + page collapses to one tag, and the standard HTML plugin then injects one `modulepreload`/`preload` hint instead of one per occurrence.
2. [x] `trackResourcesForRoute` (`packages/cli/src/lib/resource-utils.js`): dedupe `matchingRoute.resources` with a `Set`, since the same resource can legitimately be referenced by both a layout and a page (previously only the `compilation.resources` Map was deduped, while the per-route list preserved duplicates).
3. [x] Regression coverage folded into the existing `build.default.workspace-layouts-page-and-app` test case: a `shared.js` + `shared.css` referenced by the app layout, the page layout, and the page, asserting each is emitted exactly once (tag + its preload hint), with the existing merge-order assertions extended to cover them.

Proof the test catches the bug — same spec against unpatched `master` (fix stashed):

```
  6 failing
  ... should emit the shared <script> tag only once:        AssertionError: expected 3 to equal 1
  ... should emit the shared stylesheet <link> tag only once: AssertionError: expected 3 to equal 1
  ... should emit only one modulepreload <link> tag ...:      AssertionError: expected 3 to equal 1
  ... should emit only one preload <link> tag ...:            AssertionError: expected 3 to equal 1
  (+ the head <script>/<link> count assertions, 8 != 6 / 7 != 5)
```

With the fix: all 17 specs in the case pass, and the full layout suite (`build.config.layouts-directory`, all `build.default.workspace-layouts-*`, `theme-pack` develop + build — 299 specs) passes. Lint / format / ls-lint clean.

### Notes / scope

- Dedupe keys on the tag's raw `src`/`href` value, so the same file referenced by two *different* path spellings (e.g. `/scripts/shared.js` vs `../scripts/shared.js`) is still emitted per spelling — resolving to a canonical URL before comparison would be a larger change and the common case is the same absolute reference.
- A classic (non-module) duplicate of a module script also collapses to the first occurrence; previously the classic copy executed in addition to the module copy (double event listeners / side effects), which is the behavioral half of the bug beyond output bloat.


jstockdi:
Heads up before first review: CI was red on the previous head and is fixed in `3cc94d05`.

The dedupe keyed only on the `href`/`src` attribute, which collapsed two tags that are not duplicates — the standard two-tag favicon pattern in `build.default.meta`:

```html
<link rel="shortcut icon" href="/favicon.svg">
<link rel="icon" href="/favicon.svg">
```

Same `href`, different `rel`, so the second was dropped and `querySelector('head link[rel="icon"]')` returned null. That broke one test in `packages/cli/test/cases/build.default.meta` and one in `packages/plugin-markdown/test/cases/build.default.meta`.

The key is now the whitespace-normalized full tag rather than a single attribute, so only genuinely identical tags collapse and anything differing in `rel`/`type`/`media` survives. `tag.toString()` covers both the parsed-element and the generated-string branches, so the `typeof` split and the attribute argument both drop out.

Both meta cases pass again (38/38 and 37/37), and the regression case this PR added still holds (17/17). I also swept the layout and `build.default.*` cases plus the full plugin-markdown suite — 606 and 345 passing, 0 failing — since widening the key could have gone the other way.

One narrowing worth flagging rather than hiding: a `customImports` script no longer dedupes against a layout script with the same `src`, because the generated string differs textually from the parsed tag. Nothing covers that today and it is outside what #1760 targets, but say the word and I will canonicalize the key on tag name plus sorted attributes instead.

thescientist13 (commented):
Do you think it would make sense to add a develop test for this too?

thescientist13:
@jstockdi 
> One narrowing worth flagging rather than hiding: a customImports script no longer dedupes against a layout script with the same src, because the generated string differs textually from the parsed tag. Nothing covers that today and it is outside what https://github.com/ProjectEvergreen/greenwood/issues/1760 targets, but say the word and I will canonicalize the key on tag name plus sorted attributes instead.

Could you expand on this a bit more?  Are you saying that it used to dedupe on custom imports, and so this change would be a "step back" if you will?  What does the changeset look like and what additional diff would it add here?  I guess just curious your thoughts on what might the risk in accounting for them all at once?

jstockdi:
Added in `e6ac3d89` — `develop.default.workspace-layouts-page-and-app`, mirroring the build case. Verified it fails without the fix: with the dedupe neutralized the shared script and stylesheet each appear three times, once per layer.

On the `customImports` narrowing, my note was ambiguous. It isn't a step back from master. Master has no dedupe at all — an entry in `customImports` and a layout `<script src>` pointing at the same file both landed in the HTML before this PR, and both still do. What I was flagging is narrower: an earlier commit in this PR keyed on `href`/`src` alone, which did collapse those two, but that same key also collapsed `<link rel="shortcut icon">` with `<link rel="icon">` on the same href, which is what turned CI red. Moving to the full normalized tag fixed the favicon case, and the side effect is that a customImport no longer matches a layout tag.

They miss each other because `mergedLinks` and `mergedScripts` mix parsed elements from the layouts with `customImports` that are mapped into generated tag strings, so `tag.toString()` differs textually even when the two point at the same resource.

Covering them would mean keying on tag name plus sorted attribute pairs rather than the raw string, so both forms normalize to the same key. Small diff, contained in `dedupeMergedResourceTags`.

The risk is that it goes past what this issue describes. Once a customImport can collapse against a layout tag, which of the two survives decides script order in the head — and this key has already produced one regression from being widened. So I'd rather land this as-is and open a separate issue for the cross-form case with a test that pins the ordering. Happy to fold it in here instead if you'd prefer it all at once.

One note on coverage: the develop case can't reach the `resource-utils.js` half, since the `modulepreload`/`preload` tags come from the build-time optimization path and aren't in develop's served HTML. The build case stays the only coverage for that half — the two are complementary rather than one covering the other.

thescientist13:
Thanks for clarifying!

> So I'd rather land this as-is and open a separate issue for the cross-form case with a test that pins the ordering.

Agreed, this is a good fix.  I think it would be good to open a new issue for this case, and I think at the same time, we should investigate deduping `<meta>` tags too.  In general the strategy as with handling frontmatter like for `<title>` tag, is that precedence favors the page over the page layout, over the app layout.

So given this project structure
```sh
src/
  layouts/
    app.html
    blog.html
  pages/
    first-post.html
```

The priority would over who's `<meta>` tags would win out, would go in this order
1. _pages/first-post.html_
1. _layouts/blog.html_
1. _layouts/app.html_

thescientist13 (approved):

PR #1765, opened 2026-08-08
fix(cli): #1764 copy files sequentially so duplicate destinations are not corrupted

### LINKED ISSUE #1764: two copy plugins targeting the same destination produce a torn, corrupt output file
### What happened?

When two `type: "copy"` plugins resolve to the same destination path in the output directory, the build does not produce last-write-wins — it intermittently produces a **torn, corrupt file** containing bytes from both sources.

The copy lifecycle runs every copy plugin's `provider` concurrently, and then runs every `{ from, to }` pair a provider returns concurrently as well. When two of those pairs share a `to`, two `fs.copyFile()` calls target the same path at the same time. Each opens the destination with truncate and then writes at its own offset, so the shorter payload can land at offset 0 *after* the longer payload has already been written, leaving the longer file's tail behind. The result is the short file followed by the trailing bytes of the long one.

With a 110-byte source and a 50363-byte source both copied to `public/feed.xml`, a corrupted run produces a 50363-byte file that begins:

```
<?xml version="1.0"?>
<rss><channel><title>SMALL</title>
  <item><title>only</title></item>
</channel>
</rss>
BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB</description></item>
  <item><title>Item 1</title>...
```

and ends with a second `</rss>`. That is not a partial write of one input or the other, it is both interleaved, and it is a hard XML parse failure (`not well-formed (invalid token): line 6, column 47`).

Two things make this hard to catch:

- **The build reports success.** Exit code is 0, the log prints `copying file...` twice, and nothing warns that two plugins claimed the same destination.
- **It is nondeterministic.** Over 15 consecutive builds of the repro below: 7 runs produced the torn file, 7 produced the short file cleanly, and 1 produced the long file cleanly. An earlier batch of 10 runs gave 3 torn. So a project can build correctly on a developer machine and ship a corrupt asset from CI, or vice versa.

The same corruption happens when a *single* copy plugin returns two `{ from, to }` pairs with the same `to` (3 of 10 runs torn), because the inner loop is concurrent too.

For contrast, two copy plugins reading the *same source* to *different destinations* are fine — 5 of 5 runs correct. Concurrent reads are not the problem; concurrent writes to one destination are.

### Steps to reproduce

1. Create a minimal project:

```
repro/
  package.json
  greenwood.config.js
  src/
    pages/index.html
    big.xml
    small.xml
```

2. `package.json`:

```json
{
  "name": "copy-race-repro",
  "type": "module",
  "devDependencies": {
    "@greenwood/cli": "^0.34.0"
  }
}
```

3. `src/pages/index.html`:

```html
<html><body><h1>Copy race repro</h1></body></html>
```

4. Create the two payloads, with clearly different lengths so a torn write is unmistakable:

```sh
node --input-type=module -e '
import fs from "node:fs";
let big = `<?xml version="1.0"?>\n<rss><channel><title>BIG</title>\n`;
for (let i = 0; i < 400; i += 1) {
  big += `  <item><title>Item ${i}</title><description>${"B".repeat(60)}</description></item>\n`;
}
big += "</channel>\n</rss>\n";
fs.writeFileSync("src/big.xml", big);
const small = `<?xml version="1.0"?>\n<rss><channel><title>SMALL</title>\n  <item><title>only</title></item>\n</channel>\n</rss>\n`;
fs.writeFileSync("src/small.xml", small);
'
```

This gives `src/big.xml` at 50363 bytes and `src/small.xml` at 110 bytes.

5. `greenwood.config.js` — two ordinary user copy plugins, both targeting `feed.xml`:

```js
const pluginCopyBig = {
  type: "copy",
  name: "plugin-copy-big",
  provider: async (compilation) => {
    const { outputDir, userWorkspace } = compilation.context;
    return [
      {
        from: new URL("./big.xml", userWorkspace),
        to: new URL("./feed.xml", outputDir),
      },
    ];
  },
};

const pluginCopySmall = {
  type: "copy",
  name: "plugin-copy-small",
  provider: async (compilation) => {
    const { outputDir, userWorkspace } = compilation.context;
    return [
      {
        from: new URL("./small.xml", userWorkspace),
        to: new URL("./feed.xml", outputDir),
      },
    ];
  },
};

export default {
  plugins: [pluginCopyBig, pluginCopySmall],
};
```

6. Run the build several times, clearing the output between runs, and watch the size and content of `public/feed.xml`:

```sh
for i in $(seq 1 15); do
  rm -rf public .greenwood
  greenwood build > /dev/null 2>&1
  echo "run $i: $(wc -c < public/feed.xml) bytes, $(grep -c '</rss>' public/feed.xml) closing tags"
done
```

Expected: every run yields one of the two inputs intact — 110 bytes with one `</rss>`, or 50363 bytes with one `</rss>`.

Actual: many runs yield 50363 bytes with **two** `</rss>` tags — the short file's full contents followed by the tail of the long file. Roughly half the runs in my testing.

### Environment

- Greenwood: v0.34.0
- Node: 22.20
- OS: Linux

### Additional Context

Root cause is in the copy lifecycle:

- `packages/cli/src/lifecycles/copy.js:27` — `await asyncForEach(copyPlugins, ...)` iterates the copy plugins.
- `packages/cli/src/lifecycles/copy.js:30` — `await asyncForEach(locations, ...)` iterates each provider's `{ from, to }` pairs.
- `packages/cli/src/lib/async-utils.js:29` — `asyncForEach` delegates to `asyncMap`.
- `packages/cli/src/lib/async-utils.js:19` — `asyncMap` collects every callback into an array and awaits `Promise.all`, so both loops above are concurrent, not sequential.
- `packages/cli/src/lifecycles/copy.js:8` — the concurrent leaf is `await fs.copyFile(source, target)`.

`async-utils.js:18` already documents the precondition this violates: *"Constraint: mapper functions must not depend on each other."* Two copies to one destination do depend on each other — they are ordered writes to shared state.

Related but not a duplicate: **#1585** (intermittent Windows `EBUSY`) is the same family reached from a different direction. There, one plugin copies one *source* twice concurrently and Windows file locking surfaces it as an error. Here, two plugins target one *destination*, there is no error at all, and the failure mode is silent data corruption on Linux.

A likely fix is to detect duplicate destinations across the resolved copy set and fail fast, or to serialise the copy step. Happy to follow whichever direction you prefer.

jstockdi:
## Related Issue

Resolves #1764

## Documentation

N/A — no user facing API or configuration change. The only behavioural change is that when two copy operations resolve to the same destination, the last one registered now deterministically wins instead of producing a corrupt file.

## Summary of Changes

1. [x] `packages/cli/src/lifecycles/copy.js` now iterates copy plugins, and each provider's locations, with `for...of` + `await` instead of `asyncForEach`. Previously both levels fanned out through `asyncForEach` → `asyncMap` → `Promise.all`, so two `fs.copyFile` calls targeting the same destination interleaved and left a torn file — the build still exited `0`.
2. [x] `lib/async-utils.js` is deliberately untouched. It documents "Constraint: mapper functions must not depend on each other", which the copy lifecycle was violating; the other lifecycles that use it (`bundle.js`, `prerender.js`, `resource-utils.js`, `config.js`) genuinely do have independent work and still run concurrently.
3. [x] Regression coverage was folded into the existing `packages/cli/test/cases/build.plugins.copy/` case rather than a new case directory. Two new assertions cover both loops that were racing:
   - two separate copy plugins resolving to the same destination
   - one plugin returning two locations resolving to the same destination

   In both, the earlier operation is a multi-file directory copy and the later one is a single-file copy, so the slow copy reliably lands last when the two run concurrently. That makes the failure deterministic rather than the ~50% flake seen in the wild: the spec fails **10 out of 10** runs on unpatched `master` and passes **10 out of 10** runs with this change.
4. [x] No breaking changes. Configs that already had a unique destination per copy operation are unaffected.

### On the performance of serialising

Copy fan-out is very small in practice, so serialising costs essentially nothing:

- `plugin-copy-assets` returns at most one location and does a single recursive `fs.cp`
- `plugin-copy-favicon`, `plugin-copy-robots`, `plugin-copy-sitemap` return at most one location each
- `plugin-copy-manifest-json` returns exactly one
- `plugin-copy-polyfills` returns at most three

A default build with an assets directory performs two copy operations end to end (measured on `build.default.workspace-assets`: one directory copy, one file copy). The bulk of the work stays inside a single recursive `fs.cp`, whose internal parallelism is unchanged by this patch.

### Possible follow up

Erroring, or at least warning, when two copy operations resolve to the same destination would surface these configs instead of silently picking a winner. That is a behaviour change that could break a config deliberately overriding another plugin's output, so it is left out of this fix and is yours to call.

### DIFF
--- a/packages/cli/src/lifecycles/copy.js
+++ b/packages/cli/src/lifecycles/copy.js
@@ -1,5 +1,4 @@
 import fs from "node:fs/promises";
-import { asyncForEach } from "../lib/async-utils.js";
 
 async function copyFile(source, target, projectDirectory) {
   try {
@@ -24,10 +23,11 @@ const copyAssets = async (compilation) => {
   const copyPlugins = compilation.config.plugins.filter((plugin) => plugin.type === "copy");
   const { projectDirectory } = compilation.context;
 
-  await asyncForEach(copyPlugins, async (plugin) => {
+  // copies run in series so that when two of them share a destination, the last one registered wins
+  for (const plugin of copyPlugins) {
     const locations = await plugin.provider(compilation);
 
-    await asyncForEach(locations, async (location) => {
+    for (const location of locations) {
       const { from, to } = location;
 
       if (from.pathname.endsWith("/")) {
@@ -35,8 +35,8 @@ const copyAssets = async (compilation) => {
       } else {
         await copyFile(from, to, projectDirectory);
       }
-    });
-  });
+    }
+  }
 };
 
 export { copyAssets };
--- a/packages/cli/src/lifecycles/copy.js
+++ b/packages/cli/src/lifecycles/copy.js
@@ -23,7 +23,6 @@ const copyAssets = async (compilation) => {
   const copyPlugins = compilation.config.plugins.filter((plugin) => plugin.type === "copy");
   const { projectDirectory } = compilation.context;
 
-  // copies run in series so that when two of them share a destination, the last one registered wins
   for (const plugin of copyPlugins) {
     const locations = await plugin.provider(compilation);
 
--- a/packages/cli/test/cases/build.plugins.copy/build.plugins.copy.spec.js
+++ b/packages/cli/test/cases/build.plugins.copy/build.plugins.copy.spec.js
@@ -67,6 +67,17 @@ describe("Build Greenwood With: ", function () {
         "utf-8",
       );
 
+      it("should copy the whole themes directory into each shared destination", async function () {
+        for (const sharedDir of [
+          "duplicate-destination-plugins",
+          "duplicate-destination-locations",
+        ]) {
+          expect(
+            await glob.promise(path.join(this.context.publicDir, sharedDir, "*.css")),
+          ).to.have.lengthOf(16);
+        }
+      });
+
       it("should let the last of two plugins sharing a destination win", function () {
         const contents = fs.readFileSync(
           path.join(this.context.publicDir, "duplicate-destination-plugins/prism.min.css"),
--- a/packages/cli/test/cases/build.plugins.copy/build.plugins.copy.spec.js
+++ b/packages/cli/test/cases/build.plugins.copy/build.plugins.copy.spec.js
@@ -1,9 +1,13 @@
 /*
  * Use Case
- * Run Greenwood build command with custom copy plugin with deeply nested directories
+ * Run Greenwood build command with custom copy plugins with deeply nested directories, including
+ * two plugins that resolve to the same destination, and a single plugin that returns two locations
+ * resolving to the same destination
  *
  * User Result
- * Should generate a Greenwood build with a public asset folder containing contents of assets directory
+ * Should generate a Greenwood build with a public asset folder containing contents of assets
+ * directory, and with the last copy operation registered for a shared destination winning
+ * https://github.com/ProjectEvergreen/greenwood/issues/1764
  *
  * User Command
  * greenwood build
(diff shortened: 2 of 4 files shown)


jstockdi:
This is the same bug class as the CSS `url()` asset copy race in #1750 that you took recently: two copy operations targeting one destination interleave, the file lands torn, and the build still exits 0, so nothing surfaces it. The change here serializes the copies so a duplicate destination gets written once. If you would rather it match the shape you landed in that fix, I am glad to rework it to match, or narrow the scope if that reads easier. CI is green with no conflicts, it is just behind main and I can rebase whenever suits you.

packages/cli/src/lifecycles/copy.js
```diff
@@ -24,19 +23,19 @@ const copyAssets = async (compilation) => {
   const copyPlugins = compilation.config.plugins.filter((plugin) => plugin.type === "copy");
   const { projectDirectory } = compilation.context;
 
-  await asyncForEach(copyPlugins, async (plugin) => {
+  for (const plugin of copyPlugins) {
```
thescientist13, line 26: yeah, I think a comment here and link to the issue would be valuable to call out why we ended up going back to `for...of`, and in particular related to your comment here
> A likely fix is to detect duplicate destinations across the resolved copy set and fail fast, or to serialise the copy step. Happy to follow whichever direction you prefer.

Maybe instead of failing, we do a `console.warn` if we detect dups?  With a message saying something along the lines of "last one wins"?

PR #1771, opened 2026-08-08
fix(cli): #1767 serve output directory files whose extension no standard resource plugin claims

### LINKED ISSUE #1767: greenwood serve 404s files that exist in the output directory when no standard resource plugin claims the extension (sitemap.xml, robots.txt, any .xml/.txt)
### What happened?

`greenwood serve` returns `404 Not Found` for a file that exists in the output directory whenever no standard resource plugin claims that file's extension. `sitemap.xml` and `robots.txt` are the concrete cases, because Greenwood's own copy plugins put both of them into `public/` — the file is right there in the build output and the production server will not return it.

It is not limited to those two names. Any extension outside the standard plugin extension lists behaves the same way, so an arbitrary `public/assets/data.xml` or `public/assets/notes.txt` also 404s while `styles.css` and `data.json` sitting next to them are served normally.

Both server paths are affected. `packages/cli/src/commands/serve.js:10` selects `getHybridServer` when the project has dynamic routes or an `api/` directory and `getStaticServer` otherwise; `getHybridServer` composes `getStaticServer` (`serve.js:371`), so the behaviour is identical on both.

Observed on the static server (project with no API routes, port 8391):

```
GET /sitemap.xml
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
Content-Length: 9

Not Found

GET /robots.txt
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
Content-Length: 9

Not Found

GET /assets/data.xml     -> 404 Not Found   (file exists at public/assets/data.xml)
GET /assets/notes.txt    -> 404 Not Found   (file exists at public/assets/notes.txt)
GET /assets/styles.css   -> 200 OK, content-type: text/css
GET /assets/data.json    -> 200 OK, content-type: application/json
GET /                    -> 200 OK, Content-Type: text/html
```

Observed on the hybrid server (same project plus `src/pages/api/greeting.js`, port 8392) — identical:

```
GET /api/greeting        -> 200 OK, content-type: application/json
GET /sitemap.xml         -> 404 Not Found, Content-Type: text/plain; charset=utf-8, body "Not Found"
GET /robots.txt          -> 404 Not Found
GET /assets/data.xml     -> 404 Not Found
GET /assets/notes.txt    -> 404 Not Found
GET /assets/styles.css   -> 200 OK, content-type: text/css
```

`greenwood develop` returns 404 for the same paths as well (`404`, `content-type: text/plain; charset=utf-8`, empty body), so this is not a develop/serve divergence — neither command can serve these files.

### Steps to reproduce

1. Create a minimal project:

```
repro/
  greenwood.config.js
  src/
    pages/index.html
    sitemap.xml
    robots.txt
    assets/data.xml
    assets/notes.txt
    assets/styles.css
    assets/data.json
```

2. `greenwood.config.js`:

```js
export default { port: 8391 };
```

3. `src/sitemap.xml`:

```xml
  <url><loc>http://localhost/</loc></url>
</urlset>
```

`src/robots.txt` can be `User-agent: *`, `src/assets/data.xml` any XML, `src/assets/notes.txt` any text, `src/assets/styles.css` any CSS, `src/assets/data.json` any JSON. `src/pages/index.html` any HTML.

4. Build, and confirm the files are in the output:

```sh
greenwood build
ls public/sitemap.xml public/robots.txt public/assets/data.xml public/assets/notes.txt
```

The build log also reports them:

```
copying directory... src/assets/
copying file... src/robots.txt
copying file... src/sitemap.xml
```

5. Serve and request them:

```sh
greenwood serve

curl -i --path-as-is http://localhost:8391/sitemap.xml         # 404 Not Found
curl -i --path-as-is http://localhost:8391/robots.txt          # 404 Not Found
curl -i --path-as-is http://localhost:8391/assets/data.xml     # 404 Not Found
curl -i --path-as-is http://localhost:8391/assets/notes.txt    # 404 Not Found
curl -i --path-as-is http://localhost:8391/assets/styles.css   # 200 OK, text/css
curl -i --path-as-is http://localhost:8391/assets/data.json    # 200 OK, application/json
```

6. To exercise the other server path, add `src/pages/api/greeting.js` to the same project so `serve` selects `getHybridServer`, then rebuild and repeat step 5. `/api/greeting` returns 200 and the four 404s above are unchanged.

### Environment

- `ProjectEvergreen/greenwood` `master` at commit `d01e4042` ("fix(cli): #1707 skip over unsupported api files instead of aborting the build (#1708)")
- Node v22.20.0
- Linux arm64

### Additional Context

Root cause is in `getStaticServer`, `packages/cli/src/lifecycles/serve.js:223-281`.

- `serve.js:239` — `checkResourceExists(url)` passes, so the request enters the "file exists" branch.
- `serve.js:240-244` — the candidate plugins are narrowed to `plugin.isStandardStaticResource`. That flag is set in `packages/cli/src/lifecycles/config.js:36-38`: a plugin qualifies only if its name starts with `plugin-standard` (excluding `plugin-standard-html`) or is `plugin-source-maps`. The resulting extension coverage is `js`, `json`, `css`, `avif/webp/jpg/jpeg/png/gif/svg/ico`, `woff2/woff/ttf/eot`, the audio and video lists, and `wasm`. There is no `xml`, no `txt`, and no generic fallback.
- `serve.js:249-252` — `initResponse` is constructed from the current Koa context, i.e. `status: ctx.response.status`, which is still Koa's default `404` at this point in the chain, with a null body.
- `serve.js:255-259` — no plugin's `shouldServe` returns true for `.xml`/`.txt`, so `response` is never reassigned and remains that 404 `initResponse`.
- `serve.js:261` — `if (response.ok)` is therefore false, so `ctx.body`, `ctx.status`, `ctx.message` and the response headers are all left unset. The middleware simply falls through with nothing assigned. The subsequent middleware do not pick it up either: the proxy middleware at `serve.js:285` only runs when `devServer.proxy` is configured, and the page/SPA middleware at `serve.js:318-354` only acts when the path matches a route in the graph, is a SPA path, or ends in `.html`. Koa then emits its own default 404.

A detail that pins the path taken: the explicit not-found branch at `serve.js:269-274` sets `Content-Type: text/plain` with no charset, whereas a request for an existing-but-unclaimed file comes back as `text/plain; charset=utf-8` — Koa's default. Requesting a genuinely missing `/definitely-missing.xml` on the same server returns `Content-Type: text/plain` (no charset), confirming the two 404s originate from different places and that the existing file is going through the silent fall-through at `:261`.

So the outcome is a plain 404 — not an empty 200 and not a wrong `Content-Type`. The file is never read.

An observation on the scope of this, since it applies to anything Greenwood itself generates: `packages/cli/src/plugins/copy/plugin-copy-sitemap.js` copies `src/sitemap.xml` to `public/sitemap.xml` and `plugin-copy-robots.js` copies `robots.txt`, and both land in the output directory only to be 404'd by `serve`. Any output-directory meta file whose extension is not in a standard plugin's list is unreachable through the production server the same way, whether it is copied in or generated. That includes the static `public/sitemap.xml` snapshot that #1747 writes, and would include an RSS feed from #1752, both of which are `.xml`. Real static hosting serves these files, so the gap shows up when someone smoke-tests a production build locally.

jstockdi:
## Related Issue

Resolves #1767

## Documentation

No user facing documentation changes. This restores the behaviour the docs already imply for `greenwood serve` — that the production server serves what is in the output directory — so nothing in the docs needs to change. Notably `plugin-copy-robots` and `plugin-copy-sitemap` are default plugins that put `robots.txt` and `sitemap.xml` into the output directory, and until now `serve` would 404 both of them.

## Summary of Changes

`getStaticServer` in `packages/cli/src/lifecycles/serve.js` confirms the requested file exists in the output directory, then narrows the Greenwood default plugins to the ones flagged `isStandardStaticResource` and asks each one's `shouldServe`. Those plugins cover `js`, `json`, `css`, images, fonts, audio, video, `wasm` and `.map`. Nothing covers `.xml` or `.txt`, and there was no fallback, so `response` stayed the `initResponse` built from `ctx.response.status` — still Koa's default 404 at that point in the chain. `if (response.ok)` was therefore false, nothing was assigned to `ctx`, and Koa emitted its own default 404 (`text/plain; charset=utf-8`, body `Not Found`).

The result is that a file which is demonstrably present in `public/` is unreachable through the production server. `getHybridServer` composes `getStaticServer`, so both server paths behave identically.

### The fix

After the plugin loop, if and only if no plugin claimed the request (`response === initResponse`), look the extension up in a small fixed map and serve the file with that `Content-Type`:

```js
const STATIC_FALLBACK_CONTENT_TYPES = {
  csv: "text/csv",
  txt: "text/plain",
  webmanifest: "application/manifest+json",
  xml: "application/xml",
};
```

An extension that is not in the map is left exactly as it is today and still falls through to a 404.

### Why an allowlist rather than "serve any file that exists"

The obvious fix is to read and return whatever is on disk once `checkResourceExists` passes, and I deliberately did not do that. It would broaden what the production server hands out to every file that has ever landed in the output directory, including anything a `copy` plugin, an adapter, or a user's own build step drops there, and it would do so silently with no `Content-Type` the server can reason about. That is a much bigger behavioural change than the issue calls for, and it is the sort of change that is hard to walk back once people depend on it.

The allowlist keeps the change proportional. It is a closed set of text/data extensions that a static host would serve out of a web root anyway, chosen around what Greenwood itself emits (`sitemap.xml` and `robots.txt` via the default copy plugins) plus the two obvious neighbours (`.webmanifest`, `.csv`). Adding a further extension later is a one line, reviewable diff. Deliberately not included: `.html` (already handled by the page/SPA middleware further down the chain) and anything executable or binary.

This is contained entirely within `getStaticServer` and introduces no new resource plugin and no change to the plugin contract or to the `isStandardStaticResource` flag.

### This does not weaken the path traversal hardening from #1749

The fallback is nested inside the existing guard and runs only after both of its checks have already passed:

```js
const isWithinOutputDir = url.href.startsWith(outputDir.href);

if (isWithinOutputDir && (await checkResourceExists(url))) {
  // ... plugin loop, then the new fallback
}
```

The `isWithinOutputDir` boundary check added in #1749 is untouched, and the fallback cannot be reached for any URL that check rejects — a `../` or percent-encoded traversal never enters this branch at all, so it never reaches the extension lookup. The fallback also narrows further rather than widening: it only fires for a path that is inside the output directory, exists on disk, and whose extension is in the fixed map. The regression tests in `serve.default.path-traversal` still pass unchanged.

## Testing

Folded into the existing `serve.default` case rather than adding a new test directory, alongside the other per-content-type assertions there. Fixtures added: `src/sitemap.xml`, `src/robots.txt`, `src/assets/data.xml`.

New assertions cover status, `Content-Type` and body for:

- `/sitemap.xml` — 200, `application/xml`
- `/robots.txt` — 200, `text/plain`
- `/assets/data.xml` — 200, `application/xml` (so this is not special casing two filenames)

Plus a negative assertion that a genuinely missing `/assets/definitely-missing.xml` still returns 404 with body `Not Found` and `Content-Type: text/plain`, so the fix cannot be mistaken for "serve everything on disk". That negative already passed before the fix and still passes after it.

On unpatched `master` the case reports 46 passing / 9 failing. With the fix it is 55 passing / 0 failing.

`serve.default.path-traversal`, `serve.config.base-path`, `serve.default.api`, `serve.default.error`, `serve.spa` and `serve.config.static-router` were also run and are unaffected.

## Note on #1747

Not addressed here and nothing related to it is touched, but worth flagging for whoever reviews it: #1747 (dynamic sitemap) depends on `serve` returning the built `public/sitemap.xml` snapshot, and this gap is why that serve path is currently non-functional for local production smoke testing. #1747's own "Known caveats" already discloses it. This change removes that blocker without any coupling to #1747 — the same applies to an RSS feed or any other `.xml` artifact a future feature writes into the output directory.


packages/cli/src/lifecycles/serve.js
```diff
@@ -220,6 +220,14 @@ async function getDevServer(compilation) {
   return app;
 }
 
+// extensions no standard resource plugin claims, but which static hosts still serve
+const STATIC_FALLBACK_CONTENT_TYPES = new Map([
```
thescientist13, line 224: hmm, I feel like if we were going to go down this route, then maybe it should be managed through a generic plugin?  I do wonder how we curate this list though, why only these formats?  Or are they listed in MDN somehow as "standard web formats"?  I feel like csv in particular feels to specialized?

Or can we just default to some default content type for unknown static assets?

----

I feel like at minimum, _.xml_ should be a standard plugin for sure?

PR #1772, opened 2026-08-08
enhancement(cli): #1769 warn on unsupported top level greenwood config options

### LINKED ISSUE #1769: Unknown or misspelled top-level keys in greenwood.config.js are silently ignored
### What happened

A top-level key in `greenwood.config.js` that Greenwood does not recognize is dropped with no diagnostic. The build succeeds with exit code 0 and no warning, and the setting the user intended simply never applies.

This is most costly for near-miss typos of real options, because the failure mode is indistinguishable from the option not working. `basepath: "/my-app"` (lowercase `p`) builds cleanly and produces a site with `basePath: ""`. `prerendor: true` builds cleanly and produces a site with `prerender: false`.

A misspelled key and a wholly invented key are treated identically: both produce byte-identical build output to a config with no such key at all. There is no signal anywhere that the key was seen and discarded.

This contrasts with how the rest of config handling behaves. Keys Greenwood *does* recognize are validated fairly strictly, and a bad value is a hard rejection (`Configuration error: basePath must be a string...`). So a wrong *value* stops the build, but a wrong *key name* is silent. From the user's side, the stricter case is the one that is easier to debug.

### Steps to reproduce

In an otherwise empty project with a single page at `src/pages/index.html`:

`greenwood.config.js`

```js
export default {
  basepath: "/my-app",
  prerendor: true,
  notARealOption: 42,
};
```

Run `greenwood build`. Observed output:

```
Welcome to Greenwood (v0.34.0) ♻️
Running Greenwood with the build command.
Initializing project config
Generating graph of workspace files...
building from local sources...
pages to generate
 /
generated page... /
success, done generating all pages!
bundling static assets...
optimizing static pages....
copying file... .greenwood/manifest.json
```

Exit code: `0`. No warning on stdout or stderr.

Inspecting the merged configuration confirms nothing was applied:

```
merged config keys: activeContent, basePath, devServer, isolation, layoutsDirectory,
optimization, pagesDirectory, plugins, polyfills, port, prerender, useTsc, workspace
has notARealOption: false | has basepath: false | has prerendor: false
basePath: ""      (expected "/my-app" had the key been spelled basePath)
prerender: false  (expected true had the key been spelled prerender)
```

Running the same two cases separately - `{ basepath: "/my-app" }` alone, then `{ notARealOption: 42 }` alone - produces identical console output and identical exit codes, so a misspelling of a real option is not distinguishable from an unknown key.

Expected: some indication that a supplied key was not recognized, so that a typo is a quick fix rather than a hunt through the build for why an option "does not work".

### Environment

- Greenwood `master` at commit `d01e404267d628221eeeef95bab641c29f131c2f` (v0.34.0)
- Node 22.20.0
- Linux arm64

### Additional Context

Root cause is in `packages/cli/src/lifecycles/config.js:106-121`. The user config module is destructured into a fixed list of recognized names:

```js
const userCfgFile = (await import(configUrl)).default;
const {
  workspace,
  devServer,
  optimization,
  plugins,
  port,
  prerender,
  basePath,
  staticRouter,
  pagesDirectory,
  layoutsDirectory,
  activeContent,
  isolation,
  polyfills,
  useTsc,
} = userCfgFile;
```

Only those bindings are consulted from that point on. `userCfgFile` is not referenced again, and the return at `config.js:349` is `{ ...defaultConfig, ...customConfig }` - built entirely from the destructured names - so any other key on the user's config object is unreachable for the rest of the run.

To be precise about what validation does and does not exist today, on current master:

Recognized and validated, with a rejection on a bad value:

- `workspace` - must be a `URL` and must exist (`config.js:124-136`)
- `optimization` - must be one of `default`, `none`, `static`, `inline` (`config.js:138-147`)
- `activeContent` - must be a boolean (`config.js:149-154`)
- `plugins` - each entry must have a known `type`, a function `provider`, and a string `name` (`config.js:156-181`)
- `devServer.hud` - must be a boolean (`config.js:201-209`)
- `devServer.port` - must be an integer (`config.js:211-219`)
- `devServer.extensions` - must be an array (`config.js:225-233`)
- `port` - must be an integer (`config.js:236-244`)
- `basePath` - must be a string (`config.js:246-254`)
- `pagesDirectory` - must be a string (`config.js:256-262`)
- `layoutsDirectory` - must be a string (`config.js:264-270`)
- `prerender` - must be a boolean (`config.js:272-280`)
- `isolation` - must be a boolean (`config.js:287-295`)
- `staticRouter` - must be a boolean (`config.js:297-305`)
- `polyfills.importMaps` - must be a boolean (`config.js:312-320`)
- `polyfills.importAttributes` - must be an array (`config.js:322-330`)
- `useTsc` - must be a boolean (`config.js:333-341`)

Accepted without validation:

- `devServer.proxy` - assigned as-is (`config.js:221-223`)

Not checked at all, at any level:

- any top-level key outside the fourteen destructured names
- any key inside `devServer` outside `hud`, `port`, `proxy`, `extensions`
- any key inside `polyfills` outside `importMaps`, `importAttributes`

One point worth raising, in case unknown keys are tolerated deliberately: they are not reachable by anything downstream. Extra keys never enter the object returned by `readAndMergeConfig`, so plugins and adapters cannot read them either. The exported `Config` type at `packages/cli/src/types/config.d.ts:4-29` is likewise a closed object type with no index signature, listing exactly the same set of names, which reads as the config surface being intended as a known, closed set rather than an open bag. If that reading is wrong and permissiveness here is intentional, that would be useful to know and would make this a documentation matter instead.

There is existing precedent in this file for a non-fatal configuration diagnostic rather than a rejection - `config.js:191-193` emits `Configuration warning: more than one custom renderer plugin detected.` - so a warning channel already exists here; I am not proposing a particular shape for one, just noting it.

Related but distinct: #1745 covers plugin config validation errors surfacing as an unhandled rejection. That concerns the control flow of the `plugins` loop at `config.js:156-181` and does not touch the destructuring above it; the behavior reported here is present both before and after that change.

jstockdi:
## Related Issue

Resolves #1769

## Documentation

No documentation changes. The set of supported options is already documented at [Configuration](https://greenwoodjs.dev/docs/reference/configuration/) and is unchanged by this PR.

## Summary of Changes

An unsupported top level key in `greenwood.config.js` is currently dropped without any diagnostic. The user config is destructured into a fixed list of names in `readAndMergeConfig`, and the merged config is built only from those bindings, so anything else is silently discarded. A near miss typo of a real option (`basepath` for `basePath`) is indistinguishable from an option simply not working.

This adds a check right after that destructure which warns once per unrecognized top level key:

```
Configuration warning: "basepath" is not a supported configuration option and will be ignored.
```

**It warns, it does not throw.** Rejecting an unknown key would be a breaking change: any project that currently carries a stray or forward compatible key builds fine today and would start failing. That call belongs to you, not to this fix, so this only makes the discard visible. It follows the existing non-fatal precedent in this same file, the "more than one custom renderer plugin detected" warning.

Scope is deliberately **top level only**. Unknown keys nested inside `devServer` and `polyfills` are dropped the same way, but those sub-objects are looser (`devServer.proxy` is already an unvalidated pass through bag), so warning there has a higher chance of false positives. Happy to extend to the nested scopes in a follow up if you want it.

The list of supported options is derived from `defaultConfig` rather than being a second hand maintained list, so it stays in sync automatically. `staticRouter` is appended because it is the one supported option with no default value.

### Testing

New test case `packages/cli/test/cases/build.config.unknown-keys`. It builds with a config containing a misspelled real option (`basepath`), a wholly unknown option (`notARealOption`), and a supported option (`staticRouter`), then asserts warnings for the first two and no warning for the third, while the build still succeeds. On master the two warning assertions fail; with this change all pass.

I did not fold this into an existing config case because none of them exercise a successful build with unsupported keys, and the point of the test is that the build still exits cleanly. Folding it into one of the `build.config.error-*` cases would not have covered the "warn, do not throw" behavior.

### Overlap with #1746

#1746 (approved, unmerged) also touches `packages/cli/src/lifecycles/config.js`, swapping `forEach` for `for...of` in the plugins loop at lines 156-181. This change sits at lines 78-80 and 122-130, so the hunks do not collide and the two should merge in either order.

### DIFF
--- a/packages/cli/src/lifecycles/config.js
+++ b/packages/cli/src/lifecycles/config.js
@@ -78,6 +78,8 @@ const defaultConfig = {
   useTsc: false,
   workspace: new URL("./src/", cwd),
 };
+// staticRouter has no default value of its own, so it is not part of defaultConfig
+const supportedConfigOptions = [...Object.keys(defaultConfig), "staticRouter"];
 
 const readAndMergeConfig = async () => {
   // check for greenwood.config.ts or greenwood.config.js
@@ -119,6 +121,14 @@ const readAndMergeConfig = async () => {
       polyfills,
       useTsc,
     } = userCfgFile;
+
+    for (const key of Object.keys(userCfgFile)) {
+      if (!supportedConfigOptions.includes(key)) {
+        console.warn(
+          `Configuration warning: "${key}" is not a supported configuration option and will be ignored.`,
+        );
+      }
+    }
 
     // workspace validation
     if (workspace) {
--- a/packages/cli/test/cases/build.config.unknown-keys/build.config.unknown-keys.spec.js
+++ b/packages/cli/test/cases/build.config.unknown-keys/build.config.unknown-keys.spec.js
@@ -1,0 +1,80 @@
+/*
+ * Use Case
+ * Run Greenwood build command with unsupported top level options in a custom config.
+ *
+ * User Result
+ * Should complete the build and warn about each unsupported option.
+ *
+ * User Command
+ * greenwood build
+ *
+ * User Config
+ * {
+ *   basepath: "/my-app",
+ *   notARealOption: 42,
+ *   staticRouter: false
+ * }
+ *
+ * User Workspace
+ * src/
+ *   pages/
+ *     index.html
+ */
+import { expect } from "chai";
+import path from "node:path";
+import { getOutputTeardownFiles } from "../../../../../test/utils.js";
+import { Runner } from "gallinago";
+import { runSmokeTest } from "../../../../../test/smoke-test.js";
+import { fileURLToPath } from "node:url";
+
+describe("Build Greenwood With: ", function () {
+  const LABEL = "Custom Configuration with unsupported options";
+  const cliPath = path.join(process.cwd(), "packages/cli/src/bin.js");
+  const outputPath = fileURLToPath(new URL(".", import.meta.url));
+  let runner;
+  let stderr = "";
+
+  before(function () {
+    this.context = {
+      publicDir: path.join(outputPath, "public"),
+    };
+    runner = new Runner();
+  });
+
+  describe(LABEL, function () {
+    before(async function () {
+      await runner.setup(outputPath);
+
+      const build = runner.runCommand(cliPath, "build");
+
+      // the runner only surfaces stderr for a non zero exit code, and this build is expected to pass
+      runner.childProcess.stderr.on("data", function (data) {
+        stderr += data.toString();
+      });
+
+      await build;
+    });
+
+    runSmokeTest(["public", "index"], LABEL);
+
+    it("should warn about a misspelling of a supported option", function () {
+      expect(stderr).to.contain(
+        'Configuration warning: "basepath" is not a supported configuration option and will be ignored.',
+      );
+    });
+
+    it("should warn about an option Greenwood does not know about at all", function () {
+      expect(stderr).to.contain(
+        'Configuration warning: "notARealOption" is not a supported configuration option and will be ignored.',
+      );
+    });
+
+    it("should not warn about a supported option that has no default value", function () {
+      expect(stderr).to.not.contain('Configuration warning: "staticRouter"');
+    });
+  });
+
+  after(async function () {
+    await runner.teardown(getOutputTeardownFiles(outputPath));
+  });
+});
--- a/packages/cli/test/cases/build.config.unknown-keys/greenwood.config.js
+++ b/packages/cli/test/cases/build.config.unknown-keys/greenwood.config.js
@@ -1,0 +1,5 @@
+export default {
+  basepath: "/my-app",
+  notARealOption: 42,
+  staticRouter: false,
+};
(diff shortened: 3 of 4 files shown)


packages/cli/src/lifecycles/config.js
```diff
@@ -78,6 +78,8 @@ const defaultConfig = {
   useTsc: false,
   workspace: new URL("./src/", cwd),
 };
+// staticRouter has no default value of its own, so it is not part of defaultConfig
+const supportedConfigOptions = [...Object.keys(defaultConfig), "staticRouter"];
```
thescientist13, line 82: > staticRouter has no default value of its own

hmm...this just seems like an oversight?  It's documented
https://greenwoodjs.dev/docs/reference/configuration/#static-router

and captured in the type definition
https://github.com/ProjectEvergreen/greenwood/blob/master/packages/cli/src/types/config.d.ts#L26

Seems like we should just add `staticRouter: false` to the defaultConfig?

thescientist13:
Opted to classify this as an enhancement instead of fix, fyi (as I don't really consider anything to be broken here, this is just a nice bit of feedback for users)

Issue #1770, opened 2026-08-08
develop passes an empty page object to getBody for SSR pages without a dynamic segment


jstockdi:
### What happened?

For an SSR page whose route has no dynamic segment, `greenwood develop` calls the page's `getBody(compilation, page, request, params)` with `page` set to an empty object. `greenwood build` + `greenwood serve` calls the same `getBody` with the fully populated page node from the graph. Nothing errors in either mode, so every property a page reads off `page` is simply `undefined` in development and correct in production.

With a single page `src/pages/ssr.js` whose `getBody` serialises whatever `page` it is handed, the same route returns:

`greenwood develop`, `curl http://localhost:1995/ssr/`:

```
<!-- greenwood-ssr-start --><pre>{}</pre><!-- greenwood-ssr-end -->
```

`greenwood build && greenwood serve`, `curl http://localhost:8095/ssr/`:

```
<!-- greenwood-ssr-start --><pre>{
  "id": "ssr",
  "label": "Ssr",
  "title": "Ssr",
  "route": "/ssr/",
  "layout": "page",
  "data": {},
  "pageHref": "file:///path/to/project/src/pages/ssr.js",
  "outputHref": "file:///path/to/project/public/ssr/index.html",
  "isSSR": true,
  "prerender": null,
  "isolation": false,
  "hydration": false,
  "servePage": "dynamic",
  "segment": null,
  "staticPaths": null,
  "hasStaticParams": false
}</pre><!-- greenwood-ssr-end -->
```

So in `develop` the page object has zero own properties. In `serve` it has sixteen: `data`, `hasStaticParams`, `hydration`, `id`, `isSSR`, `isolation`, `label`, `layout`, `outputHref`, `pageHref`, `prerender`, `route`, `segment`, `servePage`, `staticPaths`, `title`. Every one of those is present in production and absent in development.

The practical effect is that a page reading `page.title`, `page.route`, `page.data`, `page.id` etc. renders differently in the two modes without any diagnostic. Adding a `getFrontmatter` export to the same page and having `getBody` return `` `title=${page.title}` `` gives `title=undefined` under `develop` and `title=My SSR Page` under `serve`.

This is scoped to `getBody`. `getLayout` on the same page receives the populated node in `develop` as well (it is dispatched separately from `packages/cli/src/lib/layout-utils.js:339` with `page: JSON.stringify(matchingRoute)`), so only the body call diverges.

[...]

Issue #1278, opened 2024-09-05
support hot reloading content as data changes


thescientist13:
## Summary

Currently during development, Greenwood builds the list of all pages (the graph) including reading all frontmatter, and creates a data structure that it holds in memory, as well as outputting a _graph.json_ file that can be read / fetched from userland.

However, this data / file never gets updated again, so if you change the _frontmatter_ of a page, or add a new page, the content as data in your application will never change.  Instead, you'll have to stop / start the dev server again to generate new data based on the new content.

## Details

So naturally, we'll need to do a couple things to achieve this:
1. Be able to detect a content change (e.g. markdown / HTML) or at the very least tap into the livereload lifecycle and always generate new content as data
1. Output a new _graph.json_
1. Be able to globally update `compilation.graph` in-memory data structure
1. For GraphQL, would we have to hot-reload the entire server, or can we just have those resolvers resolve the content server, instead of _graph.json_

PR #1773, opened 2026-08-08
fix(cli): #1766 return a 404 when a develop page file is deleted while the server runs

### LINKED ISSUE #1766: develop: deleting a page while the dev server is running returns a 500 with a raw ENOENT instead of a 404
### What happened?

Deleting a page file while `greenwood develop` is running makes that route return **500** with a raw `ENOENT` stack in the terminal, instead of the **404** the route now deserves.

The page graph is generated once, at startup, and every request is resolved against that in-memory graph. When a page file is removed, the stale graph entry still matches the route, so the dev server proceeds to read a file that no longer exists; the resulting `ENOENT` is caught by the generic serve handler and reported as a server error. From the outside it looks like Greenwood broke, rather than the page simply not being there any more.

Two related notes, so the scope is clear:

- Editing an existing page during `develop` works fine — page contents are read from disk per request.
- Adding a *new* page during `develop` 404s until restart. That half is the same stale-graph cause and is already covered by the hot-reloading work in #1278, so this report is about the incorrect status code / unhandled read error on delete, which is a defect regardless of whether the graph is ever refreshed.

### Steps to reproduce

1. Create a minimal project:

   ```sh
   mkdir -p my-app/src/pages && cd my-app
   echo '{"type":"module"}' > package.json
   echo 'export default { devServer: { port: 1998 } };' > greenwood.config.js
   echo '<h1>Home</h1>' > src/pages/index.html
   echo '<h1>About</h1>' > src/pages/about.html
   ```

2. Start the dev server:

   ```sh
   greenwood develop
   ```

3. Confirm both routes serve:

   ```sh
   curl -s -o /dev/null -w '%{http_code}\n' http://localhost:1998/        # 200
   curl -s -o /dev/null -w '%{http_code}\n' http://localhost:1998/about/  # 200
   ```

4. With the server still running, delete the page:

   ```sh
   rm src/pages/about.html
   ```

5. Request the deleted route:

   ```sh
   curl -s -o /dev/null -w '%{http_code}\n' http://localhost:1998/about/
   ```

   Actual: `500`, with an empty body. Expected: `404`.

   The dev server terminal prints:

   ```
   Error: ENOENT: no such file or directory, open '/path/to/my-app/src/pages/about.html'
       at async open (node:internal/fs/promises:641:25)
       at async Object.readFile (node:internal/fs/promises:1245:14)
       at async StandardHtmlResource.serve (file:///.../packages/cli/src/plugins/resource/plugin-standard-html.js:55:14)
       at async file:///.../packages/cli/src/lifecycles/serve.js:72:27
       at async file:///.../packages/cli/src/lifecycles/serve.js:54:5 {
     errno: -2,
     code: 'ENOENT',
     syscall: 'open',
     path: '/path/to/my-app/src/pages/about.html'
   }
   ```

6. Restart `greenwood develop` and request `/about/` again — now correctly `404`.

### Environment

- Greenwood: `master` at commit `d01e4042` (`packages/cli` version 0.34.0)
- Node: 22.20.0
- OS: Linux (arm64)

### Additional Context

Likely cause, for whatever it is worth: the graph is built once at `packages/cli/src/lifecycles/compile.js:113` (`generateGraph`), and both `shouldServe` and `serve` in the standard HTML resource plugin match against that frozen `compilation.graph` — `packages/cli/src/plugins/resource/plugin-standard-html.js:24` and `:41`. Because the deleted page's node is still present, `shouldServe` returns true and `serve` reaches the unguarded read at `plugin-standard-html.js:55`; the thrown `ENOENT` is then caught by the catch-all at `packages/cli/src/lifecycles/serve.js:85-88`, which sets a 500 and logs the error object.

The only file watcher registered in develop mode is livereload (`packages/cli/src/plugins/server/plugin-livereload.js:66`), which watches the workspace but only pushes browser refreshes — so a delete inside `src/` does trigger a reload, which lands the browser straight on the 500.

Happy to work on a patch once you have a view on what the right behaviour is here.

jstockdi:
## Related Issue

Resolves #1766

## Documentation

No documentation changes. This corrects an HTTP status code in `greenwood develop`; the documented behaviour is unchanged.

## Summary of Changes

Deleting a page file while the dev server is running returned a **500 with an empty body**, plus a raw `ENOENT` stack in the terminal, where a **404** is correct.

The page graph is generated once at startup, so a deleted page's node is still present. `shouldServe` in the standard HTML resource plugin still matches the route, `serve` then reaches an unguarded `fs.readFile`, and the resulting `ENOENT` is swallowed by the catch-all in `lifecycles/serve.js`, which sets a 500. Restarting `develop` makes the same route correctly 404.

The fix guards that one read: in `develop`, an `ENOENT` returns a `404` response; any other error still throws so real failures are not masked.

The `develop` gate is load-bearing rather than cosmetic. `lifecycles/prerender.js` calls `plugin.serve()` on every resource plugin and does not check `response.ok`, so an ungated 404 here is swallowed during `greenwood build`: a page deleted between graph generation and prerender produced exit `0`, a `generated page...` log line, and an **empty** `public/<route>/index.html`. Greenwood's convention for a missing page file at build time is a hard failure (`lifecycles/graph.js:179`), and the stale-graph condition that justifies a 404 only exists in `develop`, so the guard is scoped to match.

### Scope

Deliberately narrow. This does **not** refresh the graph or add a file watcher — hot reloading of the graph (which is also what makes a *newly added* page 404 until restart) is tracked in #1278 and is a feature. This is just the incorrect status code and unhandled read error on delete, which is wrong regardless of whether the graph ever refreshes.

It also does **not** change `greenwood build` or `greenwood serve`. `build` still fails loudly on a missing page file, verified by test; `serve` never reaches this plugin at all, since `lifecycles/config.js:36-38` excludes `plugin-standard-html` from `isStandardStaticResource`.

### Testing

Folded into the existing `develop.default` case rather than a new directory. A `fixtures/deleted.html` is copied to `src/pages/deleted.html` via gallinago's `setupFiles` before the server starts (so it is in the graph), the test asserts it serves `200`, deletes it while the server is still running, and asserts the same route now returns `404`. Teardown removes the copy if a failure leaves it behind.

There is no timing element — `fs.rmSync` completes before the second `fetch` — so the result is deterministic. Each of the counts below is from 5 consecutive runs of `packages/cli/test/cases/develop.default`:

| | result |
| --- | --- |
| before this change (test only) | 117 passing, 1 failing — 5/5 runs, always `AssertionError: expected 500 to equal 404` |
| with the fix | 118 passing, 0 failing — 5/5 runs |

Baseline before any change was 115 passing; the new describe adds 3 tests.

Wider runs, all green with the fix:

- all `develop.*` cases: 313 passing
- `build.default`, `build.default.pages*`, `serve.default`: 48 passing

`prettier --check` and `eslint` are clean on the changed files, as is `ls-lint`.

### Note for the maintainer

The fix for #1770 lands in this same file, `packages/cli/src/plugins/resource/plugin-standard-html.js`, but at the `getMatchingDynamicSsrRoute(...) || {}` expression around line 42 — well clear of this hunk around line 56, so the two should merge without conflict.

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-standard-html.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-html.js
@@ -56,8 +56,8 @@ class StandardHtmlResource {
       try {
         body = await fs.readFile(new URL(pageHref), "utf-8");
       } catch (e) {
-        // the graph is only generated once, so a matching route's page may have been deleted since then
-        if (e.code === "ENOENT") {
+        // in development the graph is only generated once, so a matching page may since be deleted
+        if (e.code === "ENOENT" && process.env.__GWD_COMMAND__ === "develop") {
           return new Response(null, { status: 404 });
         }
 
--- a/packages/cli/src/plugins/resource/plugin-standard-html.js
+++ b/packages/cli/src/plugins/resource/plugin-standard-html.js
@@ -53,7 +53,16 @@ class StandardHtmlResource {
     let html = "";
 
     if (isHtmlContent) {
-      body = await fs.readFile(new URL(pageHref), "utf-8");
+      try {
+        body = await fs.readFile(new URL(pageHref), "utf-8");
+      } catch (e) {
+        // the graph is only generated once, so a matching route's page may have been deleted since then
+        if (e.code === "ENOENT") {
+          return new Response(null, { status: 404 });
+        }
+
+        throw e;
+      }
 
       // purge frontmatter data from HTML files that only have a `<body>` tag
       // frontmatter outside of an <html> tag will get ignored by node-html-parser
--- a/packages/cli/test/cases/develop.default/develop.default.spec.js
+++ b/packages/cli/test/cases/develop.default/develop.default.spec.js
@@ -16,6 +16,8 @@
  * }
  *
  * User Workspace
+ * fixtures/
+ *   deleted.html (copied in as src/pages/deleted.html, then deleted mid test run)
  * src/
  *   assets/
  *     data.json
@@ -66,6 +68,7 @@ describe("Develop Greenwood With: ", function () {
   const outputPath = fileURLToPath(new URL(".", import.meta.url));
   const hostname = "http://localhost";
   const port = 1984;
+  const deletedPagePath = path.join(outputPath, "src/pages/deleted.html");
   let runner;
 
   before(function () {
@@ -77,7 +80,12 @@ describe("Develop Greenwood With: ", function () {
 
   describe(LABEL, function () {
     before(async function () {
-      await runner.setup(outputPath);
+      await runner.setup(outputPath, [
+        {
+          source: path.join(outputPath, "fixtures/deleted.html"),
+          destination: deletedPagePath,
+        },
+      ]);
 
       await new Promise((resolve, reject) => {
         runner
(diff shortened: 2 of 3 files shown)


packages/cli/src/plugins/resource/plugin-standard-html.js
```diff
@@ -53,7 +53,16 @@ class StandardHtmlResource {
     let html = "";
 
     if (isHtmlContent) {
-      body = await fs.readFile(new URL(pageHref), "utf-8");
+      try {
+        body = await fs.readFile(new URL(pageHref), "utf-8");
+      } catch (e) {
+        // in development the graph is only generated once, so a matching page may since be deleted
```
thescientist13, line 59: based on https://github.com/ProjectEvergreen/greenwood/issues/1766#issuecomment-5433499897, maybe we leave a comment here with a link to that follow up issue?

PR #1774, opened 2026-08-08
fix(cli): #1768 warn about and skip get static paths values that collide with an existing page

### LINKED ISSUE #1768: getStaticPaths value colliding with an existing static page is silently discarded and the output file is written twice
### What happened?

When a value returned from `getStaticPaths` produces the same route as an existing static page, the dynamic page is silently dropped. The static page wins, the dynamic route's `getBody` is never called, and the same output file is rendered and written twice. The build logs both pages as generated, prints no warning, and exits 0.

With `src/pages/about.html` and `src/pages/[slug].js` whose `getStaticPaths` returns `{ params: { slug: "about" } }`, `greenwood build` prints:

```
pages to generate
 /[slug]/
 /about/
 /
generated page... /
generated static path... /about/
generated page... /about/
success, done generating all pages!
```

Both `/about/` lines refer to the same output file, `public/about/index.html`. That file contains the contents of `about.html`. Adding a `console.log` to the dynamic route's `getBody` shows it is never invoked while `about.html` exists; delete `about.html` and the same `getBody` runs and its markup is emitted, so the collision is what suppresses it.

The outcome is deterministic, not a race: across 10 consecutive clean builds `public/about/index.html` was byte-identical every time (same md5), always the static page. The *order* of the two writes does vary between runs, but both writers currently produce the same bytes, so the ordering is not observable in the output today.

Nothing in the build reports the collision — no `console.warn`, no non-zero exit, nothing in `graph.json` marking the duplicate. A `getStaticPaths` value sourced from a CMS or a data file that happens to match a hand-authored page just disappears, with the site still building green.

Expected: the build surfaces the collision (a warning naming both pages at minimum), or documents which page takes precedence, rather than dropping one silently.

### Steps to reproduce

1. Create a minimal project:

   ```
   package.json          # { "type": "module" }
   src/
     pages/
       index.html
       about.html
       [slug].js
   ```

2. `src/pages/about.html`:

   ```html
   <html><head><title>about</title></head><body><h1>STATIC ABOUT PAGE</h1></body></html>
   ```

3. `src/pages/[slug].js`:

   ```js
     return [{ params: { slug: "about" } }];
   }

     return `<h1>DYNAMIC PAGE params=${JSON.stringify(params)}</h1>`;
   }
   ```

4. Run `greenwood build`.

5. Observe that the log contains both `generated static path... /about/` and `generated page... /about/`, that there is no warning, and that the exit code is 0.

6. `cat public/about/index.html` — it contains `STATIC ABOUT PAGE`. The dynamic route's body is nowhere in the output directory.

7. Optionally add `console.log("getBody called")` to `getBody` and rebuild: it never prints. Delete `src/pages/about.html` and rebuild: it prints, and `public/about/index.html` then contains `DYNAMIC PAGE`.

### Environment

- Greenwood version: `master` at commit `d01e4042` (v0.34.0)
- OS: Linux (arm64)
- Node version: 24.12.0
- Browser: N/A (build-time issue)

### Additional Context

Two separate things combine here.

1. Nothing validates `getStaticPaths` values against routes that already exist. `packages/cli/src/lifecycles/graph.js:209-210` attaches `staticPaths` to the dynamic page node, and `packages/cli/src/lib/url-utils.js:48-50` (`getStaticRouteFromDynamicRoute`) later expands each value into a concrete route, with no check that the resulting route is already occupied.

2. The winner is decided by route resolution, not by write order. When prerendering the static path, `packages/cli/src/plugins/resource/plugin-standard-html.js:41` resolves the request with `this.compilation.graph.find((node) => node.route === pathname)`, which matches the static `about.html` node (route `/about/`); the dynamic node's route is `/[slug]/` and never matches that predicate. `pageHref` therefore points at `about.html`, `isHtmlContent` is true at `:48`, and the file is read at `:55`. The SSR branch at `:77` (`matchingRoute.isSSR || matchingRouteWithSegment.isSSR`) is never reached, so the dynamic route module is not executed at all for that path.

Because of that, both prerender branches produce the static page's HTML and write it to the same scratch file: the `staticPaths` branch at `packages/cli/src/lifecycles/prerender.js:158` and the ordinary page branch at `:224`, both resolving to `.greenwood/about/index.html`. They run concurrently under the `asyncForEach` at `:85` (`Promise.all` via `packages/cli/src/lib/async-utils.js:19-31`), and instrumenting both call sites shows the two writes landing in either order across runs with identical content hashes.

The same duplication repeats in the optimize pass: `optimizeStaticPages` (`packages/cli/src/lifecycles/bundle.js:123`) reads the one scratch file twice and writes `public/about/index.html` twice, at `:162-165` and at `:195`.

So the page is rendered twice and written twice per build for no benefit, and the `getStaticPaths` value is silently inert. The double write is currently harmless only because both writers happen to carry the same bytes; if the resolution order at `plugin-standard-html.js:41` ever changed, the same code path would become an ordinary last-writer-wins race.

jstockdi:
## Related Issue

Resolves #1768

## Documentation

N / A

## Summary of Changes

When a value returned from `getStaticPaths` resolves to a route that an existing page already
generates (e.g. `src/pages/about.html` plus `src/pages/[slug].js` returning `slug: "about"`), the
static path is silently discarded and the same output file is rendered and written twice. The
dynamic route's `getBody` is never invoked, nothing is logged, and the build exits 0.

**Precedence is unchanged** — the existing page still wins, exactly as it does today, and the bytes
in the output directory are the same before and after this change. Only the silence and the
duplicate work are addressed. Deciding that a `getStaticPaths` value _should_ be able to override a
hand-authored page would be a behaviour change, so it is deliberately not made here.

1. [x] `lifecycles/graph.js` — once the graph is assembled, warn for every `getStaticPaths` value
       whose route is already generated by another page, naming the route and both source pages.
2. [x] `lifecycles/prerender.js` and `lifecycles/bundle.js` — skip those static paths instead of
       re-rendering and re-writing the other page's output. Both lifecycles process pages
       concurrently, so today the two writers target the same scratch path and the same public path
       at the same time; they happen to carry the same bytes, so nothing is corrupted, but the work
       is wasted and the build log claims to have generated a page that it did not.
3. [x] `lib/graph-utils.js` — small `getPageClaimingStaticRoute` helper shared by the three call
       sites. The check runs against the pages that actually emit a file, so a static path that
       overlaps a non-prerendered SSR route is still generated as before.
4. [x] Regression test in `packages/cli/test/cases/build.default.static-paths-route-collision/`,
       covering the warning, the skipped duplicate, the unchanged precedence of the existing page,
       and a second, non-colliding static path from the same dynamic route that must still be
       generated. Two of its four assertions fail on `master`.

Before:

```
pages to generate
 /[slug]/
 /about/
 /
generated page... /
generated static path... /about/
generated page... /about/
success, done generating all pages!
```

After:

```
pages to generate
 /[slug]/
 /about/
 /
generated page... /
generated page... /about/
success, done generating all pages!
```

plus, on stderr:

```
Duplicate route detected: the static path /about/ from pages/[slug].js is already generated by
pages/about.html, and will be skipped.  Please remove one of these pages, or change the value
returned by getStaticPaths.
```


thescientist13 (commented):
Looks like there a couple merge conflicts here, do you mind taking a look?

PR #1781, opened 2026-08-09
enhancement(plugins): #1780 warn on mismatched jsx inferred observability configuration

### LINKED ISSUE #1780: JSX Signals polyfill is gated on the plugin option alone, so a per-file inferredObservability export ships a page that throws Signal is not defined
### What happened

There are two independent switches that are both spelled `inferredObservability`, and nothing reconciles them:

1. the **per-file** `export const inferredObservability` in a `.jsx` / `.tsx` module, which WCC reads and which drives code generation (signals, `observedAttributes`, effect wiring), and
2. the **plugin option** `greenwoodPluginImportJsx({ inferredObservability: true })`, which only controls whether the `globalThis.Signal` polyfill `<script>` is injected into the HTML.

If a component opts in via the file-level export but the plugin is registered without options, the build succeeds and the prerendered HTML is correct — but the emitted bundle executes `new Signal.State(...)` / `new Signal.Computed(...)` against a global that was never defined, so **every page load throws `ReferenceError: Signal is not defined` and the component never upgrades**. There is no warning at build time.

The failure is silent: green build, correct-looking SSR markup, dead page in the browser.

### Steps to reproduce

`greenwood.config.js` — note the plugin is registered with no options:

```js
import { greenwoodPluginImportJsx } from "@greenwood/plugin-import-jsx";

export default {
  plugins: [greenwoodPluginImportJsx()],
  prerender: true,
};
```

`src/components/counter.jsx` (the counter from the plugin's own docs / test cases):

```jsx
export const inferredObservability = true;

export default class Counter extends HTMLElement {
  count;
  parity;

  constructor() {
    super();
    this.count = new Signal.State(0);
    this.parity = new Signal.Computed(() => (this.count.get() % 2 === 0 ? "even" : "odd"));
  }

  connectedCallback() {
    if (!this.shadowRoot) {
      this.attachShadow({ mode: "open" });
      this.render();
    }
  }

  render() {
    const { count, parity } = this;

    return (
      <div>
        <span class={parity.get()}>
          The count is {count.get()} ({parity.get()})
        </span>
      </div>
    );
  }
}

customElements.define("wcc-counter", Counter);
```

`src/pages/index.html`:

```html
<html>
  <head>
    <script type="module" src="../components/counter.jsx"></script>
  </head>
  <body>
    <wcc-counter></wcc-counter>
  </body>
</html>
```

Then:

```sh
greenwood build

grep -c "globalThis.Signal" public/index.html   # => 0
grep -o "new Signal.State(" public/counter.*.js  # => new Signal.State(
```

Observed: the build exits `0`, `public/index.html` contains the correct prerendered shadow DOM (`The count is 0 (even)`), zero `globalThis.Signal` scripts, and `public/counter.*.js` still calls `new Signal.State(...)`. Loading the page throws `ReferenceError: Signal is not defined`.

Expected: either the polyfill is injected, or the mismatch is reported before the browser finds it.

### Environment

- Greenwood `master` @ `37e68db3` (v0.35.0-alpha.0)
- `@greenwood/plugin-import-jsx`
- Node 24.12.0
- Linux (arm64)

### Additional Context

Root cause, all in `packages/plugin-import-jsx/src/index.js`:

- `:23` — `this.inferredObservability = options.inferredObservability;` stores the plugin option.
- `:36` — `serve()` calls `parseJsx(url)` and never passes the option down, so compilation is driven purely by the file-level export.
- `:49-53` — `shouldIntercept()` gates the polyfill injection purely on `this.inferredObservability`, so the polyfill is driven purely by the plugin option.

The per-file half is read on the WCC side in `wc-compiler/src/jsx-loader.js:411-414`. Neither side ever sees the other's value, so the two can disagree freely.

---

### Design question — which direction should this go?

I would like to settle the approach here before an implementation gets reviewed, since there are two reasonable and quite different answers.

**Option A — make the two paths consistent by auto-injecting.** When the plugin option is not set and a compiled JSX module opts in via the file-level export, inject the `Signal` polyfill anyway.

- Pro: the page actually works. The file-level export becomes sufficient on its own, which is what the docs' per-file flag reads like it should do.
- Con: it changes emitted output based on the content of a user module, and it is not straightforward to do correctly. The plugin only learns that a module opted in at the moment it *serves* that module. During `greenwood develop` the HTML document is served **before** the browser requests the `.jsx`, so an auto-inject keyed off compilation would miss the first page load and only start working after a refresh. Doing it order-independently means scanning the page graph for JSX modules at intercept time — noticeably more machinery than the bug warrants.
- Con: there is no precedent for a *plugin* auto-correcting a config mismatch. The two auto-derivation precedents in the codebase (`packages/cli/src/lifecycles/config.js:189-192`, where a custom renderer displaces the default one, and `:285-288`, where `prerender` is forced to `false` for an SPA) both live in core config for options core owns, and the second only fires when the user was silent.

**Option B — leave behaviour as-is, but surface the mismatch at build time.** Emit a configuration warning naming the offending module when a JSX module exports `inferredObservability` while the plugin option is not enabled.

- Pro: no change to emitted output, so nothing existing can regress.
- Pro: it matches the house style for relational misconfiguration. `packages/cli/src/lifecycles/config.js:194-197` already does exactly this shape for the "more than one custom renderer plugin" case — `console.warn("Configuration warning: ...")` followed by `console.debug(<offending values>)`, then carry on. `packages/cli/src/plugins/resource/plugin-node-modules.js:91-102` does the same for malformed dependency export conditions, and `packages/cli/src/lifecycles/graph.js:107` for unsupported API file extensions.
- Pro: hard-failing would be unprecedented. Every `Configuration error:` rejection in `config.js` is a type or enum check on a single option in isolation; there is no cross-option validation anywhere in the codebase, and no relational mismatch has ever been fatal.
- Con: the page is still broken; the developer has to act on the warning.

**I implemented Option B**, on the grounds that it is the conservative half — it is purely diagnostic, it works identically for `build` and `develop` because the check happens at exactly the point where both facts are known, and it follows an existing, near-identical precedent rather than inventing plugin-level auto-correction. Happy to swap to Option A if you'd rather the plugin just made the page work; the implementation is small either way.

Separately, and not addressed here: `wc-compiler/src/jsx-loader.js:413` uses `Boolean(node.declaration.declarations[0].init.raw)`, so `export const inferredObservability = false` also enables observability (`Boolean("false") === true`). That is a WCC-side issue.

jstockdi:
## Design question first — please redirect me if you'd rather go the other way

The linked issue has an open design choice and I'd rather settle it here than have you review the wrong implementation.

When a JSX module opts into signals through the per-file `export const inferredObservability` but the plugin is registered without its own `inferredObservability` option, Greenwood can either:

- **Option A — auto-inject.** Inject the `globalThis.Signal` polyfill anyway whenever a compiled module opted in, making the two switches consistent and the page actually work.
- **Option B — surface the mismatch.** Leave emitted output exactly as it is and report the mismatch at build time, so it is caught before the browser.

**This PR implements Option B.** Reasoning:

- It is the conservative half — purely diagnostic, no change to emitted output, so nothing existing can regress.
- It follows an existing, near-identical precedent. `packages/cli/src/lifecycles/config.js:194-197` handles the "more than one custom renderer plugin" misconfiguration with exactly this shape: `console.warn("Configuration warning: ...")`, then `console.debug(<offending values>)`, then carry on. `packages/cli/src/plugins/resource/plugin-node-modules.js:91-102` and `packages/cli/src/lifecycles/graph.js:107` do the same for other user-side misconfiguration. Hard-failing would be unprecedented — every `Configuration error:` rejection in `config.js` is a type or enum check on a single option in isolation, and no relational mismatch in the codebase has ever been fatal.
- The check sits in `serve()`, the one point where both facts are known, so it behaves identically for `build` and `develop`.

The main practical argument against Option A: the plugin only learns a module opted in when it *serves* that module, and during `greenwood develop` the HTML document is served **before** the browser requests the `.jsx` — so an auto-inject keyed off compilation would miss the first page load and only start working after a refresh. Making it order-independent means scanning the page graph for JSX modules at intercept time, which is meaningfully more machinery.

That said, Option A is the friendlier outcome for users and I have no strong attachment to B. **The change here is about a dozen lines and is straightforward to swap** — say the word and I'll rework it to auto-inject (and adjust the regression test to assert the polyfill script is present instead of asserting the warning).

## Related Issue

Resolves #1780

## Documentation

No documentation changes. Option B does not change any documented behaviour or add any configuration surface — it only reports an existing misconfiguration. If you prefer Option A, `packages/plugin-import-jsx/README.md` would want a note that the file-level export is sufficient on its own.

## Summary of Changes

`packages/plugin-import-jsx/src/index.js`

- `serve()` now inspects the tree it already gets back from `parseJsx` for a top-level `export const inferredObservability`. When that export is present and the plugin's own `inferredObservability` option is not enabled, it emits a `Configuration warning:` naming the offending module, following the `console.warn` + `console.debug` shape from `config.js`.
- Reported modules are tracked in a module-level `Set` because a resource plugin is instantiated per lifecycle, which otherwise repeats the same warning several times per build and on every refresh in `develop`.
- No change to what is emitted.

`packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/`

New loaders build case: the same counter component as the existing `loaders-build.prerender-inferred-observability` case (file-level `export const inferredObservability = true`), but with `greenwoodPluginImportJsx()` registered without options. It asserts the mismatch is reported, and pins the symptom it is reporting — the bundle contains `new Signal.State(` / `new Signal.Computed(` while the page contains no script defining `globalThis.Signal`.

### Fail-then-pass proof

Run with `node --import ./test/test-register.js ./node_modules/mocha/bin/mocha "./packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/*.spec.js"` (the `yarn test:loaders` harness, narrowed to the one case).

Before the change, against unmodified `packages/plugin-import-jsx/src/index.js`:

```
  Build Greenwood With:
    JSX based Signals Reactivity opted into per file while the plugin option is not enabled
      Running Smoke Tests: JSX based Signals Reactivity opted into per file while the plugin option is not enabled
        Public Directory Generated Output
          ✔ should create a public directory
          ✔ should output one graph.json file
          ✔ should not output any map files for HTML pages
      Build command output for the mismatched inferredObservability configuration
        1) should report the JSX module that opted into inferred observability
      Build command output for the emitted Signals bundle and page
        ✔ should emit a bundle that reads the Signal global
        ✔ should not contain a script tag for exposing Signal globally


  5 passing (3s)
  1 failing

  1) Build Greenwood With:
       JSX based Signals Reactivity opted into per file while the plugin option is not enabled
         Build command output for the mismatched inferredObservability configuration
           should report the JSX module that opted into inferred observability:
     AssertionError: expected '-------------------------------------…' to include 'src/components/counter.jsx'
```

Note that the two symptom assertions pass in both runs — they document the broken artifact this warning is about: a bundle reading `Signal` with nothing defining it.

After the change:

```
  Build Greenwood With:
    JSX based Signals Reactivity opted into per file while the plugin option is not enabled
      Running Smoke Tests: JSX based Signals Reactivity opted into per file while the plugin option is not enabled
        Public Directory Generated Output
          ✔ should create a public directory
          ✔ should output one graph.json file
          ✔ should not output any map files for HTML pages
      Build command output for the mismatched inferredObservability configuration
        ✔ should report the JSX module that opted into inferred observability
      Build command output for the emitted Signals bundle and page
        ✔ should emit a bundle that reads the Signal global
        ✔ should not contain a script tag for exposing Signal globally


  6 passing (3s)
```

The full `packages/plugin-import-jsx` suite passes (81 passing), and `prettier --check`, `eslint` and `tsc --project ./tsconfig.json` are clean on the changed files.

### DIFF
--- a/packages/plugin-import-jsx/src/index.js
+++ b/packages/plugin-import-jsx/src/index.js
@@ -14,6 +14,18 @@ const importMap = {
   "wc-compiler/effect": "/node_modules/wc-compiler/src/effect.js",
 };
 
+// a resource plugin is instantiated per lifecycle, so track reported modules at the module level
+const reportedObservabilityMismatches = new Set();
+
+function hasInferredObservabilityExport(tree) {
+  return tree.body.some(
+    (node) =>
+      node.type === "ExportNamedDeclaration" &&
+      node.declaration?.type === "VariableDeclaration" &&
+      node.declaration.declarations?.[0]?.id?.name === "inferredObservability",
+  );
+}
+
 class ImportJsxResource {
   constructor(compilation, options) {
     this.compilation = compilation;
@@ -35,6 +47,18 @@ class ImportJsxResource {
     // https://github.com/ProjectEvergreen/wcc/issues/116
     const tree = parseJsx(url);
     const result = generate(tree);
+
+    if (
+      !this.inferredObservability &&
+      !reportedObservabilityMismatches.has(url.pathname) &&
+      hasInferredObservabilityExport(tree)
+    ) {
+      reportedObservabilityMismatches.add(url.pathname);
+      console.warn(
+        "Configuration warning: a JSX module exports inferredObservability but the plugin option is not enabled, so the Signal polyfill will not be injected and the page will throw at runtime.  Please pass greenwoodPluginImportJsx({ inferredObservability: true }).",
+      );
+      console.debug(url.pathname);
+    }
 
     return new Response(result, {
       headers: new Headers({
--- a/packages/plugin-import-jsx/src/index.js
+++ b/packages/plugin-import-jsx/src/index.js
@@ -14,7 +14,6 @@ const importMap = {
   "wc-compiler/effect": "/node_modules/wc-compiler/src/effect.js",
 };
 
-// a resource plugin is instantiated per lifecycle, so track reported modules at the module level
 const reportedObservabilityMismatches = new Set();
 
 function hasInferredObservabilityExport(tree) {
--- a/packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/loaders-build.prerender-inferred-observability-mismatch.spec.js
+++ b/packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/loaders-build.prerender-inferred-observability-mismatch.spec.js
@@ -66,7 +66,7 @@ describe("Build Greenwood With: ", function () {
 
     describe("Build command output for the mismatched inferredObservability configuration", function () {
       it("should report the JSX module that opted into inferred observability", function (done) {
-        expect(output).to.contain(path.join("src", "components", "counter.jsx"));
+        expect(output).to.contain("src/components/counter.jsx");
 
         done();
       });
--- a/packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/loaders-build.prerender-inferred-observability-mismatch.spec.js
+++ b/packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/loaders-build.prerender-inferred-observability-mismatch.spec.js
@@ -1,0 +1,116 @@
+/*
+ * Use Case
+ * Run Greenwood build command with a JSX component that opts into Signals reactivity through the
+ * file level `export const inferredObservability` while the plugin is configured without the
+ * matching `inferredObservability` option.
+ *
+ * User Result
+ * Should report the mismatch at build time, since the emitted bundle references the Signal global
+ * that the plugin only injects when its own option is enabled.
+ * https://github.com/ProjectEvergreen/greenwood/issues/1780
+ *
+ * User Command
+ * greenwood build
+ *
+ * User Config
+ * import { greenwoodPluginImportJsx } from '@greenwood/plugin-import-jsx';
+ *
+ * {
+ *   prerender: true,
+ *   plugins: [
+ *     greenwoodPluginImportJsx()
+ *   ]
+ * }
+ *
+ * User Workspace
+ * src/
+ *   components/
+ *     counter.jsx
+ *   pages/
+ *     index.html
+ */
+import { expect } from "chai";
+import fs from "node:fs/promises";
+import { JSDOM } from "jsdom";
+import path from "node:path";
+import { runSmokeTest } from "../../../../../test/smoke-test.js";
+import { Runner } from "gallinago";
+import { fileURLToPath } from "node:url";
+
+describe("Build Greenwood With: ", function () {
+  const LABEL =
+    "JSX based Signals Reactivity opted into per file while the plugin option is not enabled";
+  const cliPath = path.join(process.cwd(), "packages/cli/src/bin.js");
+  const outputPath = fileURLToPath(new URL(".", import.meta.url));
+  let runner;
+  let output = "";
+
+  before(function () {
+    this.context = {
+      publicDir: path.join(outputPath, "public"),
+    };
+    runner = new Runner(false, true);
+  });
+
+  describe(LABEL, function () {
+    before(async function () {
+      await runner.setup(outputPath);
+      await runner.runCommand(cliPath, "build", {
+        onStdOut: (message) => {
+          output += message;
+        },
+      });
+    });
+
+    runSmokeTest(["public"], LABEL);
+
+    describe("Build command output for the mismatched inferredObservability configuration", function () {
+      it("should report the JSX module that opted into inferred observability", function (done) {
+        expect(output).to.contain(path.join("src", "components", "counter.jsx"));
+
+        done();
+      });
+    });
+
+    describe("Build command output for the emitted Signals bundle and page", function () {
+      let dom;
+      let scripts;
+
+      before(async function () {
+        dom = await JSDOM.fromFile(path.resolve(this.context.publicDir, "./index.html"));
+        scripts = await Array.fromAsync(
+          fs.glob("counter.*.js", { cwd: new URL("./public/", import.meta.url) }),
+        );
+      });
+
+      it("should emit a bundle that reads the Signal global", async function () {
+        const contents = await fs.readFile(
+          new URL(`./public/${scripts[0]}`, import.meta.url),
+          "utf-8",
+        );
+
+        expect(scripts.length).to.equal(1);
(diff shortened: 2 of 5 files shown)


packages/plugin-import-jsx/test/cases/loaders-build.prerender-inferred-observability-mismatch/loaders-build.prerender-inferred-observability-mismatch.spec.js
```diff
@@ -0,0 +1,116 @@
...
+          output += message;
+        },
+      });
+    });
+
+    runSmokeTest(["public"], LABEL);
+
+    describe("Build command output for the mismatched inferredObservability configuration", function () {
+      it("should report the JSX module that opted into inferred observability", function (done) {
+        expect(output).to.contain("src/components/counter.jsx");
```
thescientist13, line 69: I thought it would probably be also worth asserting that the key log line, the warning about the configuration error, was also displayed, otherwise just the `pathname` alone doesn't seem that helpful, IMO

I think just adding this would sufficient to make sure it got logged
```js
expect(output).to.contain("Configuration warning:");
```

packages/plugin-import-jsx/src/index.js
```diff
@@ -36,6 +47,18 @@ class ImportJsxResource {
...
+    if (
+      !this.inferredObservability &&
+      !reportedObservabilityMismatches.has(url.pathname) &&
+      hasInferredObservabilityExport(tree)
+    ) {
+      reportedObservabilityMismatches.add(url.pathname);
+      console.warn(
+        "Configuration warning: a JSX module exports inferredObservability but the plugin option is not enabled, so the Signal polyfill will not be injected and the page will throw at runtime.  Please pass greenwoodPluginImportJsx({ inferredObservability: true }).",
+      );
+      console.debug(url.pathname);
```
thescientist13, line 59: would it be useful to contextualize this log line a bit? That might be a good place to include the pathname
```js
console.warn(
  "Configuration warning: a JSX module (src/path/to/file) exports inferredObservability but the plugin option is not enabled, so the Signal polyfill will not be injected and the page will throw at runtime.  Please pass greenwoodPluginImportJsx({ inferredObservability: true }).",
);
```

WDYT?

thescientist13 (commented):
Left a couple small feedback, and seeing this more of an enhancement than a bug, so just tweaking the PR description a bit (it feels similar in nature to #1772 )

PR #1787, opened 2026-08-16
enhancement(cli): #1785 prefix all node specifiers when bundling

### LINKED ISSUE #1785: prefix all node builtin specifiers with `node:` when generating server bundles
### Current State

While Greenwood does #1324 for it's own code, in order to best support additional runtimes like [Deno](https://github.com/ProjectEvergreen/greenwood/issues/1322#issuecomment-5173619111) and [Cloudflare](https://github.com/ProjectEvergreen/greenwood/issues/1143#issuecomment-3315249759), it would be useful to have Greenwood do this automatically for bundled code, to support better portability across runtimes
- https://community.cloudflare.com/t/questions-involving-node-js-compat-issues/547534/10
- https://docs.deno.com/examples/migrate_node_project_tutorial/

> _Regarding Cloudflare, it looks like it may not be a requirement anymore, but it requires knowing the right version of wrangler to use, but given the convention over this, I think it's still worthwhile to do it in core, since it may come up for other use cases over time._

### Desired State

We can use Node to get the list of all built-ins
https://nodejs.org/docs/latest/api/module.html#modulebuiltinmodules
```ts
import { builtinModules } from 'node:module';

console.log(builtinModules);
```

Then leverage that in our Rollup configuration pretty easily
```js
import { builtinModules as builtin } from 'node:module';

// ...

function prefixNodeBuiltins() {
  return {
    name: "greenwood-prefix-node-builtins",
    resolveId: {
      order: "pre",
      handler(specifier) {
        if (builtinModules.has(specifier)) {
          return { id: `node:${specifier}`, external: true };
        }
      },
    },
  };
}
```

### Additional Context

In addition to this, we should also do a little clean up work while we're her
1. rollup config is not passing custom rollup plugins to SSR and API route configs (but we also need to filter out Greenwood default plugins like terser from mangling input sources, like WCC)
1. should also fold in https://github.com/ProjectEvergreen/greenwood/issues/1391

thescientist13:
## Related Issue

resolve #1785 

## Documentation 

N / A

## Summary of Changes

1. prefix all node specifiers when bundling
1. add test cases

### DIFF
--- a/packages/cli/src/config/rollup.config.js
+++ b/packages/cli/src/config/rollup.config.js
@@ -12,8 +12,8 @@ import { ACORN_OPTIONS } from "@greenwood/cli/src/lib/parsing-utils.js";
 // ConstructableStylesheets, JSON Modules
 const externalizedResources = ["css", "json"];
 
-// one-time conversion of array to set for quicker lookup
-const NODE_BUILTINS = new Set(builtinModules.map((specifier) => specifier.replace(/^node:/, "")));
+// one-time conversion of tje builtins array to a Set for quicker lookup
+const NODE_BUILTINS = new Set(builtinModules);
 
 // https://github.com/rollup/rollup/issues/2121
 // would be nice to get rid of this
@@ -25,7 +25,7 @@ function prefixNodeBuiltins() {
   return {
     name: "greenwood-prefix-node-builtins",
     resolveId(id) {
-      if (NODE_BUILTINS.has(id)) {
+      if (!id.startsWith("node:") && NODE_BUILTINS.has(id)) {
         return {
           id: `node:${id}`,
           external: true,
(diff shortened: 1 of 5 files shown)


packages/cli/src/config/rollup.config.js
```diff
@@ -8,14 +9,31 @@ import * as acorn from "acorn";
 import * as walk from "acorn-walk";
 import { ACORN_OPTIONS } from "@greenwood/cli/src/lib/parsing-utils.js";
 
+// ConstructableStylesheets, JSON Modules
+const externalizedResources = ["css", "json"];
+
+// one-time conversion of tje builtins array to a Set for quicker lookup
```
thescientist13, line 15: Typo here on _tje_

PR #1789, opened 2026-08-24
fix(cli): #1788 return the intended status for active content graph error paths

### LINKED ISSUE #1788: activeContent graph.json error paths return 500 instead of the intended 403/400
### What happened

Requesting `/___graph.json` from the dev server with a missing or unrecognized `x-content-key` header returns **500** with a `TypeError` stack in the server log, instead of the intended client-error status.

`serve()` computes a `status` and a `body`, then funnels every path through one return that calls `pruneGraph(body)`. `pruneGraph` does `pages.map(...)`, so it only works when `body` is an array. Two paths reach it with something else:

1. **Missing content key** — the plugin deliberately sets `status = 403` and `body = "Bad Request - No Cache Key found"`, a *string*. `"...".map` is not a function, the request throws, and the server responds 500. The 403 is unreachable dead code.
2. **Unrecognized content key** — when the key's type is not `graph`, `collection` or `route`, `body` is never assigned and stays `undefined`. `undefined.map` throws, and again the response is 500.

### Steps to reproduce

With `activeContent: true` in `greenwood.config.js`, run `greenwood develop`, then:

```sh
# expected 403, actual 500
curl -i http://localhost:1984/___graph.json

# expected a client error, actual 500
curl -i http://localhost:1984/___graph.json -H 'x-content-key: bogus-key'
```

Both return `500` and log `TypeError: pages.map is not a function`. A well-formed key (`graph`, `route-/blog`, `collection-nav`) behaves correctly.

### Environment

- Greenwood `master` @ `9b5bd41a`
- Node 24.12.0
- Linux (arm64)

### Additional context

Root cause is the single shared return in `serve()` at `packages/cli/src/plugins/resource/plugin-active-content.js:78`, which calls `pruneGraph` unconditionally; `pruneGraph` is `packages/cli/src/lib/content-utils.js:4`. Both error paths need to return before reaching it.

jstockdi:
### Related Issue

Resolves #1788

### Documentation

No documentation changes needed — this restores the status codes the plugin already intended to send, and adds no new public API or configuration.

### Summary of Changes

1. [x] Return the 403 `Bad Request - No Cache Key found` response directly when `x-content-key` is missing, before anything calls `pruneGraph`. Previously this path assigned a *string* body and then hit `pruneGraph`'s `pages.map`, throwing and turning the intended 403 into a 500 — the 403 was dead code.
2. [x] Return **400 `Bad Request - Unsupported Cache Key`** for a content key that is not `graph`, `collection-*` or `route-*`. Previously `body` stayed `undefined` and `undefined.map` threw, also producing a 500.
3. [x] The `graph` / `collection` / `route` success paths are unchanged, including the `__GWD_COMMAND__ === "build"` data-file write. The only edit to them is de-indentation from removing the wrapping `else`, plus `status` becoming the literal `200` now that the shared return is reached only on success.
4. [x] The 400 message deliberately does not echo the client-supplied key back into the response body.
5. [x] Folded the regression coverage into the existing `develop.config.active-content` case rather than adding a new fixture directory, following your note on #1720. Two new specs assert the status codes rather than stack text: no header → 403, `bogus-key` → 400.
6. [x] Verified the tests fail on unpatched `master` with exactly this symptom — `expected 500 to equal 403` and `expected 500 to equal 400`, 9 passing / 2 failing — and pass after the fix at 11 passing.
7. [x] `yarn lint` (0 errors) and `yarn format:check` clean. The neighbouring `serve.config.active-content` (35 passing) and the `build.config.base-path-active-content` / `develop.config.active-content-collection-hyphen` cases (33 passing) are unchanged.

One open question: 400 on an unrecognized key is the strict reading, since the client asked for something the server cannot supply. The lenient alternative is **200 with `[]`**, treating an unknown key as an empty result set, which would keep the data client's consumers on the happy path and never surface an error to page code. Happy to switch to that if you prefer it — the 403 half of the fix is unaffected either way.

### DIFF
--- a/packages/cli/src/plugins/resource/plugin-active-content.js
+++ b/packages/cli/src/plugins/resource/plugin-active-content.js
@@ -44,38 +44,36 @@ class ContentAsDataResource {
     const delimiterIndex = contentKey.indexOf("-");
     const type = delimiterIndex === -1 ? contentKey : contentKey.slice(0, delimiterIndex);
     const value = delimiterIndex === -1 ? "" : contentKey.slice(delimiterIndex + 1);
-    let status;
     let body;
 
     if (contentKey === "") {
-      status = 403;
-      body = "Bad Request - No Cache Key found";
+      return new Response("Bad Request - No Cache Key found", { status: 403 });
+    }
+
+    if (contentKey === "graph") {
+      body = graph;
+    } else if (type === "collection") {
+      body = filterContentByCollection(graph, value);
+    } else if (type === "route") {
+      body = filterContentByRoute(graph, value);
     } else {
-      status = 200;
+      return new Response("Bad Request - Unsupported Cache Key", { status: 400 });
+    }
 
-      if (contentKey === "graph") {
-        body = graph;
-      } else if (type === "collection") {
-        body = filterContentByCollection(graph, value);
-      } else if (type === "route") {
-        body = filterContentByRoute(graph, value);
-      }
+    if (process.env.__GWD_COMMAND__ === "build") {
+      const fileKey = `./data-${contentKey.replace(/\//g, "_")}.json`;
 
-      if (process.env.__GWD_COMMAND__ === "build") {
-        const fileKey = `./data-${contentKey.replace(/\//g, "_")}.json`;
-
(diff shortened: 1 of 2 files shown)


thescientist13 (approved):
Looks good, thank you! 💯

